PR #10, opened 2022-05-18
docs: ADR for content flexibilty

ormsbee:
This describes the high level plan for breaking up the current course experience into components that could be reconfigured to provide more flexible learning content experiences.

### DIFF
--- a/docs/decisions/0002-content-flexibility.rst
+++ b/docs/decisions/0002-content-flexibility.rst
@@ -41,7 +41,7 @@ Navigation
 
 There are two goals:
 
-#. Each of these concepts should be *extensible*, so that new types of Blocks, Units, etc. are created (the specifics of this would be in a follow-on ADR).
+#. Each of these concepts should be *extensible*, so that new types of Items, Units, etc. are created (the specifics of this would be in a follow-on ADR).
 #. We should be able to use these in different combinations. For instance, a short course may have a different Navigation type than traditional courses (e.g. by removing a level of hierarchy), but that Navigation would then still point to Sequences, which point to Units, etc. The base types described in this document provide a lowest-common denominator interface to decouple the layers from each other, so that an extended Navigation type doesn't have to be aware of the various extended Sequence types, but can just know that is pointing to some sort of Sequence.
 
 Consequences
--- a/docs/decisions/0002-content-flexibility.rst
+++ b/docs/decisions/0002-content-flexibility.rst
@@ -1,0 +1,52 @@
+2. Approach to Content Flexibility
+==================================
+
+Context
+-------
+
+Open edX courses follow a strict Course > Section > Subsection > Unit > Module hierarchy. There are a number of use cases that do not fit this pattern:
+
+* A problem bank shared via LTI, with individual problems for use in other LMS systems.
+* Short courses that do not require the "Section" middle layer of hierarchy.
+* Standalone assessments given after an informational session.
+* LabXchange-style pathways and clusters.
+
+The Learning Core should have a set of composable primitives that could be configured to construct these sorts of experiences, while still maintaining the ability to support the traditional Open edX course format.
+
+Decision
+--------
+
+The following are foundational, extensible concepts in the Learning Core, which can be combined in different ways:
+
+Block
+  A Block is a small piece of content, like a video, problem, or bit of HTML text. It has an identity, renderer, and potentially student state. It is not a container for other content, and has no child elements.
+  
+  Blocks are analogous to the "Module" portion of the traditional Open edX course.
+
+Segment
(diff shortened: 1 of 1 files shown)


bradenmacdonald:
Great job writing this up :) 👍🏻

docs/decisions/0002-content-flexibility.rst
```diff
@@ -0,0 +1,52 @@
...
+* LabXchange-style pathways and clusters.
+
+The Learning Core should have a set of composable primitives that could be configured to construct these sorts of experiences, while still maintaining the ability to support the traditional Open edX course format.
+
+Decision
+--------
+
+The following are foundational, extensible concepts in the Learning Core, which can be combined in different ways:
+
+Block
```
ormsbee, line 21: Maybe this should be "Item" instead, to differentiate it more from structural XBlocks, and to be more in line with QTI terminology?

docs/decisions/0002-content-flexibility.rst
```diff
@@ -0,0 +1,52 @@
...
+
+  A Sequence is analogous to a Subsection in a traditional Open edX course.
+
+Navigation
+  Navigation is the higher-level organization of a course or other learning context. For a traditional course, this would be what determines the Sections and what Sequences are in each.
+
+There are two goals:
+
+#. These concepts should be *extensible*, so that new types of Blocks, Units, etc. are created (the specifics of this would be in a follow-on ADR).
+#. We should be able to use these in different combinations. For instance, a short course may have a different Navigation type than traditional courses, but that Navigation would still point to Sequences, which point to Units.
```
kdmccormick, line 45: > For instance, a short course may have a different Navigation type than traditional courses, but that Navigation would still point to Sequences, which point to Units.

From this I glean that non-traditional learning contexts will still use this hierarchy, even if they customize within each level of hierarchy. Is that right?

> We should be able to use these in different combinations.

I'm not sure what you mean by this.

kdmccormick (commented):
Unsurprisingly, the plan here sounds great :) Just some minor comments on wording and naming.

docs/decisions/0002-content-flexibility.rst
ormsbee: > From this I glean that non-traditional learning contexts will still use this hierarchy, even if they customize within each level of hierarchy. Is that right?

Yes.

>> We should be able to use these in different combinations.
>
> I'm not sure what you mean by this.

Right, that wasn't clear at all. I meant both that the hierarchy can mix extensible types in the sense that you can have a static course navigation pointing to some sequences that are static and some that are dynamically generated sequences, etc. But I'll try to phrase it better.

docs/decisions/0002-content-flexibility.rst
kdmccormick: > One thing I realized about Atom is that if we have a project about Atomic Learning Units and adopt that language, it's going to be painful because Atoms in the data model would actually be one or two layers more granular than what we're presenting to the user.

@ormsbee You're right, that'd be an unfortunate mismatch in terminology. We should call the blocks `Hadrons` (just kidding :).

This has me questioning the name of the project now! "Atom" comes from the Greek word "indivisible". The name "Atomic Learning Units" implies a system of pieces learning content that cannot be broken down further, but doesn't speak the intended outcome of the project (which is to have "Composable" or "Modular" learning units, right?).

ormsbee:
@kdmccormick: Updated based on your feedback.

docs/decisions/0002-content-flexibility.rst
kdmccormick: Yeah that's fair!

docs/decisions/0002-content-flexibility.rst
```diff
@@ -0,0 +1,52 @@
...
+  A Sequence is a collection of Units that are presented one after the other, either to assess student understanding or to achieve some learning objective. 
+
+  A Sequence is analogous to a Subsection in a traditional Open edX course.
+
+Navigation
+  Navigation is the higher-level organization of a course or other learning context. For a traditional course, this would be what determines the Sections and what Sequences are in each. A more advanced Navigation type might dynamically select Sequences that are appropriate for you based on your understanding of the material.
+
+There are two goals:
+
+#. Each of these concepts should be *extensible*, so that new types of Blocks, Units, etc. are created (the specifics of this would be in a follow-on ADR).
```
kdmccormick, line 44:
```suggestion
#. Each of these concepts should be *extensible*, so that new types of Items, Units, etc. are created (the specifics of this would be in a follow-on ADR).
```

kdmccormick (approved):
One suggested update, but otherwise, 🚀 !

PR #11, opened 2022-05-18
docs: add ADR for content extensibility

ormsbee:
Outline our approach for third-party apps to extend Learning Core data models.

### DIFF
--- a/docs/decisions/0003-content-extensibility.rst
+++ b/docs/decisions/0003-content-extensibility.rst
@@ -13,7 +13,7 @@ Developer Experience Pain Points
 
 #. Plugin applications cannot make queries about learning content without calling into the edx-platform monolith, which makes testing, development, and long term maintenance more difficult.
 #. Plugins are prone to bugs around the content publishing lifecycle, such as when content that a plugin cares about is changed or deleted while a course is running.
-#. XBlocks data is stored in a way that is opaque at the database layer. Individual fields are collapsed into large key/value documents, making joins and rich database-level querying difficult.
+#. XBlock data is stored in a way that is opaque at the database layer. Individual fields are collapsed into large key/value documents, making joins and rich database-level querying difficult.
 
 Application Lifecycle and Platform Resilience Pain Points
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/decisions/0003-content-extensibility.rst
+++ b/docs/decisions/0003-content-extensibility.rst
@@ -1,0 +1,63 @@
+3. Content Extensibility Through Model Relations
+================================================
+
+Context
+-------
+
+Content extensibility has always been critical to the success of Open edX, with the most obvious example being XBlocks as a mechanism to create custom problem types. Plugin apps have also long interacted with the ModuleStore in order to pull back content information as part of new, custom functionality.
+
+We want the Learning Core to build upon these successes, while addressing a number of architectural pain points in the current system.
+
+Developer Experience Pain Points
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Plugin applications cannot make queries about learning content without calling into the edx-platform monolith, which makes testing, development, and long term maintenance more difficult.
+#. Plugins are prone to bugs around the content publishing lifecycle, such as when content that a plugin cares about is changed or deleted while a course is running.
+#. XBlocks data is stored in a way that is opaque at the database layer. Individual fields are collapsed into large key/value documents, making joins and rich database-level querying difficult.
+
+Application Lifecycle and Platform Resilience Pain Points
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Content can become completely inaccessible after a plugin has been removed. It was once the case that uninstalling an XBlock would prevent the export of any course that used that XBlock.
+#. Hierarchical nesting in import/export means that removing a container XBlock type (e.g. SplitTestBlock) would also completely remove any content in those containers from future exports.
+#. Plugin bugs have in the past broken all export functionality, making it difficult to reproduce the issue locally (for lack of access to production content data).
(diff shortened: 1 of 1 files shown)


docs/decisions/0003-content-extensibility.rst
```diff
@@ -0,0 +1,63 @@
...
+
+#. There will be a core data model that is always introspectable and exportable, without invoking plugin/extension code.
+#. Plugin apps will be able to progressively enhance this core data model by creating their own related models, typically linked to core data models using ``OneToOneField``.
+#. Plugin apps are able to extend Core concepts such as Units into what would effectively be sub-types.
+#. It will be possible to migrate existing content data over time, as new plugin apps become available.
+#. All content and versions of content will have UUIDs to allow for stable references across services.
+
+This layering of related models will add complexity to the data model, but we accept that tradeoff to decouple plugin models from the core application and from other plugins. To make this easier to deal with, openedx-learning should provide abstract models for common use cases, and expose those via a ``models_api.py`` module. This will lower the barrier to entry for developers, and allow us to more easily enforce conventions like setting ``primary_key=True`` with our ``OneToOneField`` relationships.
```
ormsbee, line 38: I just realized that there are already `models_api.py` files in edx-platform, but it looks like it's used for operations on models, not as a place to hold abstract models or other model-level linking. Maybe I should go with `api_models.py` instead? :-(

docs/decisions/0003-content-extensibility.rst
```diff
@@ -0,0 +1,63 @@
...
+Content extensibility has always been critical to the success of Open edX, with the most obvious example being XBlocks as a mechanism to create custom problem types. Plugin apps have also long interacted with the ModuleStore in order to pull back content information as part of new, custom functionality.
+
+We want the Learning Core to build upon these successes, while addressing a number of architectural pain points in the current system.
+
+Developer Experience Pain Points
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Plugin applications cannot make queries about learning content without calling into the edx-platform monolith, which makes testing, development, and long term maintenance more difficult.
+#. Plugins are prone to bugs around the content publishing lifecycle, such as when content that a plugin cares about is changed or deleted while a course is running.
+#. XBlocks data is stored in a way that is opaque at the database layer. Individual fields are collapsed into large key/value documents, making joins and rich database-level querying difficult.
```
kdmccormick, line 16:
```suggestion
#. XBlock data is stored in a way that is opaque at the database layer. Individual fields are collapsed into large key/value documents, making joins and rich database-level querying difficult.
```

kdmccormick (approved):
💯 

Just one suggestion: Would it be worth mentioning if/how this repo will interact with openedx-events and openedx-filters? I am imagining it exposing filters along the lines of `OPENEDX_LEARNING_{UNIT,BLOCK,etc.}_TYPES` ... is that right?

docs/decisions/0003-content-extensibility.rst
kdmccormick: I know it is against [edx-platform ADR 2](https://github.com/openedx/edx-platform/blob/master/docs/decisions/0002-inter-app-apis.rst) to import directly from an app's `models` module, and for good reason. But that seems to run counter to common practice Django community: Django's own standard library doesn't seem to have any qualms about it (`from django.contrib.auth.models import User`).

So, perhaps it's fine if libraries like this one expose models from `models` the modules? For internal models, we could employ another convention, like prefixing their names with an underscore or namespacing them within a `internal` package.

ormsbee: > I know it is against [edx-platform ADR 2](https://github.com/openedx/edx-platform/blob/master/docs/decisions/0002-inter-app-apis.rst) to import directly from an app's models module, and for good reason. But that seems to run counter to common practice Django community: Django's own standard library doesn't seem to have any qualms about it (from django.contrib.auth.models import User).

The `User` model is an interesting example, because Django has multiple APIs to try to make things work more smoothly, like [`get_user_model`](https://docs.djangoproject.com/en/3.2/topics/auth/customizing/#django.contrib.auth.get_user_model) and the [`AUTH_USER_MODEL`](https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-AUTH_USER_MODEL) setting. Those aren't directly relevant to our situation, but I just highlight it as an example where there's a surprising amount of machinery around model dependencies.

Also, even though we have an ADR that discourages it, [OEP-49](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0049-django-app-patterns.html) (which came after that ADR and was partially informed by it) carves out a [performance caveat](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0049-django-app-patterns.html#performance-caveat) to allow for importing and linking against certain models. So we've definitely already acknowledged the need for those dependencies. I'd add that data integrity is also a strong motivator here, particularly for cleanup situations.

> So, perhaps it's fine if libraries like this one expose models from models the modules? For internal models, we could employ another convention, like prefixing their names with an underscore or namespacing them within a internal package.

Yeah, I think some exposure is appropriate, for both performance and data integrity reasons. I think it will be more intuitive if all the models are normal and in `models.py` and there's a separate module that re-exports the explicitly re-usable models (so that we can gradually promote them over time). But I'm honestly not sure–I think we'd have to experiment with it a bit and see how it feels.

One of my overarching concerns is that we get to add little safety measures like this without adding so much that it becomes jarring for typical Django developers to jump into. Though I confess, I'm often not sure which path is better in that respect.

ormsbee:
> Just one suggestion: Would it be worth mentioning if/how this repo will interact with openedx-events and openedx-filters? I am imagining it exposing filters along the lines of OPENEDX_LEARNING_{UNIT,BLOCK,etc.}_TYPES ... is that right?

I'm honestly not sure exactly how they'll fit together at this point, so I'd rather not mention them in this ADR. I'm confident that they do complement each other, but I'm fuzzier on the specifics: for instance, `openedx-filters` is currently very edx-platform specific. Does it make sense to hook some of the filters here as well, or will there be a separate filters repo for this, or does the smaller size of this repo warrant having its own filters just as a package here (following openedx-filters conventions) instead of in a separate repo?

So I'd like to punt on that one for now.

Issue #1, opened 2022-02-06
Determine LMS Content Data Model Design


ormsbee:
This is a discovery ticket that would end in multiple ADRs in this repo. The goal is to determine the basic data structures and relationships needed to store content data for the LMS with the following goals:

1. Handle all existing course use cases.
2. Handle v2 content libraries, and library resources shared across multiple courses, with different policies/overrides.
3. Handle potentially non-XBlock content.
4. Allow for fast publishing.
5. Allow for atomic publishing.
6. Allow for easy export.
7. Allow third party applications to build their own advanced data structures on top of it.
8. Allow for efficient querying.
9. Minimize the amount of wasted space caused by nearly identical versions caused by minor edits.

There will likely be some prototyping involved in this, as well as a lot of discussion.

These ADRs would include:

- [ ] Determine high level approach to splitting reusable content and learning context specific policy.
- [ ] Determine high level approach to versioning and incremental publishing.
- [ ] Determine leaf XBlock-level modeling.
- [ ] Determine unit-level modeling.
- [ ] Determine sequence-level modeling.

----

Open edX's ModuleStore has explicit versioning capabilities (though they're not really used from LMS most of the time). Blockstore has versioning deeply baked into the design, but we've generally avoided encoding multi-version support into the post-publish data stores we build during course publish (e.g. CourseOverview, Block Transformers/Course Blocks API, etc.)

I think we've come to a point where we really do need to cross that divide and start introducing content versioning concepts to the LMS more generally. Some motivating reasons:

1. It's difficult to preview changes before publishing if the LMS can only display the published version.
11. The design of content libraries assumes that multiple versions are available simutaneously.
12. Building all these separate stores of published data takes time, and parts may fail. When this happens, our course is in an inconsistent half-published state. For example, the rendered course content may be updated but the course outline generation may have failed so that nobody can see the new content. Ideally, we'd want to give these systems time to build up the data that they need, and then change everything atomically at once.
13. Other systems like scoring and state storage would benefit from being able to store the actual version it was created against.

### Storage Scaling Issues

SplitMongo ModuleStore wasted a lot of space with old version data, eventually leading us to create [a separate cleanup script](https://github.com/openedx/tubular/blob/master/tubular/scripts/structures.py) for it. There were a number of reasons why disk usage got so bad with this system:

1. Structural data was stored inefficiently.
2. New versions were being published for tiny changes.
3. There was no cleanup.
4. Historical course data had very little use.

[...]

PR #13, opened 2022-05-26
Another pass at basic content models

ormsbee:
This splits Item-related models into a new app and otherwise follows the logical split discussed [here](https://github.com/openedx/openedx-learning/issues/1#issuecomment-1134873686).


openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
...
+    # TODO: These pointers to the latest published version are convenient, but
+    # we don't currently have the data integrity guarantees to make sure that 
+    # multiple versions aren't active at the same time. Maybe add a ref to the
+    # Item model in LearningContextVersionItemVersion, so we can force that
+    # constraint?
+    published_version = models.ForeignKey('ItemVersion', on_delete=models.RESTRICT, null=True, related_name="+")
+    first_published_at = models.DateTimeField(null=True)  # can be null if never published 
+    last_published_at = models.DateTimeField(null=True)   # can be null if never published
+
+    created = manual_date_time_field()
```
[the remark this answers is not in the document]
ormsbee: I tried to give the rationale here: https://github.com/openedx/openedx-learning/pull/13/files#diff-7844f3ef353b5f0e84d0cfdcd0798d9a0922af8d369dfabd7ce35636d9e4b22eR80-R96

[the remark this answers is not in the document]
feanil: Makes sense thanks!

openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    A particular version of an Item.
+
+    A new ItemVersion is created anytime there is either a change to the content
+    or a change to the policy around a piece of content (e.g. schedule change).
+    """
+    uuid = immutable_uuid_field()
+    item_info = models.ForeignKey('ItemInfo', on_delete=models.RESTRICT)
+    item = models.ForeignKey(Item, on_delete=models.CASCADE)
+    title = models.CharField(max_length=1000, blank=True, null=True)
```
kdmccormick, line 65: Why is `title` here and not in ItemInfo? Is it because it's example of policy, which we're distinguishing from metadata (like `type`)?

openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
...
+    uuid = immutable_uuid_field()
+    learning_context = models.ForeignKey(LearningContext, on_delete=models.CASCADE)
+    identifier = identifier_field()
+
+    # TODO: These pointers to the latest published version are convenient, but
+    # we don't currently have the data integrity guarantees to make sure that 
+    # multiple versions aren't active at the same time. Maybe add a ref to the
+    # Item model in LearningContextVersionItemVersion, so we can force that
+    # constraint?
+    published_version = models.ForeignKey('ItemVersion', on_delete=models.RESTRICT, null=True, related_name="+")
```
kdmccormick, line 36: knowledge check: this is secondary source of information, since one could authoritatively determine this by asking the `publish` app what the published LearningContextVersion is, and then looking at the ItemVersions within that LCV, right?

openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
+"""
+The item model hiearachy is: Item -> ItemVersion -> ItemInfo -> ItemRaw
+
+Item is the versionless thing that is guaranteed to exist for the lifetime of
+the LearningContext. An ItemVersion is a different version of that item for a
+given LearningContext, and may include policy changes (like grading). ItemInfo
+covers basic metadata that is intrinsic to the item itself, and now how it's
+used in a LearningContext. ItemRaw represents the raw byte data.
+
+TODO: Add link to ADR after it merges.
```
kdmccormick, line 10: Ack, sorry, I am struggling to rebuild this hierarchy in my head. I'm going to try to rephrase it here. Let me know if I'm off:
* ...Assume "LC" is some LearningContext, and "LCV" some LearningContextVersion within LC.
* An ItemRaw is a blob of byte data and a mimetype.
  * Example: an ItemRaw instance "PR", containing some OLX file
* An ItemInfo associates an ItemRaw with metadata, including its LearningContext and its item type.
  * Example: an ItemInfo instance "PI", which tells us that PR is a `problem` and belongs to LC.
* An Item is an atomic block of content associted with a LC. In isolation, an Item does not tell us about its content, because content is associated with versions and an Item is versionless.
  * Example: an Item instance "P" in LC.
 * An ItemVersion joins together an abstract Item instance with a concrete ItemInfo, denoting a particular version of the content, policy, and metadata of an Item.
   * Example: an ItemVersion instance "PV", joining together P and PI.
   * Example: an ItemVersion instance "PV2", joining together P and PI, with some updated policy.
   * Example: a later ItemVersion instance "PVx", joining together P and some modified ItemInfo "PIx".
   * and so on
 * LearningContextVersionItemVersions relate LearningContextVersions to ItemVersions. This table is the connective tissue telling us which ItemVersions are in a LearningContextVersion (and, by corollary, which Items are in a LearningContext).
   * Example: PV is in LCV
   * Example: PV2 is in "LCV2", a later version of LC
   * Example: PV2 is also in "LCV3", a later version of LC that updates some Item(s) other than P

openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
...
+class ItemInfo(models.Model):
+    """
+    Basic metadata about an Item.
+
+    This is mostly here to give supplmentary models to hang off of this and give
+    more interesting models based on the data in ItemRaw.
+    """
+    uuid = immutable_uuid_field()
+    learning_context = models.ForeignKey(LearningContext, on_delete=models.CASCADE)
+    type = models.CharField(max_length=100)
```
kdmccormick, line 85: Can we always deduce `type` from `item_raw.mime_type`?

kdmccormick (commented):
Sorry I took so long to review this. Most of my comments are just me trying to make sure I grok everything.

projects/dev.py
```diff
@@ -31,9 +31,14 @@
     'django.contrib.admin',
     'django.contrib.admindocs',
 
-    # Our own apps
-    'openedx_learning.apps.core.publish.apps.PublishConfig',
+    # Learning Core Apps
     'openedx_learning.apps.core.compose.apps.ComposeConfig',
+    'openedx_learning.apps.core.itemstore.apps.ItemStoreConfig',
+    'openedx_learning.apps.core.publish.apps.PublishConfig',
```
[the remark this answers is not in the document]
ormsbee: > The `apps` namespace package seems unnecessary and out-of-step with other edx-platform libraries. Why not just `openedx_learning.core.itemstore`?

Mostly because I wanted a place to hang `lib` for things like the field types.

> I notice that you are using verbs for some of the core app names. Wouldn't it be more standard to use nouns? (`composition`, `publishing`)

Yeah, I was going for the verb form for a while before because it was shorter, before I landed on `itemstore` for one of them. I'll switch it back over to nouns.

olx_importer/management/commands/load_course_data.py
```diff
@@ -0,0 +1,200 @@
...
+        # outside of tag declarations as well.
+        static_files_regex = r"""['"]\/static\/(.+?)["'\?]"""
+
+        for xml_file_path in content_path.iterdir():
+            items_found += 1
+            identifier = xml_file_path.stem
+            data_bytes = xml_file_path.read_bytes()
+            hash_digest = create_hash_digest(data_bytes)
+            data_str = codecs.decode(data_bytes, 'utf-8')
+            mime_type = f'application/vnd.openedx.xblock.{block_type}+xml'
```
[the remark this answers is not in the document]
ormsbee: Ah, right. I guess it doesn't hurt to add the v1, though I'm skeptical we'll ever have a v2. (I'm guessing we'd cali it something entirely different, to avoid confusion.) Will fix.

olx_importer/management/commands/load_course_data.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+            static_file_paths = frozenset(
+                f"static/{static_reference_path}"
+                for static_reference_path in re.findall(static_files_regex, data_str)    
+            )
+            for static_file_path in static_file_paths:
+                if static_file_path in static_asset_paths_to_atom_ids:
+                    atom_id = static_asset_paths_to_atom_ids[static_file_path]
+                    ContentObjectPart.objects.create(content_object=content_obj, content_atom_id=atom_id, identifier=static_file_path)
```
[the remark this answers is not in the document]
ormsbee: Ugh, right, because I didn't ever put the static asset mapping stuff back. Thank you.

openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
...
+    first_published_at = models.DateTimeField(null=True)  # can be null if never published 
+    last_published_at = models.DateTimeField(null=True)   # can be null if never published
+
+    created = manual_date_time_field()
+    modified = manual_date_time_field()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["learning_context_id", "identifier"],
```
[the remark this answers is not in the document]
ormsbee: Yeah. I think that in general, the `identifier` fields have to be namespaced so that it's entirely local to one particular LearningContext.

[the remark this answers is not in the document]
kdmccormick: In the case of XBlocks, do you see this `identifier` as being the usage key (`block-v1:a+b+c+type@video+block@xyz`) or the block id (just `xyz`)?

openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
...
+)
+from ..publish.models import LearningContext, LearningContextVersion
+
+
+class Item(models.Model):
+    """
+    This represents any content that has ever existed in a LearningContext.
+    """
+    uuid = immutable_uuid_field()
+    learning_context = models.ForeignKey(LearningContext, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
ormsbee: > Is this the LearningContext that the Item is being used in, or the LearningContext that it's defined in (eg, a content library)? Is that distinction something that even exists at this level?

The one it's being used in, because Item has the identifier, and we need to be able to hang course-level policy changes off of ItemVersion. So the hierarchy kinda looks kind of like this:

```mermaid
graph TD;
    CI[Course Item]-->CIV[Course ItemVersion];
    CIV-->LII[Library ItemInfo];
    LII-->LIR[Library ItemRaw];
    LI[Library Item]-->LIV[Library ItemVersion];
    LIV-->LII;
```

So the course makes a copy of the high level objects (`Item`, `ItemVersion`), and shares the lower-level ones (`ItemInfo`, `ItemRaw`, and anything derived from those).

openedx_learning/apps/core/itemstore/models.py
ormsbee: Yes. And maybe it's not a great idea to have redundancy like this. The main motivation for this was that a lot of the time, you want to bind data against the abstract notion of a piece of content across all its versions (e.g. most student state), and I wanted a really convenient way to get to the current "live" version from there.

openedx_learning/apps/core/itemstore/models.py
ormsbee: Yes. Renaming is desirable for libraries, because you don't know what kind of convention that library is using, e.g. "Problem 22a" may not make as much sense for the new context it's being used in.

openedx_learning/apps/core/itemstore/models.py
ormsbee: I had originally meant it as something different, but I'm not sure if that holds up in the latest iteration.

The difference I had in mind originally was something like an HTMLBlock's HTML. That it would have an `ItemInfo` that showed it was an XBlock with HTML handler, but the raw data had a MIME type of "text/html".

openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
...
+    The other fields on ItemRaw are for data that is intrinsic to the file data
+    itself (e.g. the size). Any smart parsing of the contents into more
+    structured metadata should happen in other models that hang off of ItemInfo.
+ 
+    ItemRaw models are not versioned in any way. The concept of versioning
+    exists at a higher level.
+
+    TODO: Size limit thoughts–configurable? Start at 10 MB?
+    """
+    learning_context = models.ForeignKey(LearningContext, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
ormsbee: Yes. It's mostly there for ownership/cleanup.

openedx_learning/apps/core/itemstore/models.py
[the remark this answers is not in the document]
kdmccormick: That makes sense! The graphic is immensely helpful. I have to admit that the names of the models on their own did not guide me to this understanding.

The split between [Item , ItemVersion] and [ItemInfo, ItemRaw] reminds me of the XBlock framework's distinction between Usage and Definition, respectively. Is that a fair parallel?

openedx_learning/apps/core/itemstore/models.py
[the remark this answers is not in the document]
kdmccormick: Cool. This would be helpful to note in a comment.

openedx_learning/apps/core/itemstore/models.py
kdmccormick: Makes sense!

openedx_learning/apps/core/itemstore/models.py
kdmccormick: That does sound convenient. I guess, you're either going to end up needing to keep this field up-to-date, or you're going to end up with a lot of requests making the same query ("what's the published version?"). I don't think I have the experience to know which of these is an easier challenge to manage.

projects/dev.py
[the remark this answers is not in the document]
kdmccormick: > Mostly because I wanted a place to hang lib for things like the field types.

The 2nd-level packages could be `openedx_learning.core`, `openedx_learning.lib`, and `openedx_learning.contrib`.

I remember as a new dev being frazzled by the long import paths of `openedx.core.djangoapps...` in edx-platform. I could never remember the paths; I was always copying and pasting, and it made the platform feel more overwhelming.

openedx_learning/apps/core/itemstore/models.py
```diff
@@ -0,0 +1,133 @@
...
+
+    ItemRaw stores data in a Binary BLOB `data` field. This data is not
+    normalized in any way, meaning that pieces of content that are semantically
+    equivalent (e.g. differently spaced/sorted JSON) will result in new entries.
+    This model is intentionally ignorant of what these things mean, because it
+    expects supplemental data models to build on top of it.
+
+    The other fields on ItemRaw are for data that is intrinsic to the file data
+    itself (e.g. the size). Any smart parsing of the contents into more
+    structured metadata should happen in other models that hang off of ItemInfo.
```
kdmccormick, line 103: @ormsbee I am struggling a bit with the distinction between ItemInfo and ItemRaw.

What would we lose by collapsing them into a single model? Is there something bad about structured metadata models hanging off of the same model that's storing the raw bytes?

ormsbee: I think I originally created the idea of this model (under a different name), when I was thinking it would have more stuff in it (like a local identifier associated with the content package), but those concepts were simplified out of the design. We would have to be careful to not pull back the raw payload all the time, but that's manageable.

The only remaining use case I can think of is when you want to add a little bit of metadata about the content that is not the content itself–new information entered by a human, not something that's being derived by the code from an `ItemRaw`. At that point, we have to figure out where to hang that data in a way that also clearly signals that this content has changed.

As an example, let's say we have a text description of an image.

*If we attach that data to a model that hangs off of `ItemRaw`*, then there's no indication that the content has been updated. None of the versioning mechanisms above this were invoked or needed to be invoked. So we've done an unexpected mutation, and it will invalidate assumptions around versioning and caching. We can't even create a new `ItemRaw` to bind it to (wasteful as that would be in terms of space), because `ItemRaw` uses a hash of its contents for identity.

*We could attach that data to a model that hangs off of `ItemVersion`.*  That's a little weird because that's not policy data that changes from course to course, but we would have default policy information that's attached at that layer and copied, so maybe it would work out...?

You're right in that other comment, it's very much like the definition/usage split in the old system. But maybe we should take that as a cautionary tale–people put many things in the settings scope that should have been put in definitions. If the distinction of when to use `ItemInfo` is unclear to you as an expert in this kind of system, then it will almost certainly confuse others who try to use this later. Maybe it's easier to say: "If you can completely derive it from the raw data, hang a model off of `ItemRaw`, and if it's any new data entered in by a user, hang it off of `ItemVersion`."?

projects/dev.py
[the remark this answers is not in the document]
ormsbee: Okay, I'll give that a shot. I do like not having `apps` appear twice in that path (package + apps.py). Thank you!

openedx_learning/apps/core/itemstore/models.py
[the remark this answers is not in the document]
ormsbee: I'd prefer not to have the full learning context key encoded, but since we currently allow blocks of different types to have the same ID, I think we have to encode at least something like `video+xyz`

ormsbee:
FYI @bradenmacdonald: This is the data models PR I briefly mentioned in our meeting today.

openedx_learning/apps/core/itemstore/models.py
kdmccormick: I just realized I need to check an assumption about how I've been reading this: "hanging model X off of model Y" means that X would have a foreign key to Y, right?

openedx_learning/apps/core/itemstore/models.py
bradenmacdonald: What about making `published_version` a [memoized] python getter that uses the publish app to return the published version? Or do you mean specifically you want to be able to use the django ORM methods on this `published_version` field so you can `select_related` and do things like filtering queries?

openedx_learning/apps/core/itemstore/models.py
bradenmacdonald: Should this be `title_override` then to clarify that you can optionally stick with whatever the library author gave as a title?

openedx_learning/apps/core/itemstore/models.py
ormsbee: > Or do you mean specifically you want to be able to use the django ORM methods on this published_version field so you can select_related and do things like filtering queries?

Yeah, exactly this. I want to make sure we can still do efficient filtering queries. But I think I'll actually leave this out for now and add it in later if it becomes necessary. Even if we do keep this kind of a key, it might make more sense in a separate model that specifically separates the publishing details (`published_version`, `first_published_at`, `last_published_at`, etc.).

openedx_learning/apps/core/itemstore/models.py
ormsbee: I'm inclined to not use the `_override` convention because the default content in the library would also be using this field (just under the content library's learning context).

Though maybe the title should be lifted out into a separate model? Maybe even one with language awareness?

Wait... maybe ItemRaw:ItemVersion is M:1?

I'll make comments along both of those lines for future followup, but won't make any immediate model changes.

kdmccormick:
Hey @ormsbee , are you looking for another round of review on this one?

ormsbee:
@kdmccormick: Almost. I've been doing bits and pieces in the odd-hours for this. I need to update the import script to make sure these ideas all hold together and then it'll be ready for review. Though I don't expect the models.py files to change very much at this point, so if you want to take another pass at that, you're more than welcome to.

ormsbee:
@kdmccormick: ready for another round of review, with the caveat that I'd ask that you largely ignore `load_course_data.py`  and the `staticassets` app at this point. I wanted to make sure it was functional (i.e. the data model wasn't totally broken), but the interface is completely wrong for the import script and `staticassets` is pretty half-baked at the moment.

openedx_learning/core/itemstore/models.py
```diff
@@ -0,0 +1,238 @@
...
+
+    learning_context = models.ForeignKey(LearningContext, on_delete=models.CASCADE)
+    hash_digest = hash_field()
+
+    # Per RFC 4288, MIME type and sub-type may each be 127 chars. Add one more
+    # char for the '/' in the middle, and we're at 255.
+    #
+    # TODO: maybe we should split it up into two fields instead of doing
+    #       substring searches? 
+    mime_type = models.CharField(max_length=255, blank=False, null=False)
```
kdmccormick, line 176: > maybe we should split it up into two fields instead of doing substring searches? 

Sounds like a win to me. Any downsides?

Another reason to split: [MIME types can include a `parameter` field](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#structure_of_a_mime_type), eg `text/plain;charset=utf8`. If you don't want to support the parameter field, then splitting into two fields would make that more explicit.

openedx_learning/core/itemstore/models_api.py
```diff
@@ -0,0 +1,21 @@
...
+
+from .models import ItemVersion
+
+
+class ItemVersionDataMixin(models.Model):
+    """
+    Minimal abstract model to let people attach data to ItemVersions.
+
+    The idea is that if you have a model that is associated with a specific
+    version of an item, the join is going to be 1:1 with an ItemVersion, and
```
kdmccormick, line 11: So, I'm reading that:
* each ItemVersion can only be associated with at most one of each type of "Data" model (Image, Asset, Video, etc.), but
* each ItemVersion may be associated with an arbitrary number of Content objects.

This confused me at first, because the "Data" models and the Content model occupy very similar roles in my head. But I think I'm beginning to understand it... is this a good example?
* Some ItemVersion represents a video.
  * It has multiple Content items associated with: one that points to the video file, one PNG thumbnail, and several others text files that are transcripts.
  * In the staticassets app, it has a single associated Video instance and Asset instance. Video metadata as well as information about the transcripts are part of (or hang off) of the Video model.

openedx_learning/core/itemstore/models_api.py
```diff
@@ -0,0 +1,21 @@
...
+    potentially M:1 with your data model.
```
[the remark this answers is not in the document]
ormsbee: > is Data immutable

It should be or Bad Things will happen.

> is ItemVersionData:Data an M:1 relationship because multiple versions of an Item may reference the same underlying Data?

Yes. One of the underlying assumptions here is that if you have some data that's associated with an ItemVersion, changing that data means that the Item has changed in some way, and thus forms a new ItemVersion. So if you change the retry-settings or some such for an Item, there's a new ItemVersion. But for all the other apps that didn't change their data, we can just make new entries that join the new ItemVersion with the same data that those apps had for the last ItemVersion.

openedx_learning/core/itemstore/models.py
ormsbee: I guess I'm just used to thinking them as a combined thing because I can't really see querying one in isolation from the other. But it is more normalized and it's easy to add, so 🤷 I'll add it.

openedx_learning/core/itemstore/models_api.py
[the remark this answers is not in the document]
kdmccormick: @ormsbee 

> It should be or Bad Things will happen.

Cool, I assumed so but wanted to be sure. Is there a way to indicate that, like marking it read-only in Django admin?

> Yes. ...

Makes sense 👍🏻 It would be good to include that explanation somewhere in the library before releasing v1.0. Not a blocker for this PR; I know you're still iterating.

openedx_learning/core/itemstore/models.py
```diff
@@ -0,0 +1,238 @@
...
+    ItemVersion in the current LearningContextVersion (i.e. this content was at
+    some point removed from the "published" version).
+
+    An Item belongs to one and only one LearningContext.
+
+    The UUID should be treated as immutable. The identifier field *is* mutable,
+    but changing it will affect all ItemVersions.
+    """
+    uuid = immutable_uuid_field()
+    learning_context = models.ForeignKey(LearningContext, on_delete=models.CASCADE)
```
feanil, line 38: So an item can only be in one `LearningContext`(LC), but since an `ItemVersion` can be in any number of `LearningContextVersions`(LCV) do we need a constraint somewhere that ensures that all those LCVs are from the same LC?  Or is that too paranoid at this stage?

openedx_learning/core/itemstore/models.py
```diff
@@ -0,0 +1,238 @@
...
+    MAX_SIZE = 10_000_000
+
+    learning_context = models.ForeignKey(LearningContext, on_delete=models.CASCADE)
+    hash_digest = hash_field()
+
+    # Per RFC 4288, MIME type and sub-type may each be 127 chars. Add one more
+    # char for the '/' in the middle, and we're at 255.
+    #
+    # TODO: maybe we should split it up into two fields instead of doing
+    #       substring searches?
```
feanil, line 175: When would we need to do a partial search? Is it just when looking at types and ignoring sub-types? or something more complex?

feanil (approved):
A couple of questions but nothing blocking this from landing.  The re-organization generally makes sense to me based on previous conversations.

openedx_learning/core/itemstore/models_api.py
ormsbee: Yeah, that was my intent. A couple of extra notes about it:

### Content survives after apps die.

At some point, an app becomes unsupported and disappears. At that point, all the Item/ItemVersion-related data it has dies as well. But Content is associated with the ItemVersion directly in `itemstore`, so that raw data will still export.

So I think this encourages a pattern where we store raw things in Content and then build smarter relational data structures out of them connected to ItemVersion. For instance, a ProblemBlock's raw OLX can live in Content while richer relational tables live in a new ProblemBlock app that give a richer, more queryable interface.

### ItemVersion mapping to just one row in a "Data" model is common, but not required.

I think that most of the time apps want to associate one piece of data with a given ItemVersion (e.g. a problem weight, schedule, etc.), but that's not a requirement. A rich data model of a ProblemBlock might want to create entries for each response, for instance.

openedx_learning/core/itemstore/models.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    A particular version of an Item.
+
+    A new ItemVersion should be created anytime there is either a change to the
+    content or a change to the policy around a piece of content (e.g. schedule
+    change).
+    
+    Each ItemVersion belongs to one and only one Item.
+
+    TODO: created_by field?
```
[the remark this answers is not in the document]
ormsbee: Yeah, that was my thought as well.

openedx_learning/core/itemstore/models.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    uuid = immutable_uuid_field()
+
+    item = models.ForeignKey(Item, on_delete=models.CASCADE)
+
+    # Question: Title is the only thing here that actually has human-readable
+    # text. Does it make sense to lift it out into a separate metadata model,
+    # possibly even one with language awareness?
```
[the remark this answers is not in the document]
ormsbee: Yeah, I think you're right that it belongs in a higher level. In any case, I'm definitely fine with punting this for now, even if we do bake it at this layer later. I'll remove the comment.

openedx_learning/core/itemstore/models.py
ormsbee: Yeah, that's the only situation I can really think of.

openedx_learning/core/itemstore/models.py
ormsbee: I like being paranoid about the data model in general, but I'm not sure what we can do with MySQL here, or how much overhead adding CheckConstraints is going to cause. I'll see how hard it is.

openedx_learning/core/itemstore/models_api.py
kdmccormick: > ItemVersion mapping to just one row in a "Data" model is common, but not required.

How would that look?

For example, imagine we have an ItemVersion representing a problem block, and Response data models that we want to (indirectly) hang off of it. Would it look like:

```
ItemVersion  <--(1:1)-- ItemVersionProblemBlock --(N:N)--> Response
```
or:
```
ItemVersion  <--(1:N)-- ItemVersionResponse --(N:1)--> Response
```
or something else?

ormsbee: I guess I thought it would be like:

```
ItemVersion <-- (N:1) -- ItemVersionProblemBlock -- (N:N) --> Response
```

That would still allow some central process to increment to the next version in a simple way without understanding what ItemVersionProblemBlock is. But I honestly haven't thought it through that much.

openedx_learning/core/itemstore/models_api.py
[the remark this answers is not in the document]
ormsbee: > Cool, I assumed so but wanted to be sure. Is there a way to indicate that, like marking it read-only in Django admin?

Yeah, I'll definitely do that when I start making some admin pages.

PR #17, opened 2022-08-03
chore: run make upgrade

feanil:
The version of pip-tools that was previously pinned breaks
with the latest version of pip.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -4,19 +4,19 @@
 #
 #    make upgrade
 #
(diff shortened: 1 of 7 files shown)


ormsbee (approved):
FWIW, I have another PR that removes a couple of these dependencies entirely (django-simple-history and django-model-utils). I've got a few mini-PRs along those lines that I planned to open up after #13 merges. No point in holding this off in the meantime though.

PR #19, opened 2022-08-07
Remove dependencies, add import-linting

ormsbee:
I've combined two commits into one PR because they both update dependencies:

```
feat: add import linting

import-linter is meant to help us keep our internal dependencies clear,
and prevent accidental circular dependencies or other unwanted code
relationships.

You can run the linter with the command: lint-imports
```

```
chore: remove unnecessary dependencies

The standards cookiecutter for Open edX projects includes
django-model-utils and django-simple-history. This repo doesn't really
need them because:

1. django-model-utils is only really used for TimeStampedModel, and that
   is just to get consistent naming of created/modified fields. We can
   do that without taking on the entire dependency.
2. django-simple-history offers a convenient way to make a history
   table, and we do have versioned models in many cases. However, our
   versioning needs are very explicit and particular, and they go beyond
   the simple changelog-style history that this library provides.

In general, we want to keep the dependencies in this repo as small as
possible, to make it easy to use in different projects.
```

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -1,0 +1,45 @@
+# openedx_learning is intended to be a library of apps used across multiple
+# projects, and we want to ensure certain dependency relationships. Please
+# think through any changes you make to this file carefully, and don't just
+# casually modify these linting rules to "fix the build".
+[importlinter]
+root_package = openedx_learning
+
+# This is the most basic layering for openedx_learning.
+#
+# The "lib" package is meant for low level utilities, field definitions, and the
+# like, so it's at the bottom layer.
+#
+# The "core" apps are meant to be the heart of our system, with foundational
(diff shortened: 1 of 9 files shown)


.importlinter
```diff
@@ -0,0 +1,45 @@
...
+name = Lib / Core / Contrib Layering
+type = layers
+layers=
+    openedx_learning.contrib
+    openedx_learning.core
+    openedx_learning.lib
+
+# This is layering within our Core apps.
+#
+# The lowest layer is "publishing", which holds the basic primitives needed to
```
kdmccormick, line 31: In addition to the correctness guarantees, this file provides some nice arch documentation 😄

kdmccormick (approved):
🔥

PR #25, opened 2022-09-16
More Prototyping

ormsbee:
Long and winding PR as I iterate through a few approaches, switch around the data models to reflect the Item -> Component terminology from various conversations, and put in a minimal loading script and API for the new data structures.

It's a mess of ideas, and not in a reviewable state. Likely will be extracted out into pieces for the real PR later. Published here just so people can see what I'm iterating through.


ormsbee:
Closing this for now, in favor of #28, but I may go back to this PR to grab bits of code later on.

PR #41, opened 2023-04-22
Publishing API with centralized publishing models

ormsbee:
- chore: Switch from edx-sphinx-theme to sphinx-book-theme
- docs: ADR for tagging service. (#40)
- feat: add support for static assets
- chore: add DRF, remove codecov, rebuild requirements
- feat: add publishing API


ormsbee:
Making the barest version of the APIs first (no real error handling). Will flesh out after I have the basic structure of the loading script converted to only make API calls.

ormsbee:
@bradenmacdonald, @kdmccormick, @feanil: If you have time, could you please take a look at the `model.py` files, particularly for the `publishing` and `components` apps (the `contents` app just has `RawContent` and `TextContent` extracted out of `components` without real modifications).

The API layer is a mess right now, and probably not worth looking at. I wanted to get it to a point where `load_components.py` could do all its work without manipulating the models directly. And that does work, but there's no proper error handling and the function signatures are inconsistent. I probably need another day of hacking at it to get that part into a reviewable state, and I need to step away from this for a few days. But I could use early feedback on whether the data model looks reasonable.

Thanks folks.

.importlinter
```diff
@@ -41,4 +41,5 @@ name = Core App Dependency Layering
 type = layers
 layers=
     openedx_learning.core.components
+    openedx_learning.core.contents
     openedx_learning.core.publishing
```
kdmccormick, line 45: I love this--it conveys so much about the relationship between the apps in one glance.

kdmccormick (commented):
I think the data model and how you've split it between the three apps looks great. All I have is a question about LearningPackage and some docstring nitpicks.

kdmccormick (commented):
OK, I though of real critique:

Django has built-in support for multi-table inheritance: https://docs.djangoproject.com/en/3.2/topics/db/models/#multi-table-inheritance

What do you think about using that instead of the the PublishableEntity{Version} / PublishableEntity{Version}Mixin combo? I think the resulting data model would be identical, but it'd be more idiomatic Django.

ormsbee:
@kdmccormick: That's a good point. I think I was turned off of multi-table inheritance years ago because of how strongly certain books (like Two Scoops) advise against it. My understanding was that it did a lot of surprising left outer joins. But that was years and years ago, I haven't really tested it, and you're right in that it looks like a good fit as long as those joins aren't forced onto the base class (i.e. the `publishing` app can still do its thing without ever joining on sub-class tables).

ormsbee:
@kdmccormick: Thinking on it a bit more, the reverse relations also become a bit weirder. For instance, in a non-inheritance model, the `versions` relation from `PublishableEntity` refers to `PublishableEntityVersion`, and `Component` has its own `versions` relation set to `ComponentVersion`. But if `Component` inherits from `PublishableEntity`, then `component.versions` would map to `PublishableEntityVersion`.

kdmccormick:
@ormsbee If that's the main issue, then we could avoid it by calling the generic reverse relation `PublishableEntity.entity_versions`, which would allow the subclasses define `.versions`. So you'd have:
```
component.versions              # ComponentVersions
component.publishableentity_ptr # PublishableEntity
component.entity_versions       # PublishableEntityVersions
```

>  I think I was turned off of multi-table inheritance years ago because of how strongly certain books (like Two Scoops) advise against it. My understanding was that it did a lot of surprising left outer joins. But that was years and years ago, I haven't really tested it, and you're right in that it looks like a good fit as long as those joins aren't forced onto the base class

That's good to know; I think I've only ever used it once, and it wasn't in a situation where performance was top-of-mind. If you think using Django's inhertance would lead to more surprising queries then I am OK with the hand-rolled version you have now, which at the very least is easy for us to reason about.

openedx_learning/core/publishing/models.py
```diff
@@ -13,36 +24,258 @@
...
+
+    Identifiers
+    -----------
+    The UUID is globally unique and should be treated as immutable.
+
+    The identifier field *is* mutable, but changing it will affect all
+    PublishedEntityVersions. They are locally unique within the LearningPackage.
+
+    If you are referencing this model from within the same process, use a
+    foreign key to the id. If you are referencing this PublishedEntity from an
```
bradenmacdonald, line 88: `id` is different from `identifier` right? Seems like potential for confusion. Could we at least list it below as an explicit model field instead of implicit?

openedx_learning/core/publishing/models.py
```diff
@@ -13,36 +24,258 @@
...
+
+class PublishableEntityVersion(models.Model):
+    """
+    A particular version of a PublishableEntity.
+
+    This model has its own ``uuid`` so that it can be referenced directly. The
+    ``uuid`` should be treated as immutable.
+
+    PublishableEntityVersions are created once and never updated. So for
+    instance, the ``title`` should never be modified.
```
bradenmacdonald, line 153: So if I make 100 small edits (to a draft) without publishing, there will still be 100 `PublishableEntityVersion` rows created, but just one `Draft` row which gets updated each time, and no `PublishLog[Record]`/`Published` rows. Is that right?

openedx_learning/core/publishing/models.py
```diff
@@ -13,36 +24,258 @@
 
 
 class LearningPackage(models.Model):
+    """
+    Top level container for a grouping of authored content.
+
+    Each PublishableEntity belongs to exactly one LearningPackage.
+    """
+
```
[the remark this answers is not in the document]
bradenmacdonald: I had the same question. I guess it's about namespacing identifiers? 

I have the vague impression that the `publishing` app is not as generic as it could be when it's tied to the concept of LearningPackage. But maybe I'm just not clear on the role of LearningPackage.

bradenmacdonald (commented):
Makes sense overall, though my vague gut feeling is that the `publishing` app seems to own too much logic about how things are grouped (LearningPackage) and identified (`id`, `uuid`, and `identifier` are part of `PublishableEntity`).

For example, does the `publishing` app even use the UUID field? If not, why isn't it up to each publishable model like `Component` to decide if it wants a UUID or not? I feel like it would be nice if the Publishable models/mixins could be configured to use any sort of namespacing/identifier scheme that the Publishables themselves want to use. But perhaps that adds too much complexity or abstraction.

openedx_learning/core/publishing/models.py
kdmccormick: How set are we on the term "identifier" @ormsbee @bradenmacdonald  ? Given that we have two other fields with ID in the name (`id` and `uuid`), and given that we recommend that folks *don't* rely on the field:
> The identifier is the part that is
    most likely to be human-readable, and may be exported/copied, but try not to
    rely on it, since this value may change.

it feels less like an `identifier` to me, and more like `slug`, `anchor`, or `key`.

ormsbee:
@kdmccormick:

> @ormsbee If that's the main issue, then we could avoid it by calling the generic reverse relation `PublishableEntity.entity_versions`, which would allow the subclasses define `.versions`. So you'd have:
> 
> ```
> component.versions              # ComponentVersions
> component.publishableentity_ptr # PublishableEntity
> component.entity_versions       # PublishableEntityVersions
> ```

Right, I realize we can rename things to avoid the collision in this case. I meant the general notion that inheritance will conflate relations between things in the `publishing` app and things in the apps that build on it since we'll be dragging everything into the same namespace, and I'm not sure:

* Is it worth it for the ergonomics/convenience?
* Is it worth making the field names clunkier on the `publishing` side?
* Will people shoot themselves in the foot with this somehow, because we'll add a relation to a `publishing` model at some point that breaks a subclass?
* Is subclassing better than making a mixin in terms of improving ergonomics?

I feel like the most intuitive thing for developers would be to have something where they can say:

```python
component.published_version  # This should be a ComponentVersion
component.draft_version  # This should be a ComponentVersion
```

So something where the part of the API that content apps are exposed to runs through publishing's models but spits back their own models, so they're not actually exposed to any of the publishing details. But that means making some little helper method (either inherited or a mixin) either way. Given that, I'm leaning more towards clunky-but-less-coupled (i.e. no inheritance). But I'll experiment with it some.

openedx_learning/core/publishing/models.py
ormsbee: I was borrowing the `identifier` terminology from QTI. I think I need to qualify that recommendation more. I think it's more like saying "the decision to rely on the `identifier` field is something that can be made at a higher level, but our apps don't assume it's stable"? So for us, UsageKeys (or something we derive UsageKeys from, since I don't want to hard-code the course key into it) would go here. Whether its human readable depends on the higher level system–some of our UsageKeys are, most aren't. But we want to make foreign keys in such a way that changing a UsageKey/CourseKey won't break all the things like it does today.

I think that `slug` is very specifically meant to be human-readable, and doesn't quite strike the right connotation here. Maybe `key`? I haven't really heard of `anchor` being used for this kind of thing outside of HTML.

bradenmacdonald: I like `key`

kdmccormick:
@ormsbee Thanks for considering it. When it comes to the question:

> Is it worth it for the ergonomics/convenience?

you've been actually working in this codebase and I haven't, so I think you have a much better intuition for what will feel 
good for developers using this API :)

> I feel like the most intuitive thing for developers would be to have something where they can say:
>
>      component.published_version  # This should be a ComponentVersion
>      component.draft_version  # This should be a ComponentVersion

I agree, that does look very intuitive.

It would be a change to your data model, but I think using [generic relations](https://docs.djangoproject.com/en/4.2/ref/contrib/contenttypes/#generic-relations) you could achieve something like this? Some pseudo-Django:
```python
class Versioning(Model):
    """
    Links versionless things to their draft and published versions.
    """
    # The pair of concrete types. Component/ComponentVersion, Unit/UnitVersion, etc.
    verionless_type = ForeignKey(ContentType)
    versioned_type = ForeignKey(ContentType)

    # The concrete versionless entity. An instance of a Component, Unit, etc.
    versionless_id = PositiveIntegerField()
    versionless = GenericForeignKey('verionless_type', 'versionless_id')

    # The concrete published and draft entity versions. An instance of ComponentVersion, UnitVersion, etc.
    published_id = PositiveIntegerField()
    published = GenericForeignKey('versioned_type', 'published_id')
    draft_id = PositiveIntegerField()
    draft = GenericForeignKey('versioned_type', 'draft_id')    
```

For developers using components, it'd look like:
```python
component.versioning.published  # a ComponentVersion
component.versioning.draft      # a ComponentVersion

# or, starting with a draft component:
draft.versioning.published # a ComponentVersion
draft.versioning.versionless  # a Component, equivalent to `draft.component`
```

ormsbee:
> Makes sense overall, though my vague gut feeling is that the `publishing` app seems to own too much logic about how things are grouped (LearningPackage) and identified (`id`, `uuid`, and `identifier` are part of `PublishableEntity`).
> 
> For example, does the `publishing` app even use the UUID field? If not, why isn't it up to each publishable model like `Component` to decide if it wants a UUID or not? I feel like it would be nice if the Publishable models/mixins could be configured to use any sort of namespacing/identifier scheme that the Publishables themselves want to use. But perhaps that adds too much complexity or abstraction.

As an aside, `publishing` would use the UUID directly if there was a proper REST API hooked up to it.

FWIW, this was something I struggled with in the publishing/components split, especially since Component already has its own tuple of `(namespace, type, local_identifier)`. At the end of the day, I feel like this is a place where we've gotten tripped up enough in the past that I wanted to take a strong baseline position for all publishable things–it will have _at least_ these three id-related things, and you're free to add more to your own models if necessary. We have other services, and those services will want to store references to content and versions of content, and having the UUID means that they can do so in a way that still gives us some chance of changing CourseKeys and the like some day.

I also did want a LearningPackage-wide identifier that was app-chosen and locally unique, so that there was a common way to look things up within a LearningPackage.

Honestly, the place where I feel that I'm on the shakiest ground is UUIDs for Versions of things because with the way versions work right now, every version is uniquely identifiable by a combination of PublishableEntity UUID + version_num. I have a possibly irrational dislike of compound-ids, and I wasn't sure about the version_num thing in the beginning, though I'm more committed to it now. I'm thinking of taking out the UUIDs on EntityVersion largely because I'm not sure how we'd handle them properly in import/export. (At least with Entities, we could export a mapping file of UUIDs to identifiers.)

ormsbee:
I don't know why this only occurs to me now, but a Grading Policy would be a totally reasonable thing to model as a `PublishableEntity`.

ormsbee:
Which also leads me to weird places. Would we ever have a library of grading policies and taxonomies? 😛

openedx_learning/core/publishing/models.py
ormsbee: Changed it to `key`.

openedx_learning/core/components/models.py
```diff
@@ -47,33 +42,37 @@ class Component(models.Model):
...
     """
 
-    uuid = immutable_uuid_field()
+    # This foreign key is technically redundant because we're already locked to
+    # a single LearningPackage through our publishable_entity relation. However, having
+    # this foreign key directly allows us to make indexes that efficiently
+    # query by other Component fields within a given LearningPackage, which is
+    # going to be a common use case (and we can't make a compound index using
+    # columns from different tables).
     learning_package = models.ForeignKey(LearningPackage, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
ormsbee: Added this, thank you!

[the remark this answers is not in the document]
ormsbee: Actually, nvm, this breaks:

```
❯ python manage.py migrate
SystemCheckError: System check identified some issues:

ERRORS:
oel_components.Component: (models.E041) 'constraints' refers to the joined field 'publishable_entity__learning_package'.
```

tests/openedx_learning/core/components/test_models.py
```diff
@@ -0,0 +1,43 @@
...
+        assert component.versioning.draft == component_version
+        assert component.versioning.published is None
+        publish_all_drafts(self.learning_package.pk, published_at=self.now)
+
+        # Force the re-fetch from the database
+        component = get_component_by_pk(component.pk)
+        assert component.versioning.published == component_version
+
+        # Grabbing the list of versions for this component
+        assert list(component.versioning.versions) == [component_version]
```
ormsbee, line 43: @kdmccormick: the `versions` helper interface that you suggested (though it's implemented by hand through existing model relations instead of via generic foreign keys).

ormsbee:
@kdmccormick, @bradenmacdonald, @feanil: Made updates to the models.

@kdmccormick: I ended up going for the non-inheritance approach, with a helper in the [mixin](https://github.com/openedx/openedx-learning/pull/41/files#diff-e1ff3c42e7b1471c651cbe692125f9a99cc727d031f6747b27fba23558456b83R36-R172). I tried it with generic relations and it's workable, but I felt like it added fields that weren't really necessary and I worried that it would encourage a pattern where publishing starts to know more about the specific things being published.

I'm going to fix up the APIs more, but I think I'd like to just land them in a rough shape. They'll probably need a few followup PRs to refine them, but I want to land the data models in a solid enough shape where we can start putting actual data into them (i.e stop my "blow away the world and remake the migrations" approach), and I think this PR will achieve that. The API modules at this point are mostly just to have a convenient proof-of-concept to dump stuff in and out of those models.

openedx_learning/core/publishing/models.py
```diff
@@ -1,48 +1,340 @@
...
+    created. This means there are three possible states:
 
-    It is expected that other apps make foreign keys to this table to mark when
-    their content gets published. This is to allow us to tie together many
-    different entities (e.g. Components, Units, etc.) that are all published at
-    the same time.
+    1. No Draft entry for a PublishableEntity: This means a PublishableEntity
+       was created, but no PublishableEntityVersion was ever made for it, so
+       there was never a Draft version.
+    2. A Draft entry exists and points to a PublishableEntityVersion: This is
```
feanil, line 299: Is this the state when the `PublishableEntityVersion` is also the latest published one?

feanil (approved):
I looked just at the publishing models and they make sense to me.  I had some questions/clarification requests for docstrings but nothing blocking.

openedx_learning/core/publishing/models.py
```diff
@@ -1,48 +1,340 @@
...
+    If you are referencing this model from within the same process, use a
+    foreign key to the id. If you are referencing this PublishedEntity from an
+    external system/service, use the UUID. The key is the part that is most
+    likely to be human-readable, and may be exported/copied, but try not to rely
+    on it, since this value may change.
+
+    Note: When we actually implement the ability to change identifiers, we
+    should make a history table and a modified attribute on this model.
+
+    Why are Identifiers in this Model?
```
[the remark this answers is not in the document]
ormsbee: Fair point. Making a short one around identifiers now. I'll punt others until later.

openedx_learning/core/publishing/apps.py
```diff
@@ -13,3 +13,4 @@ class PublishingConfig(AppConfig):
     name = "openedx_learning.core.publishing"
     verbose_name = "Learning Core: Publishing"
     default_auto_field = "django.db.models.BigAutoField"
+    label = "oel_publishing"
```
ormsbee, line 16: Flagging this for reviewers: does this sort of prefix label sound okay? I'm a little worried about namespace collision with app names in the LMS and potentially elsewhere (since the Django app namespace is flat), and I thought having a repo-wide prefix would help. It does mean that it's a little less intuitive when you do things like `makemigrations`, since the app will now be `oel_publishing` instead of `publishing`.

openedx_learning/core/publishing/model_mixins.py
```diff
@@ -0,0 +1,215 @@
...
+
+            # This will work, because it's never been published
+            assert component.versioning.published is None
+
+            # Publishing happens
+            publish_all_drafts(learning_package.id, published_at=now)
+
+            # This will FAIL because it's going to use the relation value
+            # cached on component instead of going to the database again.
+            # You need to re-fetch the component for this to work.
```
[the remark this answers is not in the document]
ormsbee: I'll add it, but basically you have to refetch it from the database. I'm hesitant to build anything smart around that because that's more of a Django ORM thing.

openedx_learning/core/publishing/models.py
ormsbee: Not necessarily. The Draft entry will always point to the most recently created `PublishableEntityVersion`. That may or may not be the published version–that depends on the entry in the `Published` model.

openedx_learning/core/publishing/model_mixins.py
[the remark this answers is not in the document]
kdmccormick: @ormsbee Did you consider the collapsing `VersioningHelper` directly into the `PublishableEntityMixin` class?

In addition to simplifying the class structure, having `component.published` instead of `component.versioning.published` might be more immediately obvious to devs that `.published` will not automatically be updated without a re-fetch.

If I pretend I don't know this codebase, the `.versioning.` attribute in the middle kinda hints "magic happens here", even though the it's really just a syntactic helper construct.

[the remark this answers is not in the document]
ormsbee: > @ormsbee Did you consider the collapsing `VersioningHelper` directly into the `PublishableEntityMixin` class?

I did. In fact, my original comment was proposing that. But I liked your suggestion of putting it behind `versioning` better, because I expect that a bunch of other stuff will make it into this class over time.

> In addition to simplifying the class structure, having `component.published` instead of `component.versioning.published` might be more immediately obvious to devs that `.published` will not automatically be updated without a re-fetch.
>
> If I pretend I don't know this codebase, the `.versioning.` attribute in the middle kinda hints "magic happens here", even though the it's really just a syntactic helper construct.

I mean, it is a _little_ magical, right? At least when compared to the simpler properties that just proxy to fields in the accompanying `publisher` app models (like `uuid`). But it's relatively small magic.

Maybe I should just change my implementation to always refetch. That's probably the more intuitive thing anyway, right?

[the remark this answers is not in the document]
kdmccormick: Gotcha. Sorry for the suggestion whiplash. When I suggested the `.versioning.` interface, I was thinking of it as an alternative to `PublishableEntity{Version}Mixin`. Since we're sticking with the mixins, I could go either way on `.versioning`. 

I think:
* `component.published` implies no re-fetch.
* `component.versioning.published` could go either way.
* `component.get_versioning().published` would heavily imply re-fetch.

openedx_learning/core/publishing/model_mixins.py
```diff
@@ -0,0 +1,264 @@
...
+            Return the content version object that is the current draft.
+
+            This will not force re-fetches from the database. Please read the
+            "Caching Warning" in the class docstring for more details.
+            """
+            pub_ent = self.content_obj.publishable_entity
+
+            # If no Draft object exists at all, then no PublishableEntityVersion was
+            # ever created (just the PublishableEntity)
+            if not hasattr(pub_ent, "draft"):
```
kdmccormick, line 141: Sorry, I think I'm confused, how does the `draft` attribute get assigned on the PublishableEntity instance?

openedx_learning/core/publishing/api.py
```diff
@@ -0,0 +1,192 @@
...
+        created=created,
+        updated=created,
+    )
+    package.full_clean()
+    package.save()
+
+    return package
+
+
+def create_publishable_entity(learning_package_id, key, created, created_by):
```
kdmccormick, line 52: As a general comment: The class hierarchy in openedx-learning is becoming non-trivial. That's not suprising, since we're tackling a complex problem, but having type annotations on all of these API functions could make it much easier to grok. I know you're still iterating on the APIs, so don't consider this blocking.

kdmccormick (commented):
Looking good, just a couple questions.

FYI, I paid closer attention to the models and model_mixins than I did the apis or load_components.

openedx_learning/core/publishing/apps.py
kdmccormick: Looks reasonable to me.

You could be super explicit and do `openedx_learning_publishing`. If that feels too verbose, though, then `oel_publishing` is good.

openedx_learning/core/publishing/model_mixins.py
ormsbee: That's the OneToOneField relation, but I replaced it with a straight query on the `Draft` model, which forces the refresh and is probably a lot clearer.

openedx_learning/core/publishing/api.py
ormsbee: Agreed, all api.py functions should have annotations.

ormsbee: I'll add those annotations when I add the action that actually checks them in CI. 😞 I need to do some care and feeding for the CI in this next week.

kdmccormick: 👍🏻

openedx_learning/core/publishing/api.py
```diff
@@ -0,0 +1,192 @@
...
+            published_at=published_at,
+            published_by=published_by,
+        )
+        publish_log.full_clean()
+        publish_log.save(force_insert=True)
+
+        for draft in draft_qset.select_related("entity__published"):
+            # There might not be a currently published version for this
+            # PublishableEntity yet.
+            if hasattr(draft.entity, "published"):
```
kdmccormick, line 141: @ormsbee sorry, I'm confused about this too: how does `published` get assigned on the PublishableEntity instance? Why is it not always available on the instance? I would have expected this check to look like:
```python
if draft.entity.published:
```

ormsbee: The `Published` model has the OneToOneField pointing to `PublishableEntity`. So when you're coming through the reverse relation (`entity.published`), there's a chance that there is no row in `Published`. When that happens, Django will throw a `ObjectDoesNotExist` if you try to access it, and they recommend checking with `hasattr`:

https://docs.djangoproject.com/en/3.2/topics/db/examples/one_to_one/

openedx_learning/core/publishing/api.py
```diff
@@ -0,0 +1,192 @@
...
+        def ready(self):
+            from ..publishing.api import register_content_models
+            from .models import Component, ComponentVersion
+
+            register_content_models(Component, ComponentVersion)
+
+    There may be a more clever way to introspect this information from the model
+    metadata, but this is simple and explicit.
+    """
+    return PublishableContentModelRegistry.register(
```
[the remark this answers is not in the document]
ormsbee: Added `issubclass` checks. Now it'll fail startup if it's improperly configured, with an error like:

```
raise ImproperlyConfigured(
django.core.exceptions.ImproperlyConfigured: <class 'openedx_learning.core.components.models.ComponentVersion'> must inherit from PublishableEntityMixin
```

openedx_learning/core/publishing/model_mixins.py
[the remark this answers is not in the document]
ormsbee: I'm going to stick with `versioning` for now. It is a bit magical, but I think it's more consistent with Django ORM conventions.

openedx_learning/core/publishing/api.py
ormsbee: Done.

ormsbee:
@kdmccormick: I think I've addressed all your comments and suggestions.

openedx_learning/core/publishing/model_mixins.py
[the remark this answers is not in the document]
kdmccormick: 👍🏻

openedx_learning/core/publishing/api.py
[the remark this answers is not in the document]
kdmccormick: 👌🏻

kdmccormick (approved):
🚀🚀🚀

PR #52, opened 2023-05-18
fix import linter comments

ormsbee:
Just fixing some outdated messaging and re-formatting for later readability.


ormsbee:
@bradenmacdonald, @kdmccormick, @feanil: Tiny one when you have a moment.

kdmccormick (approved):
😎

PR #54, opened 2023-05-28
feat: Tagging Django App setup 

ChrisChV:
In this PR, the tagging application and a base model for the `TagContent` have been created.

**Testing instructions**

- Read the code
- Run a server and verify that all is fine.


openedx-webhooks:

- supporting documentation

projects/dev.py
```diff
@@ -35,6 +35,7 @@
     "openedx_learning.core.publishing.apps.PublishingConfig",
     # Learning Contrib Apps
     "openedx_learning.contrib.media_server.apps.MediaServerConfig",
+    "openedx_learning.contrib.tagging.apps.TaggingConfig",
```
pomegranited, line 38: Hmm.. this tagging app is supposed to be a "plugin", which to me means:

1. it's installed alongside [edx-django-utils](https://github.com/openedx/edx-django-utils/) whose [plugin utils](https://github.com/openedx/edx-django-utils/blob/master/edx_django_utils/plugins/README.rst) which ensures that it gets automatically added to the `INSTALLED_APPS` list in the right context
1. it can be installed separately into apps where needed, and

edx-django-utils is specifically tailored to the `lms` and `cms` apps in edx-platform, and doesn't have an equivalent on this app, so..

Maybe we need to deal with the "pluggability" aspect when it comes time to extract this app out for reuse outside of openedx-learning?

openedx_learning/contrib/tagging/readme.rst
```diff
@@ -0,0 +1,26 @@
...
+of flexible content structures different from the current implementation and
+allow adaptive learning in the Open edX platform.
+
+This service has been implemented as pluggable django app. Since it is necessary for
+it to work independently of the content to which it links to.
+
+Setup
+---------
+
+**TODO:** We need to wait the discussion of the [Taxonomy discovery](https://docs.google.com/document/d/13zfsGDfomSTCp_G-_ncevQHAb4Y4UW0d_6N8R2PdHlA/edit#heading=h.o6fm1hktwp7b)
```
pomegranited, line 20: nit: [Links in RST](https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Content/Links.html) don't look like Markdown:

```suggestion
**TODO:** We need to wait the discussion of the `Taxonomy discovery <https://docs.google.com/document/d/13zfsGDfomSTCp_G-_ncevQHAb4Y4UW0d_6N8R2PdHlA/edit#heading=h.o6fm1hktwp7b>`_.
```

pomegranited (approved):
This doesn't do much yet, but it can't really until the data architecture is decided, so :+1:  with one nit, and one question about "pluggability":

- [x] I tested this by creating a virtualenv, running migrations, running a shell, and ensuring I could save a new TagContent object to the database.
- [x] I read through the code
- [x] Includes documentation

openedx_learning/contrib/tagging/readme.rst
ChrisChV: Oh, thanks! My mistake

projects/dev.py
ChrisChV: @pomegranited Good point! I didn't account for `edx-django-utils` for this. I have taken into account that it had to be ["installed alongside learning-core as a dependency in the edx-platform"](https://github.com/openedx/openedx-learning/blob/main/docs/decisions/0004-content-tagging.rst?plain=1#L14)

Another option would be to build this app to be installed alongside edx-django-utils so that we can use them with the current open-edx content and install it as a direct dependency on learning-core when we used the learning-core content. The problem is that I don't know how complicated this approach is in the long run. 

> Maybe we need to deal with the "pluggability" aspect when it comes time to extract this app out for reuse outside of openedx-learning?

I will leave this as an open question

ormsbee: > Maybe we need to deal with the "pluggability" aspect when it comes time to extract this app out for reuse outside of openedx-learning?

Yeah, I think this is something that we can address later. I don't think it needs to be a plugin-utils styled plugin, but more of a library-style app that handles tagging related concerns and is used by our other apps. But I do think this means that the tagging functionality modeling in this repo should be split up so that there is:

1. An app or group of tagging-related apps that is entirely self contained, i.e. has no foreign keys to anything in the `openedx_learning` package. If we're pretty confident that this is going to be its own top level thing that will eventually be used in other parts of the platform, we should give it its own top level package, e.g. `openedx_tagging` (or some other unique name that we can eventually register on PyPI).
2. The `openedx_learning.contrib.tagging` app, which would be where we apply tagging to `openedx_learning` models like `PublishableEntity` and `PublishableEntityVersion`. This would probably stay in this repo over the long term.

ormsbee: > Another option would be to build this app to be installed alongside edx-django-utils so that we can use them with the current open-edx content and install it as a direct dependency on learning-core when we used the learning-core content. The problem is that I don't know how complicated this approach is in the long run.

Didn't see this reply when I wrote my message. I like this approach.

pomegranited: > Another option would be to build this app to be installed alongside edx-django-utils 

I like this approach too, but it has the limitation that edx-django-utils is specifically tailored to edx-platform -- i.e., you can currently [only register plugins for "lms" and "cms"](https://github.com/openedx/edx-django-utils/blob/master/edx_django_utils/plugins/docs/how_tos/how_to_create_a_plugin_app.rst#manual-setup) -- so we'd need to enhance it to also register plugins for learning core.

So I'd rather not convert this into a plugin right now, but plan for it by respecting the encapsulation @ormsbee laid out above:

> move this to a new `openedx_tagging` package in this repo, sibling of `openedx_learning`
> `openedx_tagging` has no foreign keys to anything in the `openedx_learning` package
> `openedx_learning.contrib.tagging` contains any models that need to link the two packages.

@ChrisChV I'm happy to do this move as part of the data architecture ADR which is coming this week.

ChrisChV: I like the approach :+1: 

> @ChrisChV I'm happy to do this move as part of the data architecture ADR which is coming this week, ref https://github.com/openedx/modular-learning/issues/60

Sure, thanks @pomegranited :+1:

ChrisChV:
Hi @ormsbee I think this is ready for merge. Work here has been continued in https://github.com/openedx/openedx-learning/pull/55

pomegranited:
Ah.. this PR isn't needed anymore since we merged https://github.com/openedx/openedx-learning/pull/55, so closing it. FYI @ChrisChV @ormsbee

openedx-webhooks:
@ChrisChV Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/54&entry.1484655318=https://github.com/ChrisChV&entry.752974735=2023-05-28+22:15&entry.1917517419=2023-06-14+00:15&entry.2133058324=No) so we can improve your experience in the future.

PR #57, opened 2023-06-13
Adds Taxonomy, Tag, ObjectTag models and APIs

pomegranited:
## Description

Implements the Taxonomy, Tag, and ObjectTag models, APIs, and permissions described in the [discovery document](https://docs.google.com/document/d/13zfsGDfomSTCp_G-_ncevQHAb4Y4UW0d_6N8R2PdHlA/edit?usp=sharing) for the tagging app.

## Supporting information

Relates to https://github.com/openedx/modular-learning/issues/63

## Testing instructions

* Ensure that all models and methods from the discovery doc are implemented.
* Ensure that the tests cover the expected behavior of the models, APIs and permissions as described in the discovery doc.

## Other information

The `import_tags`, `export_tags` and `autocomplete_tags` will be added by a subsequents PRs

SystemTaxonomy will be added by a subsequent PR as a subclass of Taxonomy.

ContentTaxonomy and ContentTag will subclass these models, and will be implemented in edx-platform `openedx.features.content_tagging`.

## To do before merge

- [x] use multi-collation field types added by https://github.com/openedx/openedx-learning/pull/51
- [x] squash migrations into a single `0001_initial.py`
- [ ] squash and merge this change (don't just merge)


openedx-webhooks:
Thanks for the pull request, @pomegranited!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,440 @@
...
+        max_length=255,
+        null=True,
+        blank=True,
+        help_text=_(
+            "Used to link an Open edX Tag with a tag in an externally-defined taxonomy."
+        ),
+    )
+
+    class Meta:
+        unique_together = ["taxonomy", "parent_id", "value"]
```
bradenmacdonald, line 57: I don't think this unique constraint will work as expected when `parent_id` is NULL, because null values are always different from each other. So you can have multiple entries with the same value and a NULL parent in the same taxonomy.

I think you can achieve what you want using [`UniqueConstraint`](https://docs.djangoproject.com/en/4.2/ref/models/constraints/#uniqueconstraint) which is more flexible.

bradenmacdonald (commented):
Very cool :) I am liking this approach.

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,440 @@
...
+        help_text=_(
+            "Indicates that tags in this taxonomy need not be predefined; authors may enter their own tag values."
+        ),
+    )
+    _object_tag_class = models.CharField(
+        max_length=255,
+        null=True,
+        blank=True,
+        help_text=_(
+            "Python path.module.class of the object tag class associated with this taxonomy.",
```
bradenmacdonald, line 138: Can you remind me, what is the use case for being able to override `ObjectTag` per taxonomy?

ChrisChV:
>Other information
>
>SystemTaxonomy will be added by a subsquent PR as a subclass of Taxonomy.
>
>ContentTaxonomy and ContentTag will subclass these models, and will be implemented in edx-platform openedx.features.tagging.

Also the `import_tags`, `export_tags` and `autocomplete_tags` will be added by a subsequents PRs

ChrisChV:
@pomegranited Great work! Looks good :+1:

openedx_tagging/core/tagging/models.py
bradenmacdonald: I generally like the YAGNI philosophy so I'd say add it when needed :)

pomegranited: Cool.. I'll put this code into the discovery in case it's needed later, but will remove it from here.

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,440 @@
...
+        """
+        Returns the Tag associated with the given lineage that is associated with the current taxonomy.
+
+        Raises Tag.DoesNotExist if no corresponding tag is found before reaching TAXONOMY_MAX_DEPTH.
+
+        Performance note: may perform as many as TAXONOMY_MAX_DEPTH select queries.
+        """
+        if self.allow_free_text:
+            raise Tag.DoesNotExist(
+                _(f"Free text taxonomy {self.id} has no predefined tags")
```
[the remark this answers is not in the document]
pomegranited: Have removed this method.

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,440 @@
...
+    """
+    Returns the given fully qualified python class.
+
+    Raises ImportError if the class cannot be loaded.
+
+    ref https://stackoverflow.com/a/547867
+    """
+    module_name, class_name = module_class.rsplit(".", 1)
+    module = __import__(module_name, fromlist=[class_name])
+    return getattr(module, class_name)
```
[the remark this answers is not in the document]
pomegranited: Removed the object_tag_class property, but noted this ^ in [the discovery](https://docs.google.com/document/d/13zfsGDfomSTCp_G-_ncevQHAb4Y4UW0d_6N8R2PdHlA/edit#heading=h.brvhx3mc8b5) if we end up needing it later.

docs/decisions/0009-tagging-administrators.rst
```diff
@@ -26,6 +26,8 @@ But because permissions #2 + #3 require access to the edx-platform CMS model `Co
 
 Per `OEP-9`_, ``openedx_tagging`` will allow applications to use the standard Django API to query permissions, for example: ``user.has_perm('openedx_tagging.edit_taxonomy', taxonomy)``, and the appropriate permissions will be applied in that application's context.
 
+These rules will be enforced in the tagging `views`_, not the API or models, so that external code using this library need not have a logged-in user in order to call the API. So please use with care.
```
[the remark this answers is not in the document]
ChrisChV: @pomegranited  It's reasonable to me. Maybe it is good to add this also as a comment in the API so that it is taken into account when using it

openedx_tagging/core/tagging/api.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    return taxonomy.get_tags()
+
+
+def resync_tags() -> int:
+    """
+    Reconciles all stored ObjectTag entries with any changes made to their associated taxonomies and tags.
+    """
+    num_changed = 0
+    for object_tag in ObjectTag.objects.all():
```
bradenmacdonald, line 66: If there are a _lot_ of `ObjecTag` rows, this has the downside of loading them all from the database and then converting them all into python objects before iterating over them. It seems like using `for object_tag in ObjectTag.objects.all().iterator()` [is better](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#iterator):

> For a `QuerySet` which returns a large number of objects that you only need to access once, this can result in better performance and a significant reduction in memory.

However, it's still not as good as I hoped:

> MySQL doesn’t support streaming results, hence the Python database driver loads the entire result set into memory. 

What?! Yet another point for postgres...

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,426 @@
...
+    Open edX tags are "name:value" pairs which can be applied to objects like content libraries, units, or people.
+    Tag.taxonomy.name provides the "name" and the Tag.value provides the "value".
+    (And an ObjectTag links a Tag with an object.)
     """
-    TODO: This is a basic representation of TagContent, it may change
-    depending on the discovery in:
-    https://docs.google.com/document/d/13zfsGDfomSTCp_G-_ncevQHAb4Y4UW0d_6N8R2PdHlA/edit#heading=h.o6fm1hktwp7b
 
-    This is the representation of a tag that is linked to a content
+    id = models.BigAutoField(primary_key=True)
```
pomegranited, line 24: @ormsbee I'm not clear on when it's better to use an `id` vs a `uuid`, and so I'm not sure which is best as the primary key for these models: Tag, Taxonomy, ObjectTag.

Since we're recommending multi-table inheritance to extend these models, they will share a database. However if we ever move this app to a separate service, I can see how a separate database would prefer to store UUIDs.

What do you think?

openedx_tagging/core/tagging/api.py
pomegranited: @bradenmacdonald Cool, I didn't know about that! Will fix this.

>> MySQL doesn’t support streaming results, hence the Python database driver loads the entire result set into memory.
> 
> What?! Yet another point for postgres...

My favourite related comment so far came from this issue: https://code.djangoproject.com/ticket/2495#comment:14

>  After eight years, and considering that the fix is far from easy, could we file this under "use a better database" and close the ticket?

pomegranited:
@ChrisChV My apologies for my [rebase on `main`](https://github.com/openedx/openedx-learning/compare/71a84ce64db0f126c2d763cb43a11a243857041d..c37f9f4ffdf74e9967d37f5057f307fd28595278).. I've rolled it back and did a [merge with `main`](https://github.com/openedx/openedx-learning/pull/57/commits/fbbd6f60744f4e62c9be61dddf9b16f87e1b41bb) instead, so you should still be able to change the base branch of your open PRs to this one to ease reviews there.

But I'll need to do a rebase before I merge here so I can squash those migrations. Awaiting word on [whether we need a uuid field](https://github.com/openedx/openedx-learning/pull/57#discussion_r1236025461) before I do that.

openedx_tagging/core/tagging/models.py
ormsbee: I think it's okay to have just an `id` here, since the combination of Taxonomy and `external_id` should be enough to uniquely identify a tag. If we find a need for UUIDs, we can add them later.

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,428 @@
...
+        max_length=255,
+        null=True,
+        blank=True,
+        help_text=_(
+            "Used to link an Open edX Tag with a tag in an externally-defined taxonomy."
+        ),
+    )
+
+    class Meta:
+        index_together = ["taxonomy", "value"]
```
ormsbee, line 60: This style is deprecated in 4.2. Please use [indexes](https://docs.djangoproject.com/en/4.2/ref/models/options/#django.db.models.Options.indexes) instead.

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,428 @@
...
+            "Tag that lives one level up from the current tag, forming a hierarchy."
+        ),
+    )
+    value = case_insensitive_char_field(
+        max_length=500,
+        help_text=_(
+            "Content of a given tag, occupying the 'value' part of the key:value pair."
+        ),
+    )
+    external_id = case_insensitive_char_field(
```
ormsbee, line 50: Should there be a unique constraint on `(taxonomy, external_id)`?

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,428 @@
 """ Tagging app data models """
+from typing import List, Type
+
 from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from openedx_learning.lib.fields import MultiCollationTextField, case_insensitive_char_field
```
ormsbee, line 7: Should we move the fields stuff out of `openedx_learning` and into something else? I'm just worried about circular dependencies here, since my understanding is that eventually there will something in `openedx_learning.contrib` that imports from `openedx_tagging` to link tags to the content data models?

openedx_tagging/core/tagging/api.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    return taxonomy.get_tags()
+
+
+def resync_tags() -> int:
+    """
+    Reconciles all stored ObjectTag entries with any changes made to their associated taxonomies and tags.
+    """
+    num_changed = 0
+    for object_tag in ObjectTag.objects.all().iterator():
```
ormsbee, line 68: I realize I've been a bit in-and-out of the taxonomy/tagging model discussions, so please pardon me if you folks have already gone over this:

I wonder if this could be less expensive if we make one more layer of model indirection between the Object and the tag. Right now, it seems like we have the global representation of a Tag, and a per-tagged-thing copy in case the global thing goes away. We need that because the Taxonomies+Tags ownership and lifecycle is different from ObjectTags. The process of reconciliation is between per-object copies and the taxonomy's canonical Tag.

But do we really need that many copies? Can't we keep a more normalized representation as long as we create a copy that follows the same ownership as the ObjectTags in question?

What if we kept a copy of the Taxonomy at the layer of the importable/exportable thing (e.g. library, course, learning package)? Take a course for example. Let's say:

* ObjectTags don't have local copies of the name.
* All the ObjectTags for course content is against a course-level Taxonomy.
* Some of those are CourseTaxonomies that are entirely owned and internal to the course.
* Some of those are CourseTaxonomies that serve as links-with-copies to other Taxonomies (CourseTaxonomies controlled by other courses, OrgTaxonomies, SystemTaxonomies).
* Inconsistencies within the Course on import are errors (they can be ignored and the values dropped, or they can stop the process, but there's no reconciliation).
* Reconciliation for updated values happens _between Taxonomies_.

That would significantly cut down on the number of rows that would have to be iterated over during any sort of reconciliation process.

openedx_tagging/core/tagging/models.py
pomegranited: Hmm.. you're right.. maybe they'd be better in https://github.com/openedx/edx-django-utils/ ?

openedx_tagging/core/tagging/api.py
pomegranited: Hi @ormsbee , we haven't gone over an option like this, so thank you for raising it! And you're right, it is storage-heavy to store copies of the `name:value` on each ObjectTag, which will store the most rows of any model in this package.

But, this does seem more complicate, so do we care that much about storage space?

> Right now, it seems like we have the global representation of a Tag, and a per-tagged-thing copy in case the global thing goes away.

We have three uses for the `ObjectTag._value` field:

1. What you said ^
2. We can import tagged content without having to have a Taxonomy with these Tags in place before importing.
3. Free-text tags can be anything the content author wants, and are not hierarchical.
   So we don't store them as `Tag`s, we just create an `ObjectTag` with `_value="free text value"` and `tag=None`. This also avoids race conditions; we don't have to worry about creating both a `Tag` and an `ObjectTag` when free-text tagging an object.

(`ObjectTag._name` is also in the model for reasons 1 & 2.)

But there's a lot of dancing going on around this `_value` field, which likely means it's trying to do too much.

If I understand you correctly, you're suggesting something like:

* Taxonomy: canonical source of a taxonomy's tags and metadata. Taxonomy Admins create these, import/export tags.
* ContentTaxonomy: references a `source` Taxonomy`, and keeps a copy of its Tag tree, e.g. as ContentTaxonomyTag
* ContentTag: references a ContentTaxonomy and a ContentTaxonomyTag.
* ContentTaxonomy objects know how to reconcile changes made to its `source` Taxonomy or the source Taxonomy's Tag without losing `ContentTag` data.

So this deals with point #1 above, but how should we handle points #2 and #3?

openedx_tagging/core/tagging/models.py
bradenmacdonald: Other parts of the code that I saw recently definitely made that assumption.

openedx_tagging/core/tagging/models.py
```diff
@@ -1,39 +1,428 @@
...
+        It's also useful for a set of ObjectTags are imported from an external source prior to when a Taxonomy exists to
+        validate or store its available Tags.
+
+        Returns True if anything was changed, False otherwise.
+        """
+        changed = False
+
+        # Locate a taxonomy matching _name
+        if not self.taxonomy_id:
+            taxonomy = Taxonomy.objects.filter(name=self.name, enabled=True).first()
```
bradenmacdonald, line 406: Should we have some kind of check here if this object is even allowed to be linked to this taxonomy? What if the taxonomy is for a completely different org than the course content that's tagged?

openedx_tagging/core/tagging/api.py
bradenmacdonald: My assumption was that reconciliation would be quite a rare process. If that's the case, I would strongly lean toward the simpler version, even if it uses a bit more space for storage. (Tags are usually quite small so I don't think the space is an issue.)

Even if it's not that rare, I think there's a simpler approach: If it's `ObjectTag.objects.all()` that's the problem, I suggest changing the reconciliation API to work on a single taxonomy or a single course at a time, rather than all tags of all objects. So for example it could take a course ID as its argument, and then do the reconciliation based on `ObjectTag.objects.filter(object_type=..., object_id__startswith="block-v1:the+course+ID@")`. Then you get much more efficient (and targeted) reconciliation at appropriate times (e.g. after a course import), and the data model is simple and easy to reason about.

What about taxonomy import? It seems like you might need to iterate over all the ObjectTags in that case, to see if any of them should be "upgraded" to new values in the taxonomy based on their `name` and `value`. But I think you can actually do a much more targeted iteration by only iterating the `ObjectTag` that have either (taxonomy_id is None) or (taxonomy_id==imported_taxonomy_id and tag_id is None). Which is again much much faster than iterating all the ObjectTags, because I expect that the vast majority of ObjectTags will be filtered out by that criteria (because they're linked to a taxonomy and/or a tag).

pomegranited: @bradenmacdonald 

> If that's the case, I would strongly lean toward the simpler version, even if it uses a bit more space for storage. 

That's my thinking too, and hence this PR.. but I am open to other options @ormsbee .

> Even if it's not that rare, I think there's a simpler approach: If it's ObjectTag.objects.all() that's the problem, I suggest changing the reconciliation API to work on a single taxonomy or a single course at a time, rather than all tags of all objects. 

👍 to adding filter options to the [resync_tags](https://github.com/open-craft/openedx-learning/blob/jill/taxonomy-api/openedx_tagging/core/tagging/api.py#L63-L73) method.

> What about taxonomy import? 

Perfect use of this, yes. I'll see if it'll fit into https://github.com/open-craft/openedx-learning/pull/1 , or create a follow-up task.

Thank you!

openedx_tagging/core/tagging/models.py
pomegranited: Oo good point.. enabled by [821ce62](https://github.com/openedx/openedx-learning/pull/57/commits/821ce62f8e6f127d97f9148b8bb88e0f29247ca1).

openedx_tagging/core/tagging/models.py
pomegranited: Fixed with [a11d1c9](https://github.com/openedx/openedx-learning/pull/57/commits/a11d1c9f6718cb5866b50ac7c9941823e49a5ecb).

pomegranited:
Thank you for all your reviews and advice, @ChrisChV @ormsbee and @bradenmacdonald !

I think this is ready to merge now -- if you agree, could someone please squash+merge this? I haven't squashed and rebased the commits here to keep the conversations working, but I can do that if it's preferred.

openedx_tagging/core/tagging/api.py
ormsbee: My apologies: I thought I replied to this yesterday but I never submitted my comment.

@bradenmacdonald: Yes, my concern was much more about the number of records that need to be updated (and any potential locking), and not the space usage.

@pomegranited:

> 2. We can import tagged content without having to have a Taxonomy with these Tags in place before importing.

Yeah, what I was proposing assumes that we'd have local Taxonomies/Tags in place for import, because those taxonomies can be locally controlled at the scope the import is happening.

> 3. Free-text tags can be anything the content author wants, and are not hierarchical.
>     So we don't store them as `Tag`s, we just create an `ObjectTag` with `_value="free text value"` and `tag=None`. This also avoids race conditions; we don't have to worry about creating both a `Tag` and an `ObjectTag` when free-text tagging an object.

I'm fuzzy on this part. It doesn't have to block this PR, but don't we have to worry about race conditions where we're making the same free text value ObjectTag for the same object multiple times? It seems like it would benefit from being a separate model altogether, so we could put a uniqueness constraints around the freeform tag and the thing being tagged, independent of a taxonomy?

ormsbee: To be clear: I'm fine with the proposal to keep the general shape of the data structures now and do per-course filtering during reconciliation. We can always add in the local-taxonomy-copy thing later if it turns out we want it. The questions I'm asking now are just to improve my understanding.

ormsbee:
@pomegranited: Could you please resolve merge conflicts? (I think it's fine for you to squash.)

openedx_tagging/core/tagging/models.py
pomegranited: Great plan, thank you @ormsbee . Should I codify this in an ADR in my next PR?

pomegranited: @ormsbee  These fields are working great on MySQL, but I can't use them in sqlite in the LMS tests :(

Running the tests with the branch from https://github.com/openedx/edx-platform/pull/32518 is throwing these errors when I run them on my [devstack](https://github.com/openedx/devstack)'s `lms-shell`:

```
root@lms:/edx/app/edxapp/edx-platform# pytest openedx/features/content_tagging/
<snip>
../venvs/edxapp/lib/python3.8/site-packages/pytest_django/plugin.py:490:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../venvs/edxapp/lib/python3.8/site-packages/pytest_django/fixtures.py:122: in django_db_setup
    db_cfg = setup_databases(
../venvs/edxapp/lib/python3.8/site-packages/django/test/utils.py:179: in setup_databases
    connection.creation.create_test_db(
../venvs/edxapp/lib/python3.8/site-packages/django/db/backends/base/creation.py:74: in create_test_db
    call_command(
../venvs/edxapp/lib/python3.8/site-packages/django/core/management/__init__.py:181: in call_command
    return command.execute(*args, **defaults)
../venvs/edxapp/lib/python3.8/site-packages/django/core/management/base.py:398: in execute
    output = self.handle(*args, **options)
../venvs/edxapp/lib/python3.8/site-packages/pytest_django/fixtures.py:276: in handle
    return super().handle(*args, **kwargs)
../venvs/edxapp/lib/python3.8/site-packages/django/core/management/base.py:89: in wrapped
    res = handle_func(*args, **kwargs)
../venvs/edxapp/lib/python3.8/site-packages/django/core/management/commands/migrate.py:262: in handle
    post_migrate_apps.render_multiple([
../venvs/edxapp/lib/python3.8/site-packages/django/core/management/commands/migrate.py:263: in <listcomp>
    ModelState.from_model(apps.get_model(*model)) for model in model_keys
../venvs/edxapp/lib/python3.8/site-packages/django/db/migrations/state.py:407: in from_model
    fields.append((name, field.clone()))
../venvs/edxapp/lib/python3.8/site-packages/django/db/models/fields/__init__.py:513: in clone
    return self.__class__(*args, **kwargs)
<snip>
ERROR openedx/features/content_tagging/tests/test_api.py::TestAPIContentTaxonomy::test_content_taxonomy_enabled_for_org1 - ValueError: Cannot use db_collation with MultiCollationCharField. Please use a db_collations dict instead.
```

And I found your post related to this on https://forum.djangoproject.com/t/supporting-collation-settings-on-multiple-database-backends/12561/, but I'm unclear how (or if) I can work around this.

Should I just roll these back to being normal, default-collation CharField/TextField until we work out a solution for both sqlite and MySQL?

openedx_tagging/core/tagging/api.py
pomegranited: > It seems like it would benefit from being a separate model altogether, so we could put a uniqueness constraints around the freeform tag and the thing being tagged, independent of a taxonomy?

I'll give it some thought.. originally when I was doing the discovery, I thought we'd end up having to subclass all of these base classes, which would lead to lots and lots of classes, e.g. SystemTaxonomy, SystemTaxonomyTag, SystemTaxonomyObjectTag, ...  And so I was trying to reduce the number of classes used at the base here. But now that I've gotten into the ContentTaxonomy implementation, I didn't even need to create a ContentTag class, because all of the ObjectTag validation is handled by the taxonomy. We'll still need a FreeFormTaxonomy of some sort, even if it has no tags, because that model is where the namespace for the object tags are stored.

Unfortunately we're running short on time for this phase 1, and so I'd rather not tweak it too much more until we actually have a frontend that's using it. But if I find that the frontend is having to do a lot of `if freeform then do this else do something else" before calling the backend, then separating them out will make much more sense.

openedx_tagging/core/tagging/models.py
pomegranited: If I remove the check ([a1a90d2](https://github.com/openedx/openedx-learning/pull/57/commits/a1a90d2689dd1537e2cc87544945d3abe0362e17)) then my tests on https://github.com/openedx/edx-platform/pull/32518 pass, cf https://github.com/openedx/edx-platform/actions/runs/5383371513/jobs/9770044365

Are you ok with merging this without that check?

ormsbee: Yes. I'm going to look through it later to try to figure out why it did this, but I'm fine with removing it for now. Thank you.

ormsbee (approved):

ormsbee:
@pomegranited: is this okay for me to rebase+merge now?

pomegranited:
@ormsbee Yes please! Sorry for the delay, I didn't think you were even up :)

openedx-webhooks:
@pomegranited 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/57&entry.1484655318=https://github.com/pomegranited&entry.752974735=2023-06-13+08:10&entry.1917517419=2023-06-27+04:10&entry.2133058324=Yes) so we can improve your experience in the future.

ormsbee:
I'm a parent. I keep weird hours sometimes. 😉

PR #58, opened 2023-06-19
import/export Taxonomy API functions

ChrisChV:
## Description
Implements [import_tags](https://docs.google.com/document/d/13zfsGDfomSTCp_G-_ncevQHAb4Y4UW0d_6N8R2PdHlA/edit#heading=h.bu5slo87hawp) and [export_tags](https://docs.google.com/document/d/13zfsGDfomSTCp_G-_ncevQHAb4Y4UW0d_6N8R2PdHlA/edit#heading=h.u3ydxm3x6l0n) functions

## Supporting information
Merge after https://github.com/openedx/openedx-learning/pull/57. 

Closes https://github.com/openedx/modular-learning/issues/64

## Testing instructions

- Ensure that the tests cover the expected behaviour

### DIFF
--- a/openedx_tagging/core/tagging/api.py
+++ b/openedx_tagging/core/tagging/api.py
@@ -10,12 +10,27 @@ No permissions/rules are enforced by these methods -- these must be enforced in 
 Please look at the models.py file for more information about the kinds of data
 are stored in this app.
 """
+import csv
+import json
+from enum import Enum
+from io import StringIO, BytesIO, TextIOWrapper
 from typing import List, Type
 
+from django.db import transaction
 from django.db.models import QuerySet
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils.translation import gettext_lazy as _
 
 from .models import ObjectTag, Tag, Taxonomy
+
+csv_fields = ['id', 'name', 'parent_id', 'parent_name']
+
+class TaxonomyDataFormat(Enum):
+    """
+    Formats used to export and import Taxonomies
+    """
+    CSV = 'CSV'
+    JSON = 'JSON'
 
 
 def create_taxonomy(
@@ -29,6 +44,7 @@ def create_taxonomy(
     """
     Creates, saves, and returns a new Taxonomy with the given attributes.
     """
+
     return Taxonomy.objects.create(
         name=name,
         description=description,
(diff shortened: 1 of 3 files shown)


openedx-webhooks:

- supporting documentation

tests/openedx_tagging/core/tagging/test_api.py
```diff
@@ -205,3 +211,124 @@ def test_tag_object_invalid_tag(self):
                 "course",
             )
         assert "Invalid object tag for taxonomy" in str(exc.exception)
+
+    def test_import_tags_csv(self):
```
pomegranited, line 215: There's some lines missing from the test coverage for import and export -- could you add tests for them too?

```
openedx_tagging/core/tagging/api.py                                                 145     11     64      9    90%   185, 191, 206, 212->224, 215, 221-222, 224->exit, 287-289, 304, 310, 340->exit
```

ChrisChV: Sure, thanks!

bradenmacdonald:
@ChrisChV Until @pomegranited 's PR is merged, can you please change the "merge target" of this PR from `main` to her branch? Then it will only show the changes from your new commit. After her PR merges you can change it back to `main`.

openedx_tagging/core/tagging/api.py
```diff
@@ -98,5 +114,180 @@ def tag_object(
...
+        tag_id = tag['id']
+        tag_name = tag['name']
+        tag_parent_id = tag.get('parent_id')
+        tag_parent_name = tag.get('parent_name')
+
+        # Check if the tag has not already been created or updated
+        if tag_id not in new_tags and tag_id not in updated_tags:
+            try:
+                # Update tag
+                tag_instance = Tag.objects.get(external_id=tag_id)
```
bradenmacdonald, line 197: You also need to use `taxonomy=taxonomy` here, or you can have cross-taxonomy bugs.

openedx_tagging/core/tagging/api.py
```diff
@@ -98,5 +114,180 @@ def tag_object(
...
+                tag_instance.value = tag_name
+
+                if tag_instance.parent and (not tag_parent_id or not tag_parent_name):
+                    # if there is no parent in the data import
+                    tag_instance.parent = None
+                updated_tags.append(tag_id)
+            except ObjectDoesNotExist:
```
bradenmacdonald, line 204: Better to use `Tag.DoesNotExist` instead of the generic base class.

openedx_tagging/core/tagging/api.py
```diff
@@ -98,5 +114,180 @@ def tag_object(
...
+            if tag_parent_id and tag_parent_name:
+                # Parent creation/update
+                parent = create_update_tag({'id': tag_parent_id, 'name': tag_parent_name})
+                tag_instance.parent = parent
+
+            tag_instance.save()
+            return tag_instance
+        else:
+            # Returns the created/updated tag from history
+            return Tag.objects.get(external_id=tag_id)
```
bradenmacdonald, line 222: Same thing here, needs `taxonomy=taxonomy`. The same `external_id` can appear in multiple taxonomies.

BTW if you add something like this to the `Taxonomy` class:

```python
@property 
def tags(self):
    return Tag.objects.filter(taxonomy=self)
```

then this code can become `return taxonomy.tags.get(external_id=tag_id)`

openedx_tagging/core/tagging/api.py
pomegranited: Oh wow.. I should have seen that!

openedx_tagging/core/tagging/api.py
pomegranited: There's already `taxonomy.tag_set`, so I don't think we need another property?

bradenmacdonald: Oh right, duh... not sure why I forgot that. But yes, use that :)

openedx_tagging/core/tagging/api.py
```diff
@@ -98,5 +114,180 @@ def tag_object(
...
+            csv_reader = csv.DictReader(text_tags)
+            header_fields = csv_reader.fieldnames
+            if csv_fields != header_fields:
+                raise ValueError(
+                    _(
+                        f"Invalid CSV header: {header_fields}. Must be: {csv_fields}."
+                    )
+                )
+            tags_data = list(csv_reader)
+        else:
```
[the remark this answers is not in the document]
ChrisChV: @bradenmacdonald I think this is valid on the import. In the export I have left the verification at the beginning before loading the tags

ChrisChV:
> @ChrisChV Until @pomegranited 's PR is merged, can you please change the "merge target" of this PR from main to her branch? Then it will only show the changes from your new commit. After her PR merges you can change it back to main.

@bradenmacdonald I have not found the way to change `openedx/openedx-learning` to `open-craft/openedx-learning`. I created this https://github.com/open-craft/openedx-learning/pull/1 on `open-craft/openedx-learning`. Do yo know a way to change the "merge target" between repos? 
Personally, in these cases I do not like to have two PRs since you can lose the thread of the reviews

bradenmacdonald:
@ChrisChV Oh, I see. In that case, what you'd have to do is push Jill's branch to the `openedx` fork if you can; if not, just leave it how you have it and link to the commit. That's fine.

ChrisChV:
@ormsbee This is ready for your review

ChrisChV:
@ormsbee I merged the changes of https://github.com/openedx/openedx-learning/pull/57. Now the review is easier and it's ready

openedx_tagging/core/tagging/api.py
```diff
@@ -10,13 +10,28 @@
...
 from typing import List, Type
 
+from django.db import transaction
 from django.db.models import QuerySet
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils.translation import gettext_lazy as _
 
 from .models import ObjectTag, Tag, Taxonomy
 
+csv_fields = ['id', 'name', 'parent_id', 'parent_name']
```
pomegranited, line 26: nit: would like to make it clear that this is a constant, and it's not part of the externally-exportable python api:

```suggestion
_CSV_FIELDS = ['id', 'name', 'parent_id', 'parent_name']
```

tests/openedx_tagging/core/tagging/test_api.py
```diff
@@ -188,3 +194,205 @@ def test_tag_object(self):
...
+        object_tag = ObjectTag(
+            object_id=object_id,
+            object_type='course',
+            taxonomy=self.taxonomy,
+            tag=self.bacteria,
+        )
+        tagging_api.resync_object_tags([object_tag])
+        object_tag = ObjectTag.objects.get(object_id=object_id)
+        self.assertEqual(object_tag.tag.value, 'Bacteria')
+        self.assertEqual(object_tag._value, 'Bacteria')
```
pomegranited, line 340: nit: pylint doesn't like this, but I agree it's necessary for the test. So could you please add this here and in the two other mentions below?

```suggestion
        self.assertEqual(object_tag._value, 'Bacteria')  # pylint: disable=protected-access
```

pomegranited (approved):
@ChrisChV I found a couple of nits, but otherwise good 👍 

- [x] I tested this by running the tests locally and ensuring 100% coverage.
- [x] I read through the code and tests, and ensured that the tests cover our use cases.
- [x] Includes documentation

openedx_tagging/core/tagging/api.py
```diff
@@ -105,5 +121,180 @@ def tag_object(
...
+            )
+    except ValueError as e:
+        raise e
+    finally:
+        tags.close()
+
+
+    updated_tags = []
+
+    def create_update_tag(tag):
```
ormsbee, line 181: Please try to avoid inner functions this large and complex. The `create_update_tag` function has interesting logic but it's harder to test separately because it's nested in here.

ormsbee (changes requested):
I started writing per-line reviews, but then I thought it best to step back a bit.

Import is going to be one of those places where the complexity is going to grow a lot as time goes on. We already know that we're going to have to support a variety of modifications in the future–consolidating tags, renaming tags, etc.

We're clearly not going to implement that functionality yet. But the fact that this is going to grow a lot means that we should be really disciplined and explicit about the stages of data transformation that we have here. We also have to keep the code as painfully simple and easy to test as possible.

What I expect to see as a top level import function would be something fairly short that that calls out to other classes/functions to implement something like the following steps, where the output of one step is the input into the next:

1. Parse the raw input into a DSL of actions (e.g. CreateOrUpdate, Delete, Rename, etc.).
2. Validate that the parsed set of actions is internally consistent (e.g. we're not adding a Tag as a parent of itself). 
3. Validate that the actions make sense against the Taxonomy data being operated on (requires reading data form our Taxonomy).
4. Create a plan for the operations we'll need to do (because preview functionality is really nice to have when there's no revert button).
5. Execute that plan.

Each of those steps should be separately testable. We should be able to have a CSVParser and a JSONParser and have the inputs and outputs of their tests only be the raw data and the actions they generate.

We should be able to have a set of actions, and test the function that validates them.

When we chain things together, we should be able to set up a database state and a list of DSL statements and test the plan that we generate.

Right now, all of these concerns are mixed together in the `import_tags`. This will make it harder to debug as time goes on, and encourage the use of side-effects to effectively "message" later parts of the pipeline with dicts built earlier on.

These may not seem like immediate problems, but if we make it so that the easiest way to add new features is to add a couple dozen lines to `import_tags` and make a new 50 line inner function there, things are going to get very confusing very quickly.

ChrisChV:
Closed in favor of https://github.com/openedx/openedx-learning/pull/64

openedx-webhooks:
@ChrisChV Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/58&entry.1484655318=https://github.com/ChrisChV&entry.752974735=2023-06-19+19:39&entry.1917517419=2023-07-14+19:26&entry.2133058324=No) so we can improve your experience in the future.

PR #59, opened 2023-06-20
Autocomplete tags API function

ChrisChV:
## Description
Implements [autocomplete_tags](https://docs.google.com/document/d/13zfsGDfomSTCp_G-_ncevQHAb4Y4UW0d_6N8R2PdHlA/edit#heading=h.bxm1g2zbzdpi) function

## Supporting information
Merge after https://github.com/openedx/openedx-learning/pull/57. 

Closes https://github.com/openedx/modular-learning/issues/65

## Testing instructions
- Ensure that the tests cover the expected behaviour


openedx-webhooks:

- supporting documentation

tests/openedx_tagging/core/tagging/test_api.py
```diff
@@ -0,0 +1,218 @@
...
+    def _validate_autocomplete_tags(self, taxonomy, prefix, object_id, free_text=False):
+        """
+        Validate autocomplete tags
+        """
+
+        # Normal search
+        result = tagging_api.autocomplete_tags(taxonomy, prefix)
+        self.assertEqual(len(result), 4)
+        for tag in result:
+            self.assertEqual(tag[0].lower(), prefix)
```
pomegranited, line 192: Can you check that the tags are returned in the expected order?

```suggestion
        self.assertEqual(result, ["Animalia", "Archaea", "Archaebacteria", "Arthropoda"])
```

tests/openedx_tagging/core/tagging/test_api.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # Search with count
+        result = tagging_api.autocomplete_tags(taxonomy, prefix, count=2)
+        self.assertEqual(len(result), 2)
```
pomegranited, line 196: ditto re order:

```suggestion
        self.assertEqual(result, ["Animalia", "Archaea"])
```

tests/openedx_tagging/core/tagging/test_api.py
```diff
@@ -0,0 +1,218 @@
...
+                object_id=object_id,
+                object_type='course',
+                taxonomy=taxonomy,
+                tag=tag_model,
+                _value=tag,
+            ).save()
+
+        # Search with object
+        result = tagging_api.autocomplete_tags(taxonomy, prefix, object_id)
+        self.assertEqual(len(result), 2)
```
pomegranited, line 214: ditto re order:

```suggestion
        self.assertEqual(result, ["Archaebacteria", "Arthropoda"])
```

tests/openedx_tagging/core/tagging/test_api.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # Search with object and count
+        result = tagging_api.autocomplete_tags(taxonomy, prefix, object_id, count=1)
+        self.assertEqual(len(result), 1)
```
pomegranited, line 218: ditto re order:
```suggestion
        self.assertEqual(result, ["Archaebacteria"])
```

openedx_tagging/core/tagging/api.py
```diff
@@ -0,0 +1,136 @@
...
+    if taxonomy.allow_free_text:
+        # Free-text taxonomy
+        excluded_tags = excluded_tags.values_list('_value', flat=True)
+        result = ObjectTag.objects.filter(taxonomy=taxonomy, _value__startswith=prefix) \
+                                  .exclude(object_id=object_id) \
+                                  .exclude(_value__in=excluded_tags) \
+                                  .order_by('_value') \
+                                  .values_list('_value', flat=True)
+        # Remove repeat tags
+        result = list(set(result))[:count]
```
pomegranited, line 127: This `set` operation is mucking with the sorting, so they're not being returned in alphabetical order:

```suggestion
                                  .values_list('_value', flat=True) \
                                  .distinct()[:count]
```

openedx_tagging/core/tagging/api.py
```diff
@@ -0,0 +1,136 @@
...
+    else:
+        # Closed taxonomy
+        excluded_tags = excluded_tags.filter(tag__isnull=False).values_list('tag__id', flat=True)
+        result = Tag.objects.filter(taxonomy=taxonomy, value__startswith=prefix) \
+                            .exclude(id__in=excluded_tags) \
+                            .order_by('value') \
+                            .values_list('value', flat=True) \
+                            [:count]
```
pomegranited, line 135: Should do the same here just in case there's repeated tag values in the taxonomy:

```suggestion
                            .values_list('value', flat=True) \
                            .distinct()[:count]
```

openedx_tagging/core/tagging/api.py
```diff
@@ -0,0 +1,136 @@
...
+    the tags used by `object_id`.
+    """
+    result = []
+    excluded_tags = ObjectTag.objects.filter(object_id=object_id, taxonomy=taxonomy)
+    if taxonomy.allow_free_text:
+        # Free-text taxonomy
+        excluded_tags = excluded_tags.values_list('_value', flat=True)
+        result = ObjectTag.objects.filter(taxonomy=taxonomy, _value__startswith=prefix) \
+                                  .exclude(object_id=object_id) \
+                                  .exclude(_value__in=excluded_tags) \
```
pomegranited, line 123: Could you add comments (here and below) explaining each of these filter/excludes? I found the two excludes confusing when I read them (though I get it now).

and nit: black formatting recommends you put all of this in `(  )` instead of using `\`:
```python
result = (
    # Fetch object tags from this taxonomy whose value starts with the given prefix
    ObjectTag.objects.filter(taxonomy=taxonomy, _value__startswith=prefix)
    # omit any tags applied to the given object
    .exclude(object_id=object_id)
    # omit any free-text tags whose values match the tags on the given object
    .exclude(_value__in=excluded_tags)
    ...
)
```

pomegranited (changes requested):

openedx_tagging/core/tagging/api.py
ChrisChV: Good catch, Thanks!

openedx_tagging/core/tagging/api.py
```diff
@@ -0,0 +1,136 @@
...
+    Otherwise, it should be a list of existing Tag IDs.
+
+    Raised ValueError if the proposed tags are invalid for this taxonomy.
+    Preserves existing (valid) tags, adds new (valid) tags, and removes omitted (or invalid) tags.
+    """
+
+    return taxonomy.tag_object(tags, object_id, object_type)
+
+
+def autocomplete_tags(taxonomy: Taxonomy, prefix: str, object_id: str= None, count=10) -> List[str]:
```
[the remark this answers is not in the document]
ChrisChV: @pomegranited I think it would be better to use the first option. If we return objects that don't exist, at some point I feel like can be saved and generate errors

ChrisChV:
@pomegranited The CI tests are failing, but in my local works normally. Please let me know your results in your local

openedx_tagging/core/tagging/api.py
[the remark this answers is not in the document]
pomegranited: Sure @ChrisChV  -- you could define GenericTag with [namedtuple](https://www.geeksforgeeks.org/namedtuple-in-python/) to keep it simple.

openedx_tagging/core/tagging/models.py
```diff
@@ -326,6 +326,9 @@ class ObjectTag(models.Model):
         ),
     )
 
+    class Meta:
+        index_together = ["taxonomy", "_value"]
+
```
pomegranited, line 333: @ChrisChV  This change has been included in https://github.com/openedx/openedx-learning/pull/57/files#diff-204875f41064da3abfd4648611ad41c6f21abe09277263410e3daafbd343f2a3R349-R351

so can be removed from here.

pomegranited:
> @pomegranited The CI tests are failing, but in my local works normally. Please let me know your results in your local

@ChrisChV Hmm.. `make test` passes for me locally too, but the errors look like sorting issues.. So maybe it's something to do with CI running the tests on mysql? Maybe we need to explicitly sort the results of `get_object_tags` to keep them consistent?

openedx_tagging/core/tagging/models.py
ChrisChV: Sure, thanks :+1:

ChrisChV:
> @ChrisChV Hmm.. make test passes for me locally, and [I had the same problem on mine](https://github.com/openedx/openedx-learning/actions/runs/5342906059/jobs/9685291697?pr=57), so I think it's a sorting issue, something to do with CI running the tests on mysql?
>
> If you merge my latest changes, it should fix it.

@pomegranited Thanks. MySQL uses `__istartswith` to make case-insensitive searches.

ChrisChV:
@ormsbee It's ready for your review

ChrisChV:
@ormsbee I merged the changes of https://github.com/openedx/openedx-learning/pull/57. Now the review is easier and it's ready

openedx_tagging/core/tagging/api.py
```diff
@@ -11,12 +11,12 @@
 are stored in this app.
 """
 from typing import List, Type
+from collections import OrderedDict
```
pomegranited, line 14: nit: this isn't used?
```suggestion
```

openedx_tagging/core/tagging/models.py
```diff
@@ -456,3 +456,12 @@ def resync(self) -> bool:
             changed = True
 
         return changed
+
```
pomegranited, line 459: nit: needs one more blank line:
```suggestion
    
    
```

openedx_tagging/core/tagging/api.py
```diff
@@ -107,3 +107,73 @@ def tag_object(
...
+        )
+
+        # We need to delete the repeats here because the current database backend 
+        # don't support `distinct()` on fields
+        unique_results = {}
+        for item in result:
+            value = item['value']
+            if value not in unique_results:
+                unique_results[value] = item
+        result = list(unique_results.values())
```
pomegranited, line 175: Hmm.. I think we should either:

* enforce `unique_together = ("taxonomy", "value")` on the Tag model, OR
* return all the matching Tags, even if there are duplicates.

Because either we won't have multiple tags in a taxonomy with the same name, or we'll confuse the user by only returning the first one available when they're searching for a matching tag.

Sounds like a product question, though I'm hard-pressed to come up with a use case for duplicate tags at the leaf level anyway..

pomegranited (changes requested):
Couple of minor nits, but I've also got a question about duplicate Tags in a single taxonomy?

openedx_tagging/core/tagging/api.py
```diff
@@ -107,3 +107,73 @@ def tag_object(
...
+    Closed taxonomies return tag values that exist in the taxonomy. Also excludes all Tags that the
+    `object_id` already has. 
+
+    Free-text taxonomies return only tag values that are currently in use on (another) object. Also excludes
+    the tags used by `object_id`.
+    """
+    result = []
+
+    # Fetch tags that the object already has to exclude them from the result
+    excluded_tags = ObjectTag.objects.filter(object_id=object_id, taxonomy=taxonomy)
```
ormsbee, line 126: Please add the appropriate index for this query.

openedx_tagging/core/tagging/api.py
```diff
@@ -107,3 +107,73 @@ def tag_object(
...
+            ObjectTag.objects.filter(taxonomy=taxonomy, _value__istartswith=prefix)
+            # omit any tags applied to the given object
+            .exclude(object_id=object_id)
+            # omit any free-text tags from other objects whose values match the tags on the given object
+            .exclude(_value__in=excluded_tags)
+            # alphabetical ordering
+            .order_by('_value')
+            # obtain the values of the tags
+            .values_list('_value', flat=True)
+            # remove repeats
```
ormsbee, line 143: In what scenarios could we get repeats?

openedx_tagging/core/tagging/api.py
```diff
@@ -107,3 +107,73 @@ def tag_object(
...
+            .distinct()
+            # get only first `count` values
+            [:count]
+        )
+        # Build tag results
+        result = [TagResult(id=None, value=value) for value in result]
```
ormsbee, line 149: Given that this is likely to be used by a REST API at some point, I think it's fine to have a function that returns a QuerySet, even if this function just calls out to it and does this serialization into `TagResult`.

openedx_tagging/core/tagging/api.py
```diff
@@ -107,3 +107,73 @@ def tag_object(
...
+
+        # Obtain the value of the excluded tags
+        excluded_tags = excluded_tags.values_list('_value', flat=True)
+        result = (
+            # Fetch object tags from this taxonomy whose value starts with the given prefix
+            ObjectTag.objects.filter(taxonomy=taxonomy, _value__istartswith=prefix)
+            # omit any tags applied to the given object
+            .exclude(object_id=object_id)
+            # omit any free-text tags from other objects whose values match the tags on the given object
+            .exclude(_value__in=excluded_tags)
```
ormsbee, line 138: If we're omitting everything with the `object_id` here, why do we need to check `excluded_tags` as well (which should just be a subset)?

ormsbee (dismissed):
Mostly questions and requests around the database queries being made. Thank you!

openedx_tagging/core/tagging/api.py
ChrisChV: > Sounds like a product question, though I'm hard-pressed to come up with a use case for duplicate tags at the leaf level anyway.

It's possible to have two tags with the same value but with different parent? This is the only use case I can think of.
If that can be done, then we can't use `unique_together = ("taxonomy", "value", "parent")`, bacuse doesn't work with foreing keys.
 
I think for now it's ok to return all matching Tags

openedx_tagging/core/tagging/api.py
ChrisChV: About the free-form taxonomies, we do the search in ObjectTag. There may be the scenario where different objects have the same tag

openedx_tagging/core/tagging/api.py
ChrisChV: It's the same logic of https://github.com/openedx/openedx-learning/pull/59#discussion_r1246982861. There may be tags of other objects that the `object_id` already has. But I have realized that `.exclude(object_id=object_id)` is no necessary, because `.exclude(_value__in=excluded_tags)` also excludes the tags of `object_id`

openedx_tagging/core/tagging/api.py
ChrisChV: Sure, [added](https://github.com/openedx/openedx-learning/pull/59/files#diff-204875f41064da3abfd4648611ad41c6f21abe09277263410e3daafbd343f2a3R352) :+1:

openedx_tagging/core/tagging/api.py
ChrisChV: You are right. I have added a [separate function](https://github.com/openedx/openedx-learning/pull/59/files#diff-951a5c4d32e1301ac1c30419ebce4ada8ae17db6c93f4c2a5a34caac9eb30e0aR164) for serialization, that can be used by other internal functions

openedx_tagging/core/tagging/api.py
pomegranited: @ChrisChV 

> It's possible to have two tags with the same value but with different parent?

Yep, that's possible from a data perspective, but if Product doesn't require it, then we don't have to make room for it in our implementation.

Can you raise the question with Product, and see if they can find out whether any of the taxonomies they want to use have any duplicate leafs?

ChrisChV: @pomegranited [Question in Slack](https://openedx.slack.com/archives/C03T0RZEASG/p1688402959991889). As @ormsbee commented, I added the constraint: adb3e5afdba90aec3b78a0ec609cd9fd29b093bb

mphilbrick211:
Hi @ChrisChV! Just following up on this :)

pomegranited:
Hi @mphilbrick211 :) This is blocked on https://github.com/openedx/openedx-learning/pull/62, but we're nearly there :)

mphilbrick211:
> Hi @mphilbrick211 :) This is blocked on #62, but we're nearly there :)

Thanks, @pomegranited!

openedx_tagging/core/tagging/api.py
```diff
@@ -107,3 +106,68 @@ def tag_object(
...
+def autocomplete_tags_result(taxonomy: Taxonomy, prefix: str, object_id: str= None, count=10) -> List[TagResult]:
+    """
+    Calls `autocomplete_tags` and serialize the results into `TagResult`
+    """
+    result = autocomplete_tags(taxonomy, prefix, object_id=object_id, count=count)
+    if taxonomy.allow_free_text:
+        result = [TagResult(id=None, value=value) for value in result]
+    else:
+        result = [TagResult(id=item['id'], value=item['value']) for item in result]
+    return result
```
pomegranited, line 173: @ChrisChV I think we can let the REST API serialize the results here, so this method doesn't need to be in the python API.

openedx_tagging/core/tagging/api.py
```diff
@@ -107,3 +106,68 @@ def tag_object(
     """
 
     return taxonomy.tag_object(tags, object_id, object_type)
+
+
+def autocomplete_tags(taxonomy: Taxonomy, prefix: str, object_id: str= None, count=10) -> QuerySet:
```
pomegranited, line 111: @ChrisChV  Can we move the bulk of this functionality into the Taxonomy model? That way, if system taxonomies (for example) need to customize anything, they can just overwrite the method.

pomegranited (changes requested):
Hi @ChrisChV  -- I've left some comments in light of the soon-to-be-approved https://github.com/openedx/openedx-learning/pull/62

openedx_tagging/core/tagging/models.py
```diff
@@ -349,6 +350,7 @@ class ObjectTag(models.Model):
     class Meta:
         indexes = [
             models.Index(fields=["taxonomy", "_value"]),
+            models.Index(fields=["taxonomy", "object_id"]),
```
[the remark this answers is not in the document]
ChrisChV: Added here https://github.com/openedx/openedx-learning/pull/59/commits/507848b6a723c84c84d8b920abeb5916e23fd9f4

openedx_tagging/core/tagging/api.py
ChrisChV: Updated here https://github.com/openedx/openedx-learning/pull/59/commits/223b3a4c314d84255a36e73cd80d9b108e577568

openedx_tagging/core/tagging/api.py
ChrisChV: Updated here https://github.com/openedx/openedx-learning/pull/59/commits/223b3a4c314d84255a36e73cd80d9b108e577568

openedx_tagging/core/tagging/models.py
```diff
@@ -409,6 +413,62 @@ def tag_object(
 
         return updated_tags
 
+    def autocomplete_tags(
+        self, prefix: str, object_id: str = None, count=10
+    ) -> models.QuerySet:
```
pomegranited, line 418: I think it's going to be terribly confusing that the QuerySet returned here is a ObjectTag queryset for free-text taxonomies, but a Tag queryset for closed taxonomies. So we can't simply use the API and treat them like the same thing.

Plus, for closed taxonomies, the frontend can handle auto-completion without consulting the backend (e.g using [autocompleter](https://www.npmjs.com/package/autocompleter)), because it's already fetched all of the Tags in a Taxonomy to display the tree of options: 

autocomplete

So I think we should:

* Only autocomplete from existing ObjectTags, ignoring the free-text vs closed taxonomy distinction.
* Match `text` anywhere in the value, not just `prefix` (as shown above), by using `.filter(_value__icontains=text)` instead
* Remove the `count` parameter -- the REST API that ends up calling this can handle pagination from the returned query set.

Does that sound reasonable to you?

pomegranited (changes requested):

openedx_tagging/core/tagging/models.py
ChrisChV: Updated here https://github.com/openedx/openedx-learning/pull/59/commits/a8523b5cc3f957344f8b707f110d41db0cfc36ce

openedx_tagging/core/tagging/models.py
```diff
@@ -409,6 +413,32 @@ def tag_object(
...
+        The result is returned in alphabetical order.
+
+        The output is a QuerySet of dictionaries in with `_value` and `tag`.
+
+        Subclasses can override this method to perform their own autocomplete process.
+        """
+        # Fetch tags that the object already has to exclude them from the result
+        excluded_tags = self.objecttag_set.filter(object_id=object_id).values_list(
+            "_value", flat=True
+        )
```
pomegranited, line 428: If no object_id is provided, we don't need to make this call..  so can this be:

```suggestion
        excluded_tags = []
        if object_id:
            excluded_tags = self.objecttag_set.filter(object_id=object_id).values_list(
                "_value", flat=True
            )
```

openedx_tagging/core/tagging/api.py
```diff
@@ -136,3 +136,19 @@ def tag_object(
...
+    search: str,
+    object_id: str = None,
+) -> QuerySet:
+    """
+    Returns tag values that contains the given search string.
+    The result is returned in alphabetical order.
+
+    The output is a QuerySet of dictionaries in with `_value` and `tag`.
+
+    Subclasses can override this method to perform their own autocomplete process.
```
pomegranited, line 152: I think we can say more here about the purpose of this api function.

And note: subclasses cannot override this method because it's an API, not a model method.

```suggestion
    Provides auto-complete suggestions by matching the `search` string against existing object tags linked to the given taxonomy. A case-insensitive search is used in order to return the highest number of relevant tags.
    
    If `object_id` is provided, then object tag values already linked to this object are omitted from the returned suggestions. (Object tag values must be unique for a given object + taxonomy, and so omitting these suggestions helps users avoid duplication errors.)
    
    Returns a QuerySet of dictionaries containing distinct `value` (string) and `tag` (numeric ID) values, sorted alphabetically by `value`. The `value` is what should be shown as a suggestion to users, and if it's a free-text taxonomy, `tag` will be `None`:  we include the `tag` ID in anticipation of the second use case listed below.
    
    Use cases:
    * This method is useful for reducing tag variation in free-text taxonomies by showing users tags that are similar to what they're typing. E.g., if the `search` string "dn" shows that other objects have been tagged with "DNA", "DNA electrophoresis", and "DNA fingerprinting", this encourages users to use those existing tags if relevant, instead of creating new ones that look similar (e.g. "dna finger-printing").
    * It could also be used to assist tagging for closed taxonomies with a list of possible tags which is too large to return all at once, e.g. a user model taxonomy that dynamically creates tags on request for any registered user in the database. (Note that this is not implemented yet, but may be as part of a future change.)
```

openedx_tagging/core/tagging/models.py
```diff
@@ -409,6 +413,32 @@ def tag_object(
...
+        return (
+            # Fetch object tags from this taxonomy whose value contains the search
+            self.objecttag_set.filter(_value__icontains=search)
+            # omit any tags whose values match the tags on the given object
+            .exclude(_value__in=excluded_tags)
+            # alphabetical ordering
+            .order_by("_value")
+            # obtain tag values
+            .values("_value", "tag")
```
pomegranited, line 437: We can alias the `_value` field to `value` to make it a little nicer for users:

```suggestion
            # Alias the `_value` field to `value` to make it nicer for users
            .annotate(value=models.F('_value'))
            # obtain tag values and IDs
            .values("value", "tag")
```

Just need to update your API tests to use `tag.get("value")` instead of `tag.get("_value")`.

pomegranited (changes requested):
This is looking really solid, @ChrisChV .
I've suggested some improvements to the docstrings and an alias to make the returned tag values nicer, but will approve once they're applied.

openedx_tagging/core/tagging/models.py
ChrisChV: Oh! Nice! :+1:

openedx_tagging/core/tagging/models.py
```diff
@@ -409,6 +413,32 @@ def tag_object(
...
         return updated_tags
 
+    def autocomplete_tags(self, search: str, object_id: str = None) -> models.QuerySet:
+        """
+        Returns tag values that contains the given search string.
+        The result is returned in alphabetical order.
+
+        The output is a QuerySet of dictionaries in with `_value` and `tag`.
+
+        Subclasses can override this method to perform their own autocomplete process.
```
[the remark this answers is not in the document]
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/59/commits/593c047b8d7bf5c3a314b8a25d19f55beb5238c8

tests/openedx_tagging/core/tagging/test_api.py
```diff
@@ -339,3 +341,126 @@ def test_get_object_tags(self):
...
+        Validate autocomplete tags
+        """
+
+        # Normal search
+        result = tagging_api.autocomplete_tags(taxonomy, search)
+        tag_values = self._get_tag_values(result)
+        for value in tag_values:
+            assert search.lower() in value.lower()
+
+        print(tag_values)
```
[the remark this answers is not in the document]
ChrisChV: Fixed here: https://github.com/openedx/openedx-learning/pull/59/commits/593c047b8d7bf5c3a314b8a25d19f55beb5238c8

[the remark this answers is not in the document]
ChrisChV: Fixed here: https://github.com/openedx/openedx-learning/pull/59/commits/593c047b8d7bf5c3a314b8a25d19f55beb5238c8

tests/openedx_tagging/core/tagging/test_api.py
```diff
@@ -339,3 +341,126 @@ def test_get_object_tags(self):
...
+            'Plantae',
+            'Monera',
+            'Gastrotrich',
+            'Placozoa',
+        ] + expected_values  # To create repeats
+        open_taxonomy = self.taxonomy
+        closed_taxonomy = tagging_api.create_taxonomy(
+            "Free_Text_Taxonomy",
+            allow_free_text=True,
+        )
```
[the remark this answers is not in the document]
ChrisChV: Fixed here: https://github.com/openedx/openedx-learning/pull/59/commits/593c047b8d7bf5c3a314b8a25d19f55beb5238c8

openedx_tagging/core/tagging/models.py
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/59/commits/593c047b8d7bf5c3a314b8a25d19f55beb5238c8

openedx_tagging/core/tagging/api.py
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/59/commits/593c047b8d7bf5c3a314b8a25d19f55beb5238c8

openedx_tagging/core/tagging/models.py
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/59/commits/593c047b8d7bf5c3a314b8a25d19f55beb5238c8

openedx_tagging/core/tagging/models.py
```diff
@@ -409,6 +413,48 @@ def tag_object(
...
+        ObjectTags linked to the given taxonomy. A case-insensitive search is used in order
+        to return the highest number of relevant tags.
+
+        If `object_id` is provided, then object tag values already linked to this object
+        are omitted from the returned suggestions. (ObjectTag values must be unique for a
+        given object + taxonomy, and so omitting these suggestions helps users avoid
+        duplication errors.).
+
+        Returns a QuerySet of dictionaries containing distinct `value` (string) and `tag`
+        (numeric ID) values, sorted alphabetically by `value`.
```
bradenmacdonald, line 428: What do you think about calling it `tag_id` instead of `tag`, if it's a numeric ID?

openedx_tagging/core/tagging/api.py
```diff
@@ -136,3 +136,37 @@ def tag_object(
     Preserves existing (valid) tags, adds new (valid) tags, and removes omitted (or invalid) tags.
     """
     return taxonomy.cast().tag_object(tags, object_id)
+
+
+def autocomplete_tags(
+    taxonomy: Taxonomy,
+    search: str,
+    object_id: str = None,
+) -> QuerySet:
```
bradenmacdonald, line 145: I'm curious to know if it's possible to use `django-stubs` [`ValuesQuerySet[...]`](https://github.com/typeddjango/django-stubs/blob/e70ec130913c499e31a0ba9dbb3ae5bc33ab9eff/django-stubs/db/models/manager.pyi#L99) to define this return type more precisely, but it's not an important thing to investigate now if you're short on time.

openedx_tagging/core/tagging/api.py
```diff
@@ -136,3 +136,37 @@ def tag_object(
...
+    """
+    Provides auto-complete suggestions by matching the `search` string against existing
+    ObjectTags linked to the given taxonomy. A case-insensitive search is used in order
```
bradenmacdonald, line 148: I think it's very unexpected that this only matches existing ObjectTags.

I know the rationale: 

> for closed taxonomies, the frontend can handle auto-completion without consulting the backend (e.g using [autocompleter](https://www.npmjs.com/package/autocompleter)), because it's already fetched all of the Tags in a Taxonomy to display the tree of options

and that makes sense, but this could still be confusing.

So I suggest that you either:
1) Put that explanation into the docstring, or
2) Add a new parameter like `applied_tags_only=True`. When it's true, it only returns tags based on the existing objecttag_set (i.e. what you already implemented), and when it's `False` it raises `NotImplementedError("(explanation)...For now we recommend fetching all of the taxonomy's tags using get_tags() and filtering them on the frontend")`

bradenmacdonald (approved):
LGTM.

While I generally recommend against returning `QuerySet`s in `api.py` methods*, I think this one makes sense because it's a ValuesQuerySet. I like how the method is returning `values` and `tag`, so that we can be more flexible in the future, e.g. returning users even if their Tags don't exist yet.

I have a few questions/suggestions, but nothing blocking.

Nice work.


<small>\* _Because: "[An app’s Django models and other internal data structures should not be exposed via its Python APIs](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0049-django-app-patterns.html#api-py)"_</small>

openedx_tagging/core/tagging/api.py
ChrisChV: `ValuesQuerySet` is [imported](https://github.com/typeddjango/django-stubs/blob/e70ec130913c499e31a0ba9dbb3ae5bc33ab9eff/django-stubs/db/models/manager.pyi#L10) from `django_stubs_ext`, and needs one more dependency and [extra configuration](https://github.com/typeddjango/django-stubs/tree/master/django_stubs_ext). I think for now it would be a lot to get that type

bradenmacdonald: OK. Thanks for checking into it!

openedx_tagging/core/tagging/models.py
ChrisChV: Yes, make sense, updated [here](https://github.com/openedx/openedx-learning/pull/59/commits/7c30263a4b5774ccb9c018ec9421b5f3d6022018#diff-204875f41064da3abfd4648611ad41c6f21abe09277263410e3daafbd343f2a3R457)

openedx_tagging/core/tagging/api.py
ChrisChV: Updated [here](https://github.com/openedx/openedx-learning/pull/59/commits/7c30263a4b5774ccb9c018ec9421b5f3d6022018#diff-951a5c4d32e1301ac1c30419ebce4ada8ae17db6c93f4c2a5a34caac9eb30e0aR174)

pomegranited:
@bradenmacdonald 

> While I generally recommend against returning QuerySets in api.py methods because: "[An app’s Django models and other internal data structures should not be exposed via its Python APIs](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0049-django-app-patterns.html#api-py)"

Huh ok.. I didn't know that, and we definitely should have picked this up in the design phase.

My assumption was that the REST APIs should be as think as possible, and so should use the Python APIs (and django model rules) to encapsulate any complexity. And since we use DRF, that means preferring the Python APIs to return QuerySets.

How should this be done instead?

pomegranited:
@ChrisChV  could you rebase this on latest main and bump the [version](https://github.com/openedx/openedx-learning/blob/main/openedx_learning/__init__.py) so we can merge?

bradenmacdonald:
@pomegranited It's not a hard and fast rule. The main problem the authors of that OEP are trying to avoid is that Django model instances are very "powerful" and once you have access to one, you can make a lot of "internal" changes that may violate the assumptions of the API - for example, you can change fields and save() it, bypassing whatever `save_changes()` API may exist. Specific example: In the tagging case, we'd generally want people to use the `tag_object()` API method, _not_ modifying or deleting an `ObjectTag` instance themselves. In fact, that sort of design is one way to avoid having so many subclasses - by forcing people to use the methods like `tagging_api.tag_object(...)` you can keep all the extra validation/creation logic in that method, and avoid the need for model subclasses.

So: How can you make the REST APIs a thin wrapper around the python API, without using Django models? You can look at [this example](https://github.com/openedx/edx-platform/blob/8beb6277b3965ab191b6a2eaa57c215f8429a9c1/openedx/core/djangoapps/content_libraries/views.py#L612-L645) in content_libraries v2 as one example I'm familiar with. Other times it may make more sense to use a query set in the REST API if it makes your life a lot easier (e.g. pagination). Yet another option is have an "internal API" that uses QuerySets and then your REST API can wrap it directly and your Python API can wrap it and convert the Django objects to non-Django data structures. As far as I know, any of those are totally fine, so just go with any approach that feels clean to you. Just keep in mind that exposing Django model instances via python will tempt people to bypass your python API, which you probably want to discourage.

ChrisChV:
> @ChrisChV could you rebase this on latest main and bump the [version](https://github.com/openedx/openedx-learning/blob/main/openedx_learning/__init__.py) so we can merge?

@pomegranited  Done :+1:

pomegranited (approved):
:+1: Thank you @ChrisChV !

- [x] I tested this by checking that the tests correctly cover the expected behavioral use cases.
- [x] I read through the code.
- [x] Includes documentation -- docstrings

openedx-webhooks:
@ChrisChV 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/59&entry.1484655318=https://github.com/ChrisChV&entry.752974735=2023-06-20+20:10&entry.1917517419=2023-08-06+23:13&entry.2133058324=Yes) so we can improve your experience in the future.

PR #62, opened 2023-07-05
Allow custom Taxonomy, ObjectTag subclasses to customize tagging behavior

pomegranited:
## Description

Adds fields and methods to Taxonomy to support use of Taxonomy subclasses to customize tagging behavior, e.g. when returning the list of available tags. Ensures that Taxonomy instances returned from the API are cast where possible, and makes it easier for these subclasses to be used.

Also adds support for ObjectTag subclasses via `cast` and `copy` methods, however full use of this must be handled by the implementing API -- here in oel_tagging, we just assume ObjectTag.

* Adds `Taxonomy._taxonomy_class` field to store the Taxonomy subclass that should be used to instantiate this instance.
* Adds methods to Taxonomy to customize ObjectTag validation based on taxonomy, tag, and object fields, to make it easier for Taxonomy subclasses to validate the parts they care about.
* Adds fields to Taxonomy for use by the system taxonomies being created: `system_defined`, and `visible_to_authors`.

## Supporting information

Part of https://github.com/openedx/modular-learning/issues/63, addresses https://github.com/openedx/edx-platform/pull/32518#discussion_r1248904313

## Testing instructions

But you can check that the added tests cover the expected polymorphism behavior supported by this change, and that tests are still covering 100% of the `openedx_tagging` module.

## Other information

See https://github.com/openedx/edx-platform/pull/32661 for this branch's use with content tagging.


openedx-webhooks:
Thanks for the pull request, @pomegranited!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_tagging/core/tagging/models.py
```diff
@@ -136,20 +139,64 @@ class Taxonomy(models.Model):
             "Indicates that tags in this taxonomy need not be predefined; authors may enter their own tag values."
         ),
     )
+    system_defined = models.BooleanField(
```
ChrisChV, line 142: @pomegranited We need to make this field non-editable after taxonomy creation. Maybe an `editable=False` can work?

openedx_tagging/core/tagging/models.py
```diff
@@ -456,3 +463,103 @@ def resync(self) -> bool:
...
+            return False
+
+        if check_object and not self._check_object():
+            return False
+
+        return True
+
+
+# Register the object tag classes in reverse order for how we want them considered
+register_object_tag_class(ObjectTag)
```
ChrisChV, line 563: @pomegranited Maybe we can explicitly add the order in this function? I say this in case object tag classes are registered in other files/modules, in those cases I would not be sure which one goes first or which one goes second

ChrisChV:
> We can do this by assigning a custom ObjectTag subclass to each system taxonomy,
and using the ObjectTag class to return candidate tags with get_tags / autocompleted tags instead of using Taxonomy.

:+1: to this. Yes, some system defined taxonomies need to overwrite get_tags

openedx_tagging/core/tagging/api.py
```diff
@@ -10,33 +10,48 @@
...
 def create_taxonomy(
-    name,
-    description=None,
+    name: str,
+    description: str = None,
     enabled=True,
     required=False,
     allow_multiple=False,
     allow_free_text=False,
+    system_defined=False,
```
ChrisChV, line 29: @pomegranited We should not be able to create system defined in this generic api. In my PR I could create another way to create them more consistently

ChrisChV:
@pomegranited In systems-defined taxonomies, one more field is necessary: ​​[visible](https://docs.google.com/document/d/1k2-ziCkCGT8AcGh2zorVinF0MH-2bF6_mAl0SFxr7EQ/edit#heading=h.86rnwzf8q3ep). Maybe we can add that field in the `Taxonomy` class? Now I have no idea how to handle that in a subclass

openedx_tagging/core/tagging/models.py
```diff
@@ -456,3 +463,103 @@ def resync(self) -> bool:
...
+        If `check_object` is False, then we skip validating the object ID/type.
+        """
+        # Must be linked to a free-text taxonomy
+        if check_taxonomy and (
+            not self.taxonomy_id or not self.taxonomy.allow_free_text
+        ):
+            return False
+
+        # Open taxonomies don't need an associated tag, but we need a value.
+        if check_tag and not self.value:
```
ChrisChV, line 511: @pomegranited Writing my code I realized that we can also use a `_check_tag` function here, similar to `_check_object`. This is because in the system defined it is only necessary to overwrite the validation of the tag. I think it would be better to add it in your PR

pomegranited: Good point -- please see [3e53687](https://github.com/openedx/openedx-learning/pull/62/commits/3e536874b07f5f4cb79b7156f87866b2dafbe10e).

ObjectTag now has 3 methods for this: `_is_valid_taxonomy`, `_is_valid_tag`, and `_is_valid_object`, which subclasses can override. And with the above change, they'll be used whenever `object_tag.is_valid()` is called as well as by `ObjectTag.valid_for`, which the object tag class registry uses to locate a valid subclass.

openedx_tagging/core/tagging/models.py
pomegranited: Sure thing, done with [f260ddd](https://github.com/openedx/openedx-learning/pull/62/commits/f260ddd8bf78840329f2ccff8862f0d9a60126e3)

openedx_tagging/core/tagging/api.py
pomegranited: Fixed with [321fd46](https://github.com/openedx/openedx-learning/pull/62/commits/321fd464bcdc4b1bc6dd39b1175180c0177b95a4).

openedx_tagging/core/tagging/models.py
pomegranited: Fixed with [321fd46](https://github.com/openedx/openedx-learning/pull/62/commits/321fd464bcdc4b1bc6dd39b1175180c0177b95a4).

openedx_tagging/core/tagging/models.py
```diff
@@ -558,6 +555,38 @@ def is_valid(self, check_taxonomy=True, check_tag=True, check_object=True) -> bo
 
         return True
 
+    def get_tag_options(self) -> Iterator[TagResult]:
```
pomegranited, line 563: @ChrisChV 

> 👍 to this. Yes, some system defined taxonomies need to overwrite get_tags

I chose to address this here, by converting the functionality from [Taxonomy.get_tags](https://github.com/openedx/openedx-learning/pull/62/commits/321fd464bcdc4b1bc6dd39b1175180c0177b95a4#diff-204875f41064da3abfd4648611ad41c6f21abe09277263410e3daafbd343f2a3L207-L233) to this ObjectTag method.

This is the ClosedObjectTag version of this, which yields actual tags.
The [superclass version yields nothing](https://github.com/openedx/openedx-learning/pull/62/commits/321fd464bcdc4b1bc6dd39b1175180c0177b95a4#diff-204875f41064da3abfd4648611ad41c6f21abe09277263410e3daafbd343f2a3R455-R462), so OpenObjectTags have no tag options by default.

System taxonomy ObjectTag subclasses can use this to generate a dynamic list of tag options as needed.

openedx_tagging/core/tagging/api.py
```diff
@@ -67,30 +96,57 @@ def resync_object_tags(object_tags: QuerySet = None) -> int:
...
+        .select_related("tag", "taxonomy")
+        .order_by("id")
+    )
+    if object_type:
+        tags = tags.filter(object_type=object_type)
+    if taxonomy:
+        tags = tags.filter(taxonomy=taxonomy)
+
+    for tag in tags:
+        # We can only validate tags with taxonomies, because we need the object_tag_class
```
pomegranited, line 146: remove: I don't think this comment is valid anymore.
```suggestion
```

openedx_tagging/core/tagging/models.py
```diff
@@ -141,11 +141,19 @@ class Taxonomy(models.Model):
...
             " taxonomy admins will not be permitted to modify them.",
         ),
     )
+    visible_to_authors = models.BooleanField(
+        default=True,
+        editable=False,
+        help_text=_(
+            "Indicates whether this taxonomy should be visible to object authors."
+        ),
+    )
```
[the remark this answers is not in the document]
ChrisChV: Thanks!

ChrisChV:
@pomegranited Looks good :+1: 

- [X] I tested this by running the tests.
- [X] I read through the code and tests with care

ChrisChV (approved):

pomegranited:
Hi @ormsbee CC @bradenmacdonald Ready for your eyes :)

openedx_tagging/core/tagging/models.py
```diff
@@ -456,3 +490,95 @@ def resync(self) -> bool:
...
+        return bool(self._value)
+
+    def _check_object(self):
+        """
+        Returns True if this ObjectTag has a valid object.
+
+        Subclasses should override this method to perform any additional validation for the particular type of object tag.
+        """
+        # Must have a valid object id/type:
+        return self.object_id and self.object_type
```
pomegranited, line 530: @ChrisChV So, I haven't yet found a use for this `ObjectTag.object_type` field..

* The ObjectTag subclasses that require UsageKey/CourseKey formatting for `course_id` can get the object type from the object key.
* We'll be creating a sytem-defined Taxonomy which automatically creates ObjectTags to mark "Content Type".

However, we might need a way to associate Taxonomy instances with the various object types they support. E.g, the [system-defined Language Taxonomy may be used on Course, Block, Library, and Library-type content items](https://github.com/open-craft/openedx-learning/pull/3#discussion_r1258942677). So when we're in the Course editor and we ask the API "What taxonomies are available here?", we'll need to know which ones are enabled for use with Courses.

I don't know if this is required for MVP though, since I think all the taxonomies discussed will be used by all the different content types. But Taxonomy Admins might want to create some that are only used for Blocks, for example.

So if we do need to support that, I'd propose using a many-to-many mapping like we did with [TaxonomyOrg](https://github.com/openedx/edx-platform/pull/32661/files#diff-8a644422ddd022128d13b6ca7c8b672a33d048d1816a80a80491246bdd76f607R13), and extending the [`get_taxonomies` API](https://github.com/openedx/openedx-learning/pull/62/files#diff-951a5c4d32e1301ac1c30419ebce4ada8ae17db6c93f4c2a5a34caac9eb30e0aR55) to support querying by object type.

ChrisChV: @pomegranited It's a great idea!. On the [MVP requeriments ](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/3695247491/Content+Tagging+Approach+MVP#Features-%26-Requirements) there is:

> Course Settings: Content Authors can turn taxonomies (instance and org-levels) on/off at the course level.

I feel that your proposal can also be adapted to that requirement. But I feel like it would have to be a many-to-many relationship with [TaxonomyOrg](https://github.com/openedx/edx-platform/pull/32661/files#diff-8a644422ddd022128d13b6ca7c8b672a33d048d1816a80a80491246bdd76f607R13) and extend the [get_taxonomies_for_org](https://github.com/openedx/edx-platform/pull/32661/files#diff-8a644422ddd022128d13b6ca7c8b672a33d048d1816a80a80491246bdd76f607R13).

openedx_tagging/core/tagging/models.py
```diff
@@ -456,3 +490,95 @@ def resync(self) -> bool:
             changed = True
 
         return changed
+
+
+class OpenObjectTag(ObjectTag):
+    """
+    Free-text object tag.
+
+    Only needs a free-text taxonomy and a value to be valid.
```
bradenmacdonald, line 499: OK, sorry for the slow response, but I'm definitely confused here.

I thought the plan was to have different subclasses of `ObjectTag` based on what type of content object was being tagged:

> Maybe it's ObjectTag that needs the subclassing? It could be made abstract, and have subclasses like CourseTag, BlockTag, etc? And instead of having a completely generic object_id field, we make the object_id return a different real field value in the various subclasses (sort of like what pk does for Django models)?

> But to tag content in the LMS/Studio, we need to enforce ``object_id`` as a CourseKey or UsageKey type. So to do this, we subclass ``ObjectTag``, and use the ``openedx.features.tagging.registry`` to register the subclass, so that it will be picked up when this tagging API creates or resyncs object tags.

And I see that _is_ happening here, but we also have "open" vs. "closed" object tags, so we end up with `OpenCourseObjectTag`, `OpenBlockObjectTag`, `ClosedCourseObjectTag`, `ClosedBlockObjectTag`. 

I thought that the "type" of `ObjectTag` being used would be stored in the ObjectTag row itself, but it seems like it's in `Taxonomy.object_tag_class`. In that case, how could we use, say, a Language (System) Taxonomy to tag both courses and blocks (and anything else that might come up in the future)?

The split between `OpenObjectTag` and `ClosedObjectTag` seems strange to me, and I don't see any requirement for this in the current code. Could we not just have one `ObjectTag` base class with

```
    def _check_tag(self):
        """
        Returns True if this ObjectTag has a valid tag value.
        Subclasses should override this method to perform any additional validation for the particular type of object tag.
        """
        # Open taxonomies don't need an associated tag, but we need a value.
        return bool(self._value) if self.taxonomy.allow_free_text else bool(self.tag_id)
```
and then subclassing is only needed for different types of objects, not different types of taxonomy?

Or is there a more significant implementation difference that I'm not seeing?


I also wonder if there are three types of Taxonomy (at the moment), not two:
* "Free-form Taxonomy" - no `Tag` values exist, and the taxonomy is just whatever `ObjectTag._value`s are in use.
* "Closed Taxonomy" - There is a list of allowed `Tag` values stored in the database, though some ObjectTags may have newer or older tag values, hence they keep a text copy of the tag in `_value`
* "Virtual Taxonomy" which is generated dynamically by python code, e.g. get a list of enabled languages (for "Language" taxonomy), get list of authors, get list of organizations, etc. The `Tag` values don't necessarily need to exist in the database, but they could if that makes the implementation easier rather than more complex. Personally I wonder if storing them in the database requires too much synchronization effort.

pomegranited: @bradenmacdonald I'm so confused now. I need more help than just these reviews.

bq. The split between OpenObjectTag and ClosedObjectTag seems strange to me, and I don't see any requirement for this in the current code. Could we not just have one ObjectTag base class with

I follow you here -- it's an artificial distinction that doesn't warrant this complexity.

bq. I thought that the "type" of ObjectTag being used would be stored in the ObjectTag row itself, but it seems like it's in Taxonomy.object_tag_class.

This was my attempt at linking the ObjectTag custom classes to their taxonomies, so that we validate a given ObjectTag for use with a taxonomy. B how is the base `oel_tagging` library supposed to support ObjectTag subclass casting when it doesn't even know about different object types?

bq. I also wonder if there are three types of Taxonomy (at the moment), not two:

Yep -- the "system-defined" taxonomies in the MVP are "virtual taxonomies", and we're not creating Tags for all of them: [ModelObjectTags](https://github.com/open-craft/openedx-learning/pull/3/files#diff-2ecc1274a15be58e16fa617e6e5860e1b2815b6cbeffa200daea9676c07c4bfcR91) keep an association to another Model (like Organization or User). 

But since we're trying to subclass ObjectTag instead of Taxonomy here, the logic for this is all down in the ObjectTag subclasses, and this doesn't seem clean either.

I've scheduled a meeting to discuss, because however we're doing these designs and discussions isn't working :(

docs/decisions/0007-tagging-app.rst
```diff
@@ -19,15 +19,15 @@ Taxonomy
...
+So in edx-platform, we will create the ``openedx.features.content_tagging`` app, to contain the models and logic for linking Organization owners to Taxonomies. There's no need to subclass Taxonomy here; we can enforce access using the ``content_tagging.api``.
 
 ObjectTag
 ~~~~~~~~~
 
 Similarly, the ``openedx_tagging`` module defined ``openedx_tagging.core.models.ObjectTag``, also self-contained to the
 ``openedx_tagging`` app.
 
-But to tag content in the LMS/Studio, we create ``openedx.features.tagging.models.ContentTag``, which subclasses ``ObjectTag``, and can then reference functionality available in the platform code.
+But to tag content in the LMS/Studio, we need to enforce ``object_id`` as a CourseKey or UsageKey type. So to do this, we subclass ``ObjectTag``, and use the ``openedx.features.tagging.registry`` to register the subclass, so that it will be picked up when this tagging API creates or resyncs object tags.
```
bradenmacdonald, line 30: If `ObjectTag.object_id` is immutable (as I think it should be), then this is really just a one-time validation that needs to happen when the `ObjectTag` is first created. In that case, a subclass is probably overkill, as there is no ongoing difference in behavior for a Course ObjectTag vs. a Block ObjectTag. You just need some way to validate the `object_id` during initial creation of `ObjectTag`, but that logic doesn't need to live in openedx-learning's tagging core.

You can absolutely keep the proxy models you've made as a nice way to do that, and they can also provide a secondary benefit of checking the `object_id` format when you're loading tags. For example, if you're writing some code that is strictly dealing with XBlocks, you can choose to use `BlockObjectTag.objects.filter(...)` to load them, and `BlockObjectTag.objects.create(...)` to create a new tag, and get that extra layer of validation. But `openedx-learning` doesn't really need to know that you're doing that.

So: what I think you don't really need is the `object_tag_class` field or `cast_object_tag` behavior, since there's no ongoing difference in behavior once the ObjectTags are saved to the database.

Also, if we say that `object_id` is always an [opaque key](https://github.com/openedx/opaque-keys/tree/master/opaque_keys/edx), then `object_type` is somewhat redundant as it can be derived from the prefix of the `object_id`. If we're not using it now, I'd suggest removing it and we can always add it later if needed.

docs/decisions/0012-system-taxonomy-creation.rst
```diff
@@ -4,25 +4,23 @@
...
-System-defined Taxonomy creation
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+System Tag lists and validation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Each System-defined Taxonomy has its own class, which is used for tag validation (e.g. ``LanguageSystemTaxonomy``, ``OrganizationSystemTaxonomy``).
-Each can overwrite ``get_tags``; to configure the valid tags, and ``validate_object_tag``; to check if a list of tags are valid.
-Both functions are implemented on the ``Taxonomy`` base class, but can be overwritten to handle special cases.
+Each System-defined Taxonomy will have its own ``ObjectTag`` subclass which is used for tag validation (e.g. ``LanguageObjectTag``, ``OrganizationObjectTag``).
+Each subclass can overwrite ``get_tags``; to configure the valid tags, and ``is_valid``; to check if a list of tags are valid.  Both functions are implemented on the ``ObjectTag`` base class, but can be overwritten to handle special cases.
```
bradenmacdonald, line 19: I think this is a conflict in requirements. We need to decide if `ObjectTag` is subclassed/specialized to handle different types of objects (`BlockObjectTag`/`CourseObjectTag`) or different types of Taxonomies (`LanguageObjectTag`/`OrganizationObjectTag`). I don't think it's possible or at least not very clean to have both dimensions of subclass.

bradenmacdonald: Personally, I don't think it makes sense to have specialized `ObjectTag` subclasses for System Taxonomies.

I think that special logic belongs in the Taxonomy subclasses, which don't necessarily have to be django model subclasses but can just be proxies or functions.


When displaying tags in the LMS or Studio, we would often want to filter out invalid tags (e.g. from an imported course) or at least display them differently, so I think it's important that `get_object_tags(..., valid_only=True)` is extremely efficient, ideally only using a single database query. For example, if there were a bunch of Author Tags, we wouldn't want each AuthorObjectTag tag making a separate DB query to check if the user still exists in the system. By having a single ObjectTag implementation with the simple logic that you already have in this PR (in `ObjectTag. _check_tag`), we avoid that problem.

One thing to think about is that for the system taxonomies like Language, Author, etc., we don't necessarily have to have them all in the database nor keep them strictly free-form. For Language or Author in particular, the system creates those tags automatically and it could just create the `Tag` for the required language or author if it doesn't exist, right before it creates an `ObjectTag`. Also for Language and Author, we need to do some translating before displaying the tags to the users - displaying the localized name of the language and displaying the username instead of the user ID. I think that sort of translation/lookup should happen in the Taxonomy level or higher so that it can be properly batched, not done on a per-tag basis in case there's several tags of the same type.

Likewise if we ever allow users to manually tag language or author, even if all the possible authors don't exist as Tag instances or free-form values, we could still override an `autocomplete` or `list_available_tags` API for that taxonomy to provide all the possible authors, and then when the ObjectTag is about to be saved, we could create the corresponding `Tag` in the database. But again that feels like behavior that is specific to the taxonomy, not the ObjecTag to me.

bradenmacdonald: I like most of the cleanups in this PR, but my recommendation after thinking about it (which you can ignore or push back on!) would be:
* Remove `object_tag_class` and `cast_object_tag`
* Remove references to specialized ObjectTag subclasses
* Consolidate the migrations
And then just merge it as a nice "cleanup" PR.

docs/decisions/0007-tagging-app.rst
pomegranited: Done: https://github.com/openedx/openedx-learning/pull/62/files#diff-f181041ce998b0b874170c0564ba6c7c7e4cc88e1a28d0f2cf928160e3c3f9d2R56-R59

openedx_tagging/core/tagging/models.py
```diff
@@ -136,20 +140,109 @@ class Taxonomy(models.Model):
...
+        help_text=_(
+            "Indicates whether this taxonomy should be visible to object authors."
+        ),
+    )
+    _taxonomy_class = models.CharField(
+        null=True,
+        max_length=255,
+        help_text=_(
+            "Taxonomy subclass used to instantiate this instance."
+            "Must be a fully-qualified module and class name.",
```
bradenmacdonald, line 163: What will the behavior be if the taxonomy class is no longer found? It would be good to state that explicitly here.

Plugins will come and go, so we need to be sure that falling back to the "generic" behavior will always at least preserve existing tags, but perhaps not allow new tags to be added. I believe you already have implemented this and have a test for it; I just think it's good to explain here as well.

openedx_tagging/core/tagging/models.py
```diff
@@ -195,38 +288,72 @@ def validate_object_tag(
         """
         Returns True if the given object tag is valid for the current Taxonomy.
 
-        Subclasses can override this method to perform their own validation checks, e.g. against dynamically generated
-        tag lists.
+        Subclasses should override the internal _validate* methods to perform their own validation checks, e.g. against
+        dynamically generated tag lists.
```
bradenmacdonald, line 292: It would be good to have a fully dynamic taxonomy (perhaps prime numbers or fibonacci numbers, something like that) in the test suite here. That can come in a follow-up PR though, based on learnings from the system taxonomies PR.

bradenmacdonald (approved):

docs/decisions/0007-tagging-app.rst
```diff
@@ -19,15 +19,15 @@ Taxonomy
...
+So in edx-platform, we will create the ``openedx.features.content_tagging`` app, to contain the models and logic for linking Organization owners to Taxonomies. Here, we can subclass ``Taxonomy`` as needed, preferably using proxy models. The APIs are responsible for ensuring that any ``Taxonomy`` instances are cast to the appropriate subclass.
 
 ObjectTag
 ~~~~~~~~~
 
 Similarly, the ``openedx_tagging`` module defined ``openedx_tagging.core.models.ObjectTag``, also self-contained to the
 ``openedx_tagging`` app.
 
-But to tag content in the LMS/Studio, we create ``openedx.features.tagging.models.ContentTag``, which subclasses ``ObjectTag``, and can then reference functionality available in the platform code.
+But to tag content in the LMS/Studio, we need to enforce ``object_id`` as a CourseKey or UsageKey type. So to do this, we subclass ``ObjectTag``, and use this class when creating object tags for the content taxonomies. Once the ``object_id`` is set, it is not editable, and so this key validation need not happen again.
```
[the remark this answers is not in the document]
pomegranited: Noted, and I agree. And since the system taxonomies will be implemented in this library, they'll be separate from the "content" concerns anyway, and so can be re-used for user tagging.

I will at least remove the phrase "content taxonomies" from these docs.

openedx_tagging/core/tagging/models.py
```diff
@@ -136,20 +140,109 @@ class Taxonomy(models.Model):
...
+    def taxonomy_class(self, taxonomy_class: Union[Type, None]):
+        """
+        Assigns the given taxonomy_class's module path.class to the field.
+
+        Raises ValueError if the given `taxonomy_class` is a built-in class; it should be a Taxonomy-like class.
+        """
+        if taxonomy_class:
+            if taxonomy_class.__module__ == "builtins":
+                raise ValueError(
+                    f"Unable to assign taxonomy_class for {self}: {taxonomy_class} must be a class like Taxonomy"
```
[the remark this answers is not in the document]
pomegranited: I think I was holding onto the idea that we might have non-model classes for this stuff.. But I'll let that go and make this clearer.

openedx_tagging/core/tagging/models.py
pomegranited: Yep, I'll make sure this test case is covered by the system taxonomy PR.

openedx_tagging/core/tagging/models.py
```diff
@@ -456,3 +608,22 @@ def resync(self) -> bool:
...
+    @classmethod
+    def cast(cls, object_tag: "ObjectTag") -> "ObjectTag":
+        """
+        Returns a cls instance with the same properties as the given ObjectTag.
+        """
+        return cls().copy(object_tag)
+
+    def copy(self, object_tag: "ObjectTag") -> "ObjectTag":
+        """
+        Copy the fields from the given Taxonomy into the current instance.
```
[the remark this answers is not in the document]
pomegranited: Fixed: https://github.com/openedx/openedx-learning/pull/62/commits/1094d24c5f99a375144b13171c88efec27a9244b#diff-204875f41064da3abfd4648611ad41c6f21abe09277263410e3daafbd343f2a3R621

openedx_tagging/core/tagging/models.py
pomegranited: Good point, done: https://github.com/openedx/openedx-learning/pull/62/commits/1094d24c5f99a375144b13171c88efec27a9244b#diff-204875f41064da3abfd4648611ad41c6f21abe09277263410e3daafbd343f2a3R162-R163

ormsbee (approved):

openedx_tagging/core/tagging/models.py
ChrisChV: Yes, I will keep that in mind now that I make the changes in my PR

bradenmacdonald:
Thanks @ormsbee. Could you please hit merge on this? It seems I don't have permission and I'm not sure who else does.

pomegranited:
Huge thank you @bradenmacdonald , @ChrisChV and @ormsbee ! I think retaining the discussions here is important for this PR, but I don't want to muddy the `main` branch, so will squash + merge now.

openedx-webhooks:
@pomegranited 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/62&entry.1484655318=https://github.com/pomegranited&entry.752974735=2023-07-05+07:29&entry.1917517419=2023-07-20+00:15&entry.2133058324=Yes) so we can improve your experience in the future.

PR #63, opened 2023-07-10
Taxonomy view/management REST APIs

rpenido:
## Description
This PR adds support for creating, viewing, updating and deleting Taxonomies via REST APIs, with the permissions applied accordingly.

## Supporting Information
* Closes https://github.com/openedx/modular-learning/issues/73

## Testing instructions
* Ensure that all endpoints in the related issue are implemented.
* Ensure that the tests cover the expected behaviour of the view and permissions as described in the related issue.

## To do before merge
- [ ] squash and merge this change (don't just merge)

### DIFF
--- a/openedx_tagging/core/tagging/urls.py
+++ b/openedx_tagging/core/tagging/urls.py
@@ -1,0 +1,10 @@
+"""
+Tagging API URLs.
+"""
+
+from django.urls import path, include
+
+from .rest_api import urls
+
+app_name = "oel_tagging"
+urlpatterns = [path("", include(urls))]
--- a/tests/openedx_tagging/core/tagging/test_views.py
+++ b/tests/openedx_tagging/core/tagging/test_views.py
@@ -1,0 +1,365 @@
+"""
+Tests tagging rest api views
+"""
+
+import ddt
+from django.contrib.auth import get_user_model
+from django.urls import reverse
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from openedx_tagging.core.tagging.models import Taxonomy
+
+User = get_user_model()
+
+
+def check_taxonomy(
+    data,
+    id,
+    name,
+    description=None,
+    enabled=True,
+    required=False,
+    allow_multiple=False,
+    allow_free_text=False,
+    system_defined=False,
+    visible_to_authors=True,
+):
+    assert data["id"] == id
+    assert data["name"] == name
+    assert data["description"] == description
+    assert data["enabled"] == enabled
+    assert data["required"] == required
+    assert data["allow_multiple"] == allow_multiple
+    assert data["allow_free_text"] == allow_free_text
+    assert data["system_defined"] == system_defined
+    assert data["visible_to_authors"] == visible_to_authors
+
+
+@ddt.ddt
+class TestTaxonomyViewSet(APITestCase):
+    def setUp(self):
+        super().setUp()
+
+        self.user = User.objects.create(
+            username="user",
+            email="user@example.com",
+        )
+
+        self.staff = User.objects.create(
+            username="staff",
+            email="staff@example.com",
+            is_staff=True,
+        )
+
+    @ddt.data(
+        (None, status.HTTP_200_OK, 3),
+        (1, status.HTTP_200_OK, 2),
+        (0, status.HTTP_200_OK, 1),
+        (True, status.HTTP_200_OK, 2),
+        (False, status.HTTP_200_OK, 1),
+        ("True", status.HTTP_200_OK, 2),
+        ("False", status.HTTP_200_OK, 1),
+        ("1", status.HTTP_200_OK, 2),
+        ("0", status.HTTP_200_OK, 1),
+        (2, status.HTTP_400_BAD_REQUEST, None),
+        ("invalid", status.HTTP_400_BAD_REQUEST, None),
+    )
+    @ddt.unpack
+    def test_list_taxonomy_queryparams(self, enabled, expected_status, expected_count):
+        Taxonomy.objects.create(name="Taxonomy enabled 1", enabled=True).save()
+        Taxonomy.objects.create(name="Taxonomy enabled 2", enabled=True).save()
+        Taxonomy.objects.create(name="Taxonomy disabled", enabled=False).save()
+
+        url = reverse("oel_tagging:taxonomy-list")
+
+        self.client.force_authenticate(user=self.staff)
+        if enabled is not None:
+            response = self.client.get(url, {"enabled": enabled})
+        else:
+            response = self.client.get(url)
+        assert response.status_code == expected_status
+
+        # If we were able to list the taxonomies, check that we got the expected number back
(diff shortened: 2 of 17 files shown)


openedx-webhooks:
Thanks for the pull request, @rpenido!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

pomegranited:
@rpenido Ok to leave this PR here in Draft, but until https://github.com/openedx/openedx-learning/pull/58 is merged, do you mind to submit your work to a open-craft PR against that branch instead? Then we can see the diff between what you're adding and what that PR adds.

rpenido:
@pomegranited Updated this PR. I think we are ready to review here!

openedx_tagging/core/tagging/urls.py
```diff
@@ -0,0 +1,10 @@
+"""
+Tagging API URLs.
+"""
+
+from django.urls import path, include
+
+from .rest_api import urls
+
+app_name = "oel_tagging"
```
rpenido, line 9: Removing this breaks the tests. Is there a way to set the app_name only for tests?

pomegranited: Nah, it's ok to leave this here.

pomegranited (approved):
Hi @rpenido this is looking really good!

I've suggested two really minor changes, but this is ready for upstream review once they're addressed.

:+1:

- [x] I tested this by running the dev server:
   ```
   # Run migrations
   python3 manage.py migrate
   # Create a superuser
   python3 manage.py createsuperuser # and used the shell to create non-superusers
   # Run the dev server
   python3 manage.py runserver
   # Login using Django Admin (non-staff users can login, they just can't see Django Admin)
   # http://127.0.0.1:8000/admin/
   # REST API is here:
   # http://127.0.0.1:8000/tagging/rest_api/v1/
   ```
- [x] I read through the code
- [x] Includes documentation

pomegranited:
Thanks for applying those changes @rpenido ! This is ready for @ormsbee and/or @bradenmacdonald 's review now.

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -0,0 +1,365 @@
...
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    @ddt.data(
+        (None, status.HTTP_403_FORBIDDEN),
+        ("user", status.HTTP_403_FORBIDDEN),
+        ("staff", status.HTTP_201_CREATED),
+    )
+    @ddt.unpack
+    def test_create_taxonomy(self, user_attr, expected_status):
+        url = reverse("oel_tagging:taxonomy-list")
```
bradenmacdonald, line 146: Nit: In my opionion, it's better to hard-code the REST API URLs in tests, so that if you change the URL (breaking backwards compatibility), the test will fail. If you use reverse, you can change the URLs which breaks backwards compatibility, but the tests will still pass.

This is an optional piece of feedback.

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -0,0 +1,365 @@
...
+                    "enabled": False,
+                    "required": False,
+                },
+            )
+
+    @ddt.data(
+        (False, False, status.HTTP_200_OK),
+        (False, True, status.HTTP_200_OK),
+        (True, False, status.HTTP_403_FORBIDDEN),
+        (True, True, status.HTTP_403_FORBIDDEN),
```
bradenmacdonald, line 307: We allow changing a taxonomy from system_defined=False to system_defined=True, but not vice versa? That seems weird.

If we say that "Even taxonomy admins cannot change system taxonomies." then I don't think anyone should be allowed to create them via the API either.

bradenmacdonald (commented):
From a quick look, this looks fine. Just had two questions. I don't need to review this again; @pomegranited can give final approval and merge.

tests/openedx_tagging/core/tagging/test_views.py
pomegranited: That's a good point.. @rpenido do you have time to make this change to your tests this sprint? If not, then we can leave this out.

I recommend putting the hard-coded strings as constants up at the top and use string formatting, e.g.

TAXONOMY_LIST_URL = `/tagging/rest_api/v1/taxonomy/'
TAXONOMY_VIEW_URL = f`/tagging/rest_api/v1/taxonomy/{taxonomy.pk}'

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -0,0 +1,365 @@
...
+
+        self.client.force_authenticate(user=self.staff)
+        response = self.client.patch(
+            url, {"system_defined": update_value}, format="json"
+        )
+        assert response.status_code == expected_status
+
+        if status.is_success(expected_status):
+            response = self.client.get(url)
+            assert response.data["system_defined"] == False
```
pomegranited, line 327: Could you make this test clearer with:

```suggestion
        # Verify that system_defined has not changed
        response = self.client.get(url)
        assert response.data["system_defined"] == create_value
```

tests/openedx_tagging/core/tagging/test_views.py
pomegranited: > We allow changing a taxonomy from system_defined=False to system_defined=True, but not vice versa? That seems weird.

It's OK, that's not what this test does.. It shows that `system_defined` is _always ignored_ when passed as part of the taxonomy patch data, so it can't be changed via the REST API.

However, we can make this clearer with [this suggestion](https://github.com/openedx/openedx-learning/pull/63/files#r1275737975).

> If we say that "Even taxonomy admins cannot change system taxonomies." then I don't think anyone should be allowed to create them via the API either.

Yep, that's already true, see [test_create_taxonomy_system_defined](https://github.com/openedx/openedx-learning/pull/63/files#diff-2f11bfdf39bf0bee3b4778c0d84478fa6afd170e57245f804c77b2e9b4d865a4R184-R194).

rpenido: The different status is because for `system_defined=True`, we have a rule (status=403):
https://github.com/openedx/openedx-learning/blob/a7bc37af850cbdc56537d7c8455a050423049708/openedx_tagging/core/tagging/rules.py#L25-L32

For `system_defined=False`, we just drop the value because of the `editable=False` in the model field (status=200):
https://github.com/openedx/openedx-learning/blob/a7bc37af850cbdc56537d7c8455a050423049708/openedx_tagging/core/tagging/models.py#L143-L150

rpenido: Done: 2ccae75d75227729a3beca0b010d0995e4430697

rpenido:
@pomegranited I made the changes with new commits: 2ccae75d75227729a3beca0b010d0995e4430697 and cec362d2cfc5342949e743cd6097c1ca750d8dd5. That way, I think it is easier to review.

Let me know if I should squash the commits, or you can do it yourself.

Thank you!

PR #67, opened 2023-07-21
System defined taxonomies

ChrisChV:
## Description

Subclasses related to build System-defined taxonomies:

- [Docs updated](https://github.com/openedx/openedx-learning/pull/67/files#diff-e20a5299480027adca1a87aac4dea6d00797403cc0a23aab43bf0204ddb3a069) to reflect the new architecture of taxonomies.
- `models.py` moved to `models/base.py`
- `tag_ref` [setter](https://github.com/openedx/openedx-learning/pull/67/files#diff-7ffb9100f9426c81c59e88e9beceb9e3a8089da152d1d605b7de97990e6f0d50R540) to allow override how to set a tag on an object_tag
- `object_tag_class` property [added to Taxonomy](https://github.com/openedx/openedx-learning/pull/67/files#diff-7ffb9100f9426c81c59e88e9beceb9e3a8089da152d1d605b7de97990e6f0d50R183)
- Base system-defined subclasses added.
- Language tags fixture.
- Taxonomy.system_defined updated to be a [hardcoded property](https://github.com/openedx/openedx-learning/pull/67/commits/1847248b48176014b989a25e03ec00b9dda5f26e#diff-7ffb9100f9426c81c59e88e9beceb9e3a8089da152d1d605b7de97990e6f0d50R203).

## Testing instructions

But you can check that the added tests cover the expected polymorphism behavior supported by this change, and that tests are still covering 100% of the openedx_tagging module.

## Other information

Relates to https://github.com/openedx/modular-learning/issues/77


openedx-webhooks:

- supporting documentation

pomegranited (approved):
:+1: This is looking great @ChrisChV !

I left some suggestions and a nits, but if you could address them, I think it's ready for upstream review. Feel free to merge https://github.com/open-craft/openedx-learning/pull/8 or close and rewrite as you see fit?

- [x] I tested this with the PR test instructions, and by ensuring code coverage for automated tests.
- [x] I read through the code and tests
- [x] Includes documentation

openedx_tagging/core/tagging/admin.py
```diff
@@ -1,8 +1,99 @@
 """ Tagging app admin """
+from django import forms
 from django.contrib import admin
 
 from .models import ObjectTag, Tag, Taxonomy
 
-admin.site.register(Taxonomy)
-admin.site.register(Tag)
+
+def check_taxonomy(taxonomy: Taxonomy):
```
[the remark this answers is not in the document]
ChrisChV: Ok, I understand the change

pomegranited:
Thanks for applying those changes @ChrisChV ! I think this is ready for @ormsbee and/or @bradenmacdonald 's eyes now.

openedx_tagging/core/tagging/models/system_defined.py
```diff
@@ -0,0 +1,267 @@
...
+        Subclasses can override this method to perform their own instance validation checks.
+        """
+        object_tag = self.object_tag_class.cast(object_tag)
+        return bool(object_tag.get_instance())        
+
+    def _check_tag(self, object_tag: ObjectTag) -> bool:
+        """
+        Returns True if the instance is valid
+        """
+        return super()._check_tag(object_tag) and self._check_instance(object_tag)
```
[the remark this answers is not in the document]
bradenmacdonald: During day-to-day operations, does it even matter if there are ModelObjectTags that refer to non-existent objects?

Certainly when displaying tags for a particular (existing) content object, I think we want to be able to hide invalid tags whose taxonomy doesn't exist or doesn't match. But if there are a bunch of extra ObjectTags in the database pointing to deleted content, does it really matter? We can just have some kind of cleanup script that runs once in a while, to delete them.

openedx_tagging/core/tagging/models/base.py
```diff
@@ -431,6 +447,7 @@ class ObjectTag(models.Model):
     id = models.BigAutoField(primary_key=True)
     object_id = case_insensitive_char_field(
```
bradenmacdonald, line 448: I just noticed this, but why is `object_id` case insensitive? Shouldn't it be case sensitive? `abc` and `ABC` are usually different IDs.

openedx_tagging/core/tagging/models/system_defined.py
```diff
@@ -0,0 +1,271 @@
...
+    @property
+    def system_defined(self) -> bool:
+        """
+        Indicates that tags and metadata for this taxonomy are maintained by the system;
+        taxonomy admins will not be permitted to modify them.
+        """
+        return True
+
+
+class ModelObjectTag(ObjectTag):
```
bradenmacdonald, line 33: How does this interact with [`ContentObjectTag`](https://github.com/open-craft/edx-platform/blob/76d8dcca420d45000ea9ae9326c3ab9a6de30f95/openedx/features/content_tagging/models.py#L85-L104)?

I don't really like that sometimes the type of taxonomy (system-defined or not) determines which subclass of `ObjectTag` to use (`ModelObjectTag`), and other times the type of content (XBlock or Course) determines it (`ContentObjectTag`). I know that they're both driven by the taxonomy actually, since ContentTaxnonomy is a type of Taxonomy too, but it seems like there's some conceptual confusion. (I'm not asking for changes just yet, but trying to understand what the idea is here).

bradenmacdonald (commented):
I probably won't get time for a full review today, but I started reading through the code.

openedx_tagging/core/tagging/models/system_defined.py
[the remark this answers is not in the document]
bradenmacdonald: Oh sorry, I thought this was talking about tagging non-existing _content_ but it's talking about verifying if the Tag is valid (e.g. an author tag on an XBlock is invalid if the author has been deleted). Yeah we do want to check that.

docs/decisions/0012-system-taxonomy-creation.rst
```diff
@@ -30,26 +31,28 @@ We have two ways to handle Tags creation and validation for System-defined Taxon
...
 
-This taxonomy depends on a core data model, but simplifies the creation of Tags by allowing free-form tags,
-but we can validate the tags using the ``validate_object_tag`` method. For example we can put the ``AuthorSystemTaxonomy`` associated with
-the ``User`` model and use the ``ID`` field as tags. Also we can validate if an ``User`` still exists or has been deleted over time.
+Closed Taxonomies that depends on a core data model. Ex. AuthorTaxonomy with Users as Tags
+
+#. Tags are created on the fly when new ObjectTags are added.
+#. Tag.external_id we store an identifier from the instance (eg. User.pk).
+#. Tag.value we store a human readable representation of the instance (eg. User.username).
+#. Resync the tags to re-fetch the value.
```
bradenmacdonald, line 45: OK, I definitely like the approach described here ^^

But what I'm still caught up on is `ModelObjectTag`. Do we need to subclass ObjectTag and push this logic into it? From reading the approach described here, and from what we discussed, I assumed we'd happily just use the normal `ObjectTag` or even `ContentObjectTag`. Because we're creating `Tag` instances as needed, and the regular `ObjectTag` already works correctly with `Tag` instances in Taxonomies.

Of course you need some logic to:
* **list available tags** even before they've been created as `Tag` instances: e.g. list all known users or languages → this belongs in the Taxonomy subclass but I don't think it's even necessary for this project, because we never need to display the "available" tags of system taxonomies to users. The tags will get auto-created by the system and the system already knows what user/language is in use. There's no UI and no dropdown menu to show.
*  **auto-create `Tag` instances** when tagging an object, if they don't yet exist. Sure, but this logic could live anywhere - in the `edx-platform` code that's applying the User and Language tags, in the `tag_object()` API, in the `Taxonomy.tag_object()` API - I don't think it is necessary to put this into an `ObjectTag` subclass, and doing so creates a lot of complexity.
   * Think MVP, and what's the simplest solution? Just auto-create the tag in the same place you apply the tag to some content. No special behavior needed in this repo at all. In fact I almost wonder if the User Taxonomy is just a regular closed taxonomy that's not editable, and the platform code can create `Tag`s in it just-in-time... What other functionality do we need here? 
* **auto-delete `Tag` instances** when the upstream object is deleted: we don't really need to worry about this for now, as languages won't be deleted and user deletion is very rare. In fact, I don't think user deletion even really happens; it's just retirement/anonymization that happens per GDPR. So in the future we could add a signal listener that detects "user retirement" and then deletes any `Tag` and `ObjectTag`s associated with the user, but I would leave that out of scope for now. We don't even really know how/where such user tags will be used yet.

openedx_tagging/core/tagging/models/system_defined.py
pomegranited: Hmm.. maybe ModelObjectTag's functionality should be in a mixin instead, so that we can have ModelContentObjectTags..

But that probably also means that the SystemTaxonomy / ModelSystemTaxonomy functionality should be in a mixin too, same reason.. is this getting to complicated?

docs/decisions/0012-system-taxonomy-creation.rst
pomegranited: I agree with most of what you've said here, except for:

> Just auto-create the tag in the same place you apply the tag to some content. No special behavior needed in this repo at all.

I think providing a `ModelTaxonomy` (or mixin) that handles Tag auto-creation is broadly useful.

And I think it belongs in this openedx-learning library, because it's not content-specific and could be useful when tagging other types of objects, like people or forum posts. However, if it makes more sense for this to live in edx-platform for the MVP, I'm OK with that, we can always move it into openedx-learning later.

openedx_tagging/core/tagging/models/base.py
pomegranited: Ah ok.. I think that came from a misunderstanding about content UsageKeys.. I thought they were case insensitive.

@ChrisChV  Do you mind to fix this here? Sorry there's been so many little mistakes like this..

pomegranited: @bradenmacdonald Are you ok with the other case-insensitive char fields defined here?

docs/decisions/0012-system-taxonomy-creation.rst
ChrisChV: > list available tags even before they've been created as Tag instances: e.g. list all known users or languages → this belongs in the Taxonomy subclass but I don't think it's even necessary for this project, because we never need to display the "available" tags of system taxonomies to users. The tags will get auto-created by the system and the system already knows what user/language is in use. There's no UI and no dropdown menu to show.

That's true except for language, it's the only [system-defined taxonomy that can be "edited"](https://docs.google.com/document/d/1k2-ziCkCGT8AcGh2zorVinF0MH-2bF6_mAl0SFxr7EQ/edit#heading=h.uj6l8t9yvrwn) in a content by the content author.

> I think providing a ModelTaxonomy (or mixin) that handles Tag auto-creation is broadly useful.

I agree with Jill too, in addition to the auto creation, it checks that the instance (User, Organization) exists to see if the tag is valid.

> auto-delete Tag instances when the upstream object is deleted: we don't really need to worry about this for now, as languages won't be deleted and user deletion is very rare. In fact, I don't think user deletion even really happens; it's just retirement/anonymization that happens per GDPR. So in the future we could add a signal listener that detects "user retirement" and then deletes any Tag and ObjectTags associated with the user, but I would leave that out of scope for now. We don't even really know how/where such user tags will be used yet.

I agree, the simplest thing is to check if the user or organization exists. 

To clarify: the ModelTaxonomy is not used for languages, it is only used for the taxonomy of users and organizations

openedx_tagging/core/tagging/models/system_defined.py
```diff
@@ -0,0 +1,271 @@
...
+    Language System-defined taxonomy
+
+    The tags are filtered and validated taking into account the
+    languages available in Django LANGUAGES settings var
+    """
+
+    class Meta:
+        proxy = True
+
+    def get_tags(self, tag_set: models.QuerySet = None) -> List[Tag]:
```
bradenmacdonald, line 242: nit: This `tag_set` parameter is unused here, and undocumented in the base class. I think the API is cleaner if you just keep it as `get_tags(self) -> List[Tag]` and move the prefetch/sort code to a helper method like `Taxonomy._sort_tag_results()`. Up to you though.

docs/decisions/0012-system-taxonomy-creation.rst
bradenmacdonald: > I think providing a ModelTaxonomy (or mixin) that handles Tag auto-creation is broadly useful.

👍🏻 Sure, good point. That's totally reasonable, and that sounds like a mixin to me. After all, in the platform if we keep the "ContentTaxonomy" vs "regular Taxonomy" distinction, it would have to be a mixin, to allow Content tagging with auto-creation of tags.

> That's true except for language, it's the only [system-defined taxonomy that can be "edited"](https://docs.google.com/document/d/1k2-ziCkCGT8AcGh2zorVinF0MH-2bF6_mAl0SFxr7EQ/edit#heading=h.uj6l8t9yvrwn) in a content by the content author.

Oh OK. I've asked for clarification on that because I heard the opposite. But in any case, it seems we agree that "list available tags" should live in the Taxonomy. (I think it should be similar to what you implemented in `LanguageTaxonomy.get_tags() -> List[Tag]` but we'll probably eventually need an API that returns tag strings/IDs, not `Tag` objects because they may not exist yet. Imagine doing an auto-complete to tag users and there are 1,000,000 users in the system - we don't want to create `Tag` objects for each user unless they're actually used as a tag. Or maybe that would be a separate `search_available_tags` API. In any case, future PR.)

> I agree with Jill too, in addition to the auto creation, it checks that the instance (User, Organization) exists to see if the tag is valid.

That's fine. I do think however that belongs in the Taxonomy, not the ObjectTag itself. Because it's mostly a question of validating the User/Org then creating the corresponding `Tag` instance. The ObjectTag can validate itself with the relation to the `Tag` and doesn't need any special logic to handle ongoing validation related to the User/Org. If necessary, the Taxonomy can have some code to keep the `Tag` instances in sync with the User/Org models, but I think we can leave that out of scope for now to keep it simple.

openedx_tagging/core/tagging/models/base.py
bradenmacdonald: Please fix that in a small, focused, separate PR. Just add it to the list you're making. I don't want to muddy the waters or delay this PR any longer than I already have, and smaller focused PRs are usually better.

I am not sure about `_name` but I guess it makes sense to be case-insensitive.
If `_value` holds IDs it should be case-sensitive I think.

tests/openedx_tagging/core/tagging/test_models.py
```diff
@@ -119,16 +138,16 @@ class TestModelTagTaxonomy(TestTagTaxonomyMixin, TestCase):
 
     def test_system_defined(self):
         assert not self.taxonomy.system_defined
-        assert self.system_taxonomy.system_defined
+        assert self.system_taxonomy.cast().system_defined
```
bradenmacdonald, line 141: Needing to remember this `.cast` is likely to be a problem in the future.

My recommendation (for you to consider in the future!) is that you add a decorator like `@auto_cast` to the base class properties like `.system_defined` that will detect when the developer forgot to `.cast()` and either auto-cast or throw an exception. Probably the latter to be more predictable.

bradenmacdonald (approved):
OK, so: I don't think there's a good justification for `ModelObjectTag` and I'd like to see you either prove that it's important to the design despite the complexity it adds, or just remove it. I think everything else we've landed on makes sense.

You can merge this now and remove ModelObjectTag soon in a separate PR, or remove it now, or justify why we want to keep it.

bradenmacdonald:
One thing to think about for the future:

I think that our conceptual model may be missing a key piece: a model to track _how_ a `Taxonomy` should be used.

Right now, there's no place to say "I want to use the Language Taxonomy to tag Content Objects belonging to X organization" or "I want users from org X to use OrgXTaxonomy to tag Course objects belonging to the organization" or "I want users from org Y to use the SpokenLanguage taxonomy to tag learners".

If we had a `TaxonomyUseCase` (?) model, it could specify who, what, why, where -> which org/users can use which taxonomy for which purpose to tag what type of objects using which ObjectTag subclass.

That would solve the weird problem now where if you use LanguageSystemTaxonomy to tag an XBlock, it gets a plain `ObjectTag`, but if you use an org's closed taxonomy, it gets a `ContentObjectTag`. I think both should be using `ContentObjectTag` if they're both tagging content objects, but currently you have to specify the ObjectTag subclass at the Taxonomy level, not as part of defining a particular use case for the taxonomy.

Anyhow, I'm not saying to do it this way or to launch a major refactor. Just keep this idea in mind while we're iterating the models in the future, in case it resonates with you or inspires something.

docs/decisions/0012-system-taxonomy-creation.rst
pomegranited: > Or maybe that would be a separate search_available_tags API. In any case, future PR.)

Yep, agreed -- and note we don't need to search model tags in MVP (User and Organization system taxonomies are our only model system taxonomies, and neither are editable by the content authors).

> I do think however that belongs in the Taxonomy, not the ObjectTag itself...but I think we can leave that out of scope for now to keep it simple.

Cool, thanks for your clarifications here @bradenmacdonald . I'll note this on my "cleanup" task.

openedx_tagging/core/tagging/models/system_defined.py
pomegranited: Funny.. that's exactly what @ChrisChV did initially, and I asked him to change it to this instead, because I think this is cleaner :)

I'll note this in my cleanup task, but to keep things from flip-flopping here, I'd rather leave it as-is for now.

Thank you @bradenmacdonald !

openedx_tagging/core/tagging/models/base.py
```diff
@@ -239,12 +260,11 @@ def copy(self, taxonomy: "Taxonomy") -> "Taxonomy":
         self.required = taxonomy.required
         self.allow_multiple = taxonomy.allow_multiple
         self.allow_free_text = taxonomy.allow_free_text
-        self.system_defined = taxonomy.system_defined
         self.visible_to_authors = taxonomy.visible_to_authors
         self._taxonomy_class = taxonomy._taxonomy_class
         return self
 
-    def get_tags(self) -> List[Tag]:
+    def get_tags(self, tag_set: models.QuerySet = None) -> List[Tag]:
```
pomegranited, line 267: I forgot to document this added parameter here -- @ChrisChV could you add `tag_set` to the docstring?

openedx_tagging/core/tagging/models/base.py
pomegranited: cf https://github.com/openedx/modular-learning/issues/84

openedx_tagging/core/tagging/models/system_defined.py
pomegranited: Cleanup task: https://github.com/openedx/modular-learning/issues/85

openedx_tagging/core/tagging/models/system_defined.py
bradenmacdonald: haha ooops... I didn't read the earlier review. And this is definitely subjective.

ChrisChV:
> You can merge this now and remove ModelObjectTag soon in a separate PR, or remove it now, or justify why we want to keep it.

@bradenmacdonald I also agree that it should be removed. @pomegranited added it to do it in a separate PR: https://github.com/openedx/openedx-learning/pull/67#discussion_r1275584286

> That would solve the weird problem now where if you use LanguageSystemTaxonomy to tag an XBlock, it gets a plain ObjectTag, but if you use an org's closed taxonomy to tag the same XBlock, it gets a ContentObjectTag. I think both should be using ContentObjectTag if they're both tagging content objects, but currently you have to specify the ObjectTag subclass at the Taxonomy level, not as part of defining a particular use case for the taxonomy.

We are currently solving this by creating[ subclasses for tagging content](https://github.com/openedx/edx-platform/pull/32869/files#diff-1f01045e9b5f4e8b46c76dec746b3a595ff3b11513c3a538175c406ec98719c0R56). But yes, I agree that there must be a simpler way to do it.

bradenmacdonald:
> We are currently solving this by creating[ subclasses for tagging content](https://github.com/openedx/edx-platform/pull/32869/files#diff-1f01045e9b5f4e8b46c76dec746b3a595ff3b11513c3a538175c406ec98719c0R56). But yes, I agree that there must be a simpler way to do it.

@ChrisChV OK, somehow I really didn't understand that until now. So to tag _content_ with the (built-in, system defined) language Taxonomy, we cast it to `ContentLanguageTaxonomy` which is **still the same taxonomy**, but we now have a different "view" of that Taxonomy which is now using `ContentObjectTag` to tag objects? Is that right? I guess that works, and I'm fine with that for now, but it seems like a strange use of subclassing to me, as there's nothing really different about `ContentLanguageTaxonomy` vs `LanguageTaxonomy` except for what type of `ObjectTag` it applies, and that only affects the `add_tag` action, not what the overall Taxonomy _is_. I thought that just using [`tag_content_object()`](https://github.com/openedx/edx-platform/blob/6598abbb6b2a07f2c4ecaf278ef109027e0e32b3/openedx/features/content_tagging/api.py#L122-L149) would be enough to apply a `ContentObjectTag` instead of an `ObjectTag`.

Like if you had a `Person` class and the person is signing autographs, and now you want the person to sign autographs using blue ink instead of black ink, it makes much more sense to say `Person.sign_autograph(book_to_sign, using_ink=Ink.Blue)` than to say `Person.cast_to(BlueInkUsingPerson).sign_autograph(book_to_sign)`.


I have already approved this PR, but I just noticed there is no test that actually applies some system-defined tags to content objects. Shouldn't we add that in?

ChrisChV:
@bradenmacdonald Thanks for your feedback!

> OK, somehow I really didn't understand that until now. So to tag content with the (built-in, system defined) language Taxonomy, we cast it to ContentLanguageTaxonomy which is still the same taxonomy, but we now have a different "view" of that Taxonomy which is now using ContentObjectTag to tag objects? Is that right?

Yes, it is. 

> I have already approved this PR, but I just noticed there is no test that actually applies some system-defined tags to content objects. Shouldn't we add that in?

Yes, sorry, I added the system defined to the api test, that includes tag objects: 034726df9374ca549c4947b76556f41090a5a5cc

pomegranited:
Sorry @ChrisChV one last change request: could you bump the [version](https://github.com/open-craft/openedx-learning/blob/chris/system-defined-taxonomies/openedx_learning/__init__.py) to `"0.1.1"`? We'll tag it after merging to produce a new release.

bradenmacdonald (approved):

PR #72, opened 2023-08-11
docs: ADR for pagination and repr of single taxonomy view api [FC-0030]

ChrisChV:
This PR adds an ADR to define the representation of the tags and the pagination in the single taxonomy view api

Reference issue: https://github.com/openedx/modular-learning/issues/75

### DIFF
--- a/docs/decisions/0014-single-taxonomy-view-api.rst
+++ b/docs/decisions/0014-single-taxonomy-view-api.rst
@@ -7,8 +7,8 @@ Context
 
 This view returns tags of a closed taxonomy (for MVP has not been implemented yet
 for open taxonomies). It is necessary to make a decision about what structure the tags are going 
-to have and how the pagination is going to work. It was taken into account that a taxonomy can
-have a large number of tags and are mostly represented as trees.
+to have, how the pagination is going to work and how will the search for tags be implemented.
+It was taken into account that a taxonomy can have a large number of tags and are mostly represented as trees.
 
 **Branch:** Is the representation of a root Tag and all its children up to the leaves.
 
@@ -82,6 +82,17 @@ Children do not affect pagination in any way.
   the branch has hundreds of children, and they would still all be brought.
 
 
+Search tags
+~~~~~~~~~~~~
+
+Support tag search on the backend. Return a subset of matching tags in the format proposed
+in this document.
+
+**Pros**
+
+- It is the most scalable way.
+
+
 Rejected Options
 -----------------
 
@@ -131,3 +142,15 @@ Ex. If the ``page_size`` is 100, when fetching the first root tag, which has 10 
 - All branches are variable in size, therefore a variable number of root tags
   would be returned. This would cause interfaces between taxonomies to be inconsistent
   in the number of root tags shown.
+
+
+Search on frontend
+~~~~~~~~~~~~~~~~~~
+
+We constrain the number of tags allowed in a taxonomy for MVP, so that the API 
+can return all the tags in one page. So we can perform the tag search on the frontend.
+
+**Cons:**
+
+- It is not scalable
+- Sets limits of tags that can be created in the taxonomy
--- a/docs/decisions/0014-single-taxonomy-view-api.rst
+++ b/docs/decisions/0014-single-taxonomy-view-api.rst
@@ -14,10 +14,21 @@ It was taken into account that a taxonomy can have a large number of tags and ar
 
 For the decisions, the following use cases were taken into account:
 
-- List of root tags that can be expanded to show children tags.
-    - This list can be sorted alphabetically: A-Z (default) and Z-A
-- The user can expand all root tags.
-- The user can search for tags.
+- As a taxonomy administrator, I want to see all the tags available for use with a closed taxonomy,
+  so that I can see the taxonomy's structure in the management interface.
+     - As a taxonomy administrator, I want to see the available tags as a lits of root tags
+       that can be expanded to show children tags.
+     - As a taxonomy administrator, I want to sort the list of root tags alphabetically: A-Z (default) and Z-A.
+     - As a taxonomy administrator, I want to expand all root tags to see all children tags.
+     - As a taxonomy administrator, I want to search for tags, so I can find root and children tags more easily.
+- (TODO: discuss with UX) As a course author, when I am editing the tags of a component, I want to see all the tags available
+  from a particular taxonomy that I can use.
+
+Excluded use cases:
+
+- As a content author, when searching/filtering a course/library, I want to see which tags are applied to the content
+  and use them to refine my search. - This is excluded from this API's use case because this is automatically handled
+  by elasticsearch/opensearch.
 
 
 Decision
--- a/docs/decisions/0014-single-taxonomy-view-api.rst
+++ b/docs/decisions/0014-single-taxonomy-view-api.rst
@@ -1,0 +1,126 @@
+13. Single taxonomy view API
+=====================================
+
+
+Context
+--------
+
+This view returns tags of a closed taxonomy (for MVP has not been implemented yet
+for open taxonomies). It is necessary to make a decision about what structure the tags are going 
+to have and how the pagination is going to work. It was taken into account that a taxonomy can
+have a large number of tags and are mostly represented as trees.
+
+**Branch:** Is the representation of a root Tag and all its children up to the leaves.
+
+
+Decision
+---------
+
+
+Tag representation
+~~~~~~~~
+
+Return a list of root tags and within each one have the list of children tags. Each root would have
+its entire branch. The list of root tags will be ordered alphabetically as is each listing at each level of the tree.
+
+{
+    "count": 100,
+    "tags": [
+        {
+            "id": "tag_1",
+            "value": "Tag 1",
+            "taxonomy_id": "1",
+            "sub_tags": [
+                {
+                    "id": "tag_2",
+                    "value": "Tag 2",
+                    "taxonomy_id": "1",
+                    "sub_tags": [
+                        (....)
+                    ]
+                },
+                (....)
+            ]
+        },
+        (....)
+    ]
+}
+
+
+**Pros:**
+
+- The edX's interfaces show the tags in the form of a tree.
+- The frontend needs no further processing as it is in a displayable format.
+- It is kept as a simple implementation.
+
+**Cons:**
+
+- More implementation on the API side.
+
+
+Pagination
+~~~~~~~~~~~
+
+Apply the pagination only in the root tags and bring the entire branch of each root.
+Children do not affect pagination in any way.
+
+**Pros**
+
+- It is the simplest way.
+
+**Cons**
+
+- The children would not have pagination, in the long run there may be cases in which
+  the branch has hundreds of children, and they would still all be brought.
+
+
+Rejected Options
+-----------------
+
+
+Render as a simple list of tags
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Return a simple list of tags, regardless of whether it is root or leaf.
+
+**Pros:**
+
+- It is simple and does not need further implementation and processing in the API.
+
+**Cons:**
+
+- It is more work to re-process all that list in the frontend to know who it is
+whose father.
+- In no edX's interface is it used this way and it would be a very specific use case.
+- Pagination would be more complicated to perform.
+
+
+
+Get the branch in another call
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+Get the root tags in one call and all children tags of a branch in another call.
+This second function is called when the user expands the parent tag.
+
+**Cons:**
+
+- In the UI there is the functionality *Expand all*, another view would have to 
+  be made to handle this functionality in a scalable way.
+- A user could make many calls; every time a parent is opened.
+
+
+
+Add the children to the pagination
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ex. If the ``page_size`` is 100, when fetching the first root tag, which has 10 children tags, 
+11 tags are counted for the total and there would be reamin 89 tags to be obtained.
+
+**Cons:**
+
+- If there is a branch with a number of tags that exceeds ``page_size``, 
+  it would only return that branch.
+- All branches are variable in size, therefore a variable number of root tags
+  would be returned. This would cause interfaces between taxonomies to be inconsistent
+  in the number of root tags shown.
--- a/docs/decisions/0014-single-taxonomy-view-api.rst
+++ b/docs/decisions/0014-single-taxonomy-view-api.rst
@@ -56,7 +56,7 @@ In both cases the results are paginated. In addition to the common pagination me
 - Total number of pages.
 - Total number of root/children tags.
 - Range index of current page, Ex. Page 1: 1-12, Page 2: 13-24.
-- Total number of children of each root tag.
+- Total number of children of each tag.
 
 The pagination of root tags and child tags are independent.
 In order to be able to fulfill the functionality of "Expand-all" in a scalable way,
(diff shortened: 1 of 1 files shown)


openedx-webhooks:

- supporting documentation

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,126 @@
...
+Return a simple list of tags, regardless of whether it is root or leaf.
+
+**Pros:**
+
+- It is simple and does not need further implementation and processing in the API.
+
+**Cons:**
+
+- It is more work to re-process all that list in the frontend to know who it is
+whose father.
```
pomegranited, line 93: formatting nit:
```suggestion
- It is more work to re-process all that list in the frontend to know who it is whose father.
```

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,126 @@
...
+Children do not affect pagination in any way.
+
+**Pros**
+
+- It is the simplest way.
+
+**Cons**
+
+- The children would not have pagination, in the long run there may be cases in which
+  the branch has hundreds of children, and they would still all be brought.
```
pomegranited, line 74: If we need to mitigate this, we could enforce a "maximum number of tags per branch" during tag import.

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,126 @@
...
+
+**Pros:**
+
+- The edX's interfaces show the tags in the form of a tree.
+- The frontend needs no further processing as it is in a displayable format.
+- It is kept as a simple implementation.
+
+**Cons:**
+
+- More implementation on the API side.
```
pomegranited, line 58: Do you think we should revert this decision to better support tree-based queries? https://github.com/open-craft/openedx-learning/blob/chris/single-taxonomy-view-ADR/docs/decisions/0008-tagging-tree-data-arch.rst

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,126 @@
...
+
+**Pros**
+
+- It is the simplest way.
+
+**Cons**
+
+- The children would not have pagination, in the long run there may be cases in which
+  the branch has hundreds of children, and they would still all be brought.
+
```
pomegranited, line 75: I agree with your chosen option @ChrisChV , that makes total sense.

There's two more use cases implied by the UI:

tag_search_sort

**Sort A-Z or Z-A**

This sorts the top-level tags, not the child tags underneath (cf [figma comment](https://www.figma.com/file/DmUC2e7BAoQsGDq4KgApPa?node-id=1237:117997&mode=design#504230063)).

**Search for tags in a taxonomy**

This one's a little more complicated since it could be done a couple of ways.

Options I see:

1. We support tag search on the backend, and return a subset of matching tags in the format proposed here.
  It means a backend API hit every time someone searches for a tag, but will scale.
1. We constrain the number of tags allowed in a taxonomy for MVP, so that the API can return all the tags in one page.
  This allows the frontend to handle tag search all on its own, which is performant, but doesn't scale.

@ChrisChV  and @bradenmacdonald  How do you think we should handle this?

pomegranited: @bradenmacdonald Yeah, I don't think there's an upper bound planned for post-MVP, and there's already [huge taxonomies being discussed](https://openedx.slack.com/archives/C03T0RZEASG/p1692043507665879), as you noted. So I agree: using the backend for search is the safest (and simplest) option.

ChrisChV: Considering [taxonomies like this](https://openedx.slack.com/archives/C03T0RZEASG/p1692043507665879), I also agree with the search option in the backend. But another concern arises, in the [interface to tag an object](https://github.com/openedx/openedx-learning/pull/59#discussion_r1274400990), will all the tags be brought in as initially thought? I think this page should use this same view (with pagination) and the same search

docs/decisions/0014-single-taxonomy-view-api.rst
ChrisChV: I think that a use case has not yet emerged to change that decision. Since the depth limit is still 3. But I would also like to know your thoughts on this

mphilbrick211:
Hi @ChrisChV! Is this pull request ready for review?

docs/decisions/0014-single-taxonomy-view-api.rst
bradenmacdonald: @ChrisChV Given the huge taxonomies shown, I think we'll have to find a different approach for that page. We can't bring in all the tags as we initially thought. It would be too big.

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,156 @@
...
+=====================================
+
+
+Context
+--------
+
+This view returns tags of a closed taxonomy (for MVP has not been implemented yet
+for open taxonomies). It is necessary to make a decision about what structure the tags are going 
+to have, how the pagination is going to work and how will the search for tags be implemented.
+It was taken into account that a taxonomy can have a large number of tags and are mostly represented as trees.
```
bradenmacdonald, line 11: Please add the use cases of this API:
* "As a taxonomy administrator, I want to see all the tags available for use with a closed taxonomy, so that I can see the taxonomy's structure in the management interface."
* As a course author, when I am editing the tags of a component, I want to see all the tags available from a particular taxonomy that I can use. (**maybe**? I think we'll need to add this use case, as we discussed.)

Non-use case:
* As a content author, when searching/filtering a course/library, I want to see which tags are applied to the content and use them to refine my search. - this is _excluded_ from this API's use case because this is automatically handled by elasticsearch/opensearch.

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,156 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+Get the root tags in one call and all children tags of a branch in another call.
+This second function is called when the user expands the parent tag.
+
+**Cons:**
+
+- In the UI there is the functionality *Expand all*, another view would have to 
+  be made to handle this functionality in a scalable way.
```
bradenmacdonald, line 127: I personally like the "Get the branch in another call" approach the most. It's simple: first load only the first page of root tags, and then the user can load additional pages of root tags or expand any root tag and load [the first page of] its children. And so on, recursively.

If the main con is
> In the UI there is the functionality *Expand all*, another view would have to  be made to handle this functionality in a scalable way.
I think we can easily work around that by disabling the "Expand all" option for taxonomies that have > 1,000 tags.

Likewise, for taxonomies with < 1,000 tags we can pre-load all the data from the whole taxonomy.

But don't go changing the ADR just yet - let's ask on Slack for input.

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,156 @@
...
+- A user could make many calls; every time a parent is opened.
```
bradenmacdonald, line 128: That's fine, these calls should be extremely fast and performant to serve.

docs/decisions/0014-single-taxonomy-view-api.rst
bradenmacdonald: We can ask the product folks, but I don't think this restriction is going to work. This whole approach assumes that there are a lot of root tags and very few children per root, but I think the opposite will generally be more common: only a few root tags and potentially thousands of child or grandchild tags. Think of the tree of life used in Biology - it has only three root tags (Bacteria, Achaea, and Eukaryota) but there are over 1 million species in the taxonomy. Of course that's an extreme case, but I think that shape of tree is more common than many root tags and few children. See also LabXchange, or Amazon product categories, or anything like that.

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,156 @@
...
+It was taken into account that a taxonomy can have a large number of tags and are mostly represented as trees.
+
+**Branch:** Is the representation of a root Tag and all its children up to the leaves.
+
+For the decisions, the following use cases were taken into account:
+
+- List of root tags that can be expanded to show children tags.
+    - This list can be sorted alphabetically: A-Z (default) and Z-A
+- The user can expand all root tags.
+- The user can search for tags.
```
bradenmacdonald, line 20: We need to clarify in this ADR how this search is going to work, and how it interacts with pagination.

My suggestion is something like this:
1. On the backend when serving this API, if there is a search term like "cat", filter all the tags to only those that contain the string "cat" as well as their ancestor tags.
2. If the total number that matched is less than ~200, send them all to the frontend in a JSON tree structure.
3. If the total number that matched is more than that, send only [the first page of] root tags to the frontend. Each root tag can be expanded by another API call, and/or additional pages of root tags can be loaded.

But let's discuss this with product and UX team before making too many changes here.

bradenmacdonald (commented):
Hmm, I like most of your ideas but I'm not sure if this approach is going to work. Let's discuss with UX/product and see what they say.

bradenmacdonald:
@mphilbrick211 Is the "FC" / funded contribution label missing from this repo, or has it been removed in general? This is an FC PR.

docs/decisions/0014-single-taxonomy-view-api.rst
bradenmacdonald: I [asked](https://www.figma.com/file/DmUC2e7BAoQsGDq4KgApPa?node-id=1878:29199&mode=design#533948123) on Figma.

mphilbrick211:
Hi @bradenmacdonald - looks like the label isn't in this repo.

For FC projects, would you mind please putting the FC-XXXX ID number in the PR title? That will also help me know which project it is.

bradenmacdonald:
@mphilbrick211 Yep, we try to do that generally, sorry we missed it here. I've added it now and created the FC label too.

docs/decisions/0014-single-taxonomy-view-api.rst
ChrisChV: Oh Yes, you are right, it makes sense. I think we can mitigate the "_Cons_" of "Get the branch in another call"

docs/decisions/0014-single-taxonomy-view-api.rst
ChrisChV: I was thinking of using the same view for both screens, since the screen to tag an object must also be paginated, I'm going to [ask](https://www.figma.com/file/DmUC2e7BAoQsGDq4KgApPa?node-id=1515:84066&mode=design#535383862) UX about this

docs/decisions/0014-single-taxonomy-view-api.rst
ChrisChV: @bradenmacdonald I have updated the document with what was discussed. 

I've been thinking, if we are going to return complete taxonomies that have less than 1000 tags, we can do a search in the frontend with these taxonomies. I have not added it to the document, because I think that many conditionals are already being added and it will complicate the implementation. What do you think?

bradenmacdonald: Yes, I think so too. Although I also don't want to complicate things too much by having two different implementations of search, based on the taxonomy size. Maybe there's some way we can make the "< 1,000 so load all and search via frontend" just a subset / special case of the general version. Something like "if all tags are loaded into the frontend, do a local filter, else do a server-side search and refresh the tags shown" would work in both cases.

bradenmacdonald:
FYI @ormsbee , please review this ADR for pagination of tags if you are interested.

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,185 @@
...
+- Paging both root tags and children mitigates the huge number of tags that can be.
+
+Search tags
+~~~~~~~~~~~~
+
+Support tag search on the backend. Return a subset of matching tags.
+We will use the same views to perform a search with the same logic:
+
+- **get_root_tags(search: str)**
+- **get_children_tags(search: str)**
```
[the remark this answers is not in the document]
ChrisChV: Yes, I like it :+1:

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,185 @@
...
+
+Context
+--------
+
+This view returns tags of a closed taxonomy (for MVP has not been implemented yet
+for open taxonomies). It is necessary to make a decision about what structure the tags are going 
+to have, how the pagination is going to work and how will the search for tags be implemented.
+It was taken into account that taxonomies commonly have the following characteristics:
+
+- It have few root tags.
```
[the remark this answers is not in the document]
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/72/commits/10a02a06febd7840f1ee3cce68ef1b32be5c7ebc

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,185 @@
...
+- It have a very large number of children for each tag.
+- It is mostly represented as trees on frontend.
```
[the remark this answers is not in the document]
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/72/commits/10a02a06febd7840f1ee3cce68ef1b32be5c7ebc

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,185 @@
...
+- Create a ``TAGS_THRESHOLD`` (default: 1000).
+- If ``taxonomy.tags.count < TAGS_THRESHOLD``, then ``get_root_tags`` will return all tags on the taxonomy,
+  roots and children.
+- Otherwise, ``get_root_tags`` will only return root tags, and it will be necessary
+  to use ``get_children_tags`` to return children. Also the "Expand-all" functionality will be disabled. 
+
+**Pros**
+
+- It is the simplest way.
+- Paging both root tags and children mitigates the huge number of tags that can be.
```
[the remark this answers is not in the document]
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/72/commits/10a02a06febd7840f1ee3cce68ef1b32be5c7ebc

docs/decisions/0014-single-taxonomy-view-api.rst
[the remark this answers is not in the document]
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/72/commits/10a02a06febd7840f1ee3cce68ef1b32be5c7ebc

docs/decisions/0014-single-taxonomy-view-api.rst
```diff
@@ -0,0 +1,185 @@
...
+**Otherwise**::
+
+  {
+    "count": 100,
+    "tags": [
+        {
+            "id": "tag_1",
+            "value": "Tag 1",
+            "taxonomy_id": "1",
+            "sub_tags_link": "http//api-call-to-get-children.com"
```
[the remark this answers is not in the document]
ChrisChV: Thanks!

ChrisChV:
Hi @bradenmacdonald, I think we can merge this ADR

openedx-webhooks:
@ChrisChV 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/72&entry.1484655318=https://github.com/ChrisChV&entry.752974735=2023-08-11+17:56&entry.1917517419=2023-09-12+20:39&entry.2133058324=Yes) so we can improve your experience in the future.

PR #73, opened 2023-08-16
Enforce quality checks and type hints, improve quality and typing

bradenmacdonald:
As it turns out, the CI checks for this repo were not running pylint or mypy, and there were a ton of lint errors and typing bugs.

This repo's packages also didn't have any `py.typed` files, so whatever type hints they declared were being ignored by other code like edx-platform code that uses this.

This PR:
* configures CI to run quality checks, including using `mypy` to do static type checking of the code.
* Adds more type hints, both to make code visible to `mypy` and to improve its quality
* fixes all the lint issues found
* fixes all the type hint issues found

A few of these issues automatically identified are clearly bugs, which are now fixed :)

Private-ref: FAL-3476


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_learning/core/publishing/api.py
```diff
@@ -131,7 +135,7 @@ def publish_from_drafts(
             learning_package_id=learning_package_id,
             message=message,
             published_at=published_at,
-            published_by=published_by,
+            published_by_id=published_by,
```
bradenmacdonald, line 138: This was a bug: the function accepted an `int` but assigned it to a `User` model attribute.

openedx_learning/core/publishing/api.py
```diff
@@ -93,7 +97,7 @@ def learning_package_exists(key: str) -> bool:
     """
     Check whether a LearningPackage with a particular key exists.
     """
-    LearningPackage.objects.filter(key=key).exists()
+    return LearningPackage.objects.filter(key=key).exists()
```
bradenmacdonald, line 100: This was a bug - missing return statement.

openedx_learning/core/components/admin.py
```diff
@@ -120,29 +125,8 @@ def get_queryset(self, request):
         )
 
 
-def is_displayable_text(mime_type):
```
bradenmacdonald, line 123: This function was no longer used, after a prior refactor.

openedx_learning/core/contents/api.py
```diff
@@ -31,8 +40,11 @@ def create_raw_content(
     return raw_content
 
 
-def create_text_from_raw_content(raw_content, encoding="utf-8-sig"):
-    text = codecs.decode(raw_content.file.open().read(), "utf-8-sig")
+def create_text_from_raw_content(raw_content: RawContent, encoding="utf-8-sig") -> TextContent:
+    """
+    Create a new TextContent instance for the given RawContent.
+    """
+    text = codecs.decode(raw_content.file.open().read(), encoding)
```
bradenmacdonald, line 47: This was a bug (I assume): the API accepted an `encoding` parameter but ignored it, and always used `"utf-8-sig"`

openedx_learning/core/contents/api.py
```diff
@@ -41,37 +53,49 @@ def create_text_from_raw_content(raw_content, encoding="utf-8-sig"):
...
+    or create it if it doesn't exist.
+    """
     with atomic():
         raw_content, rc_created = get_or_create_raw_content(
-            learning_package_id, data_bytes, mime_type, created, hash_digest=None
+            learning_package_id, data_bytes, mime_type, created, hash_digest
         )
         if rc_created or not hasattr(raw_content, "text_content"):
-            text = codecs.decode(data_bytes, "utf-8-sig")
+            text = codecs.decode(data_bytes, encoding)
```
bradenmacdonald, line 98: Same here: `encoding` parameter was silently ignored.

openedx_learning/lib/fields.py
```diff
@@ -19,7 +19,7 @@
 from .validators import validate_utc_datetime
 
 
-def create_hash_digest(data_bytes):
+def create_hash_digest(data_bytes: bytes) -> str:
```
bradenmacdonald, line 22: With a type annotation, this argument could be renamed from `data_bytes` to `data: bytes` which is much more clear (`data_bytes` could be a size in bytes, for example). But the name `data_bytes` is already used by `create_raw_content` in the public API so I'm not sure if it's OK to change.

openedx_tagging/core/tagging/api.py
```diff
@@ -20,19 +21,19 @@
...
-    taxonomy_class: Type = None,
+    taxonomy_class: type[Taxonomy] | None = None,
 ) -> Taxonomy:
     """
     Creates, saves, and returns a new Taxonomy with the given attributes.
     """
     taxonomy = Taxonomy(
         name=name,
-        description=description,
+        description=description or "",
```
bradenmacdonald, line 36: Per Django convention, `NULL`/`None` should not be used for text fields.

openedx_tagging/core/tagging/import_export/exceptions.py
```diff
@@ -63,8 +72,8 @@ class InvalidFormat(TagParserError):
     Exception used when there is an error with the format
     """
 
-    def __init__(self, tag: dict, format: str, message: str, **kargs):
-        self.tag = tag
+    def __init__(self, tag: dict | None, format: str, message: str, **kargs):
+        super().__init__(tag)
```
bradenmacdonald, line 76: These exception classes are quite messy and could use a refactor. Once you add type hints, the problems become more obvious. In particular, none of them were calling `super().__init__()`, and the purpose of the `tag` field is unclear. Some have hard-coded messages and some let you override the `message`. For now I just made the minimal changes to get type checks passing.

openedx_tagging/core/tagging/import_export/parsers.py
```diff
@@ -201,24 +201,24 @@ class JSONParser(Parser):
...
+    def _load_data(cls, file: BytesIO) -> tuple[list[dict], list[TagParserError]]:
         """
         Read a .json file and validates the root structure of the json
         """
         file.seek(0)
         try:
             tags_data = json.load(file)
         except json.JSONDecodeError as error:
-            return None, [
+            return [], [
```
bradenmacdonald, line 217: This was returning `(None, list)` in a clear violation of the declared return type `(list, list)`

openedx_tagging/core/tagging/import_export/parsers.py
```diff
@@ -255,25 +255,25 @@ class CSVParser(Parser):
...
         """
         file.seek(0)
         text_tags = TextIOWrapper(file, encoding="utf-8")
         csv_reader = csv.DictReader(text_tags)
         header_fields = csv_reader.fieldnames
-        errors = cls._verify_header(header_fields)
+        errors = cls._verify_header(list(header_fields or []))
         if errors:
-            return None, errors
+            return [], errors
```
bradenmacdonald, line 272: Same here, invalid return of `None`

openedx_tagging/core/tagging/import_export/api.py
```diff
@@ -124,14 +125,18 @@ def get_last_import_status(taxonomy: Taxonomy) -> TagImportTaskState:
     Get status of the last import task of the given taxonomy
     """
     task = _get_last_import_task(taxonomy)
-    return task.status
+    if task is None:
+        raise ValueError("No import task was created yet.")
+    return TagImportTaskState(task.status)
```
bradenmacdonald, line 130: This was a bug: the typing declared that it returned enum instances, but it was returning the enum _values_ which are just plain strings. Also it wasn't handling the case where there was no "last import task".

openedx_tagging/core/tagging/models/base.py
```diff
@@ -119,7 +121,6 @@ class Taxonomy(models.Model):
         ),
     )
     description = MultiCollationTextField(
-        null=True,
```
bradenmacdonald, line 122: Django convention is not not use NULL for text fields, unless there's a very good reason, which doesn't seem to be the case here.

tests/openedx_tagging/core/tagging/test_system_defined_models.py
```diff
@@ -98,31 +94,35 @@ def test_implementation_error(self, taxonomy_cls, expected_exception):
...
-
-        assert self.user_taxonomy.validate_object_tag(
-            object_tag=object_tag,
-            check_object=False,
-            check_taxonomy=False,
-            check_tag=True,
-        ) == expected
+    # FIXME: something is wrong with this test case. It's setting the string
+    # "tag_id" as the primary key (integer) of the Tag instance, and it mentions
+    # "parent validation" but there is nothing to do with parents here.
```
bradenmacdonald, line 99: I'm not sure what was going on with this test case. It was failing and I don't know what it's even trying to do, so for now I commented it out.

tests/openedx_tagging/core/tagging/test_system_defined_models.py
```diff
@@ -205,29 +205,33 @@ class TestLanguageTaxonomy(TestTagTaxonomyMixin, TestCase):
...
-            id=tag_id,
-            taxonomy=self.language_taxonomy,
-            value="_val",
-            external_id=lang,
-        )
-        object_tag = ObjectTag(
-            object_id="id",
-            tag=tag,
-        )
-        self.language_taxonomy.validate_object_tag(
```
bradenmacdonald, line 225: This function just returns a `bool` so this test was having no effect. I commented it out because again I'm not sure what it's supposed to be doing.

openedx_tagging/core/tagging/models/base.py
```diff
@@ -344,11 +345,11 @@ def _check_taxonomy(
         Subclasses can override this method to perform their own taxonomy validation checks.
         """
         # Must be linked to this taxonomy
-        return object_tag.taxonomy_id and object_tag.taxonomy_id == self.id
+        return bool(object_tag.taxonomy_id) and object_tag.taxonomy_id == self.id
```
rpenido, line 349: I'm unsure if it applies here, but in the edgy case where `taxonomy_id=0`, `bool(object_tag.taxonomy_id)` will mistakenly return `False`.
```suggestion
        return object_tag.taxonomy is not None and object_tag.taxonomy_id == self.id
```

openedx_learning/core/components/admin.py
```diff
@@ -1,34 +1,39 @@
...
 from .models import (
     Component,
     ComponentVersion,
     ComponentVersionRawContent,
 )
-from openedx_learning.lib.admin_utils import ReadOnlyModelAdmin
 
 
 class ComponentVersionInline(admin.TabularInline):
+    """Inline admin view of ComponentVersion from the Component Admin"""
```
rpenido, line 16: Nit: Our [python](https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#docstrings) guidelines state that we should:
> "Always format docstrings using the multi-line convention, even if there’s only one line of content".

I saw that you're fixing a lot of functions without docstring here. Is it a good time to fix this (at least in the changed ones)?

openedx_learning/lib/collations.py
```diff
@@ -77,7 +77,7 @@ def db_collation(self, value):
 
         This can be removed when we move to Django 4.2.
         """
-        pass
+        pass  # pylint: disable=unnecessary-pass
```
rpenido, line 80:
```suggestion
```

If we have a docstring I think is safe to remove the `pass`, right?

openedx_learning/lib/fields.py
```diff
@@ -172,7 +172,7 @@ class MultiCollationCharField(MultiCollationMixin, models.CharField):
     PostgreSQL. Even MariaDB is starting to diverge from MySQL in terms of what
     collations are supported.
     """
-    pass
+    pass  # pylint: disable=unnecessary-pass
```
rpenido, line 175:
```suggestion
```
Ditto

openedx_learning/lib/fields.py
```diff
@@ -183,4 +183,4 @@ class MultiCollationTextField(MultiCollationMixin, models.TextField):
     forces the compatible charset to be set in MySQL, and that's the part that
     matters for our purposes. So for example, if you set 
     """
-    pass
+    pass  # pylint: disable=unnecessary-pass
```
rpenido, line 186:
```suggestion
```
Ditto

openedx_learning/rest_api/v1/components.py
```diff
@@ -7,8 +7,11 @@
 
 
 class ComponentViewSet(viewsets.ViewSet):
+    """
+    Example endpoints for a Components REST API. Not implemented.
+    """
     def list(self, request):
-        items = Component.objects.all()
+        items = Component.objects.all()  # pylint: disable=unused-variable
```
rpenido, line 14: Nit: remove pylint comment
```suggestion
        _items = Component.objects.all()
```

rpenido (commented):
Hi @bradenmacdonald! Awesome work here.

I just pointed out some nits, but feel free to push back if you want to!

We are working a lot with this repository, so we should merge this ASAP to avoid a rebase hell.

PS: Running `make quality` I got a warning on the console:
```
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.
```
Changing the `pylintrc` as suggested in the warning removed it.
```
[EXCEPTIONS]
overgeneral-exceptions = builtins.Exception
```

tests/openedx_tagging/core/tagging/test_system_defined_models.py
rpenido: @ChrisChV Could you please check this when you have time?

tests/openedx_tagging/core/tagging/test_system_defined_models.py
rpenido: @ChrisChV Same here.

openedx_learning/lib/collations.py
bradenmacdonald: > If we have a docstring I think is safe to remove the pass, right?

Yes, totally safe. I just left it in because it's a more explicit signal that the class/function is supposed to be blank. But you're right, maybe it's cleaner to just remove it.

openedx_tagging/core/tagging/models/system_defined.py
```diff
@@ -1,4 +1,6 @@
 """ Tagging app system-defined taxonomies data models """
+from __future__ import annotations
+
 import logging
 from typing import Any, List, Type, Union
```
rpenido, line 5:
```suggestion
from typing import Any, Type, Union
```
I think we have an unused import here.

openedx_learning/core/components/admin.py
bradenmacdonald: Oh, I didn't know that was the official guideline. That's actually my personal preference too; I just used the one line form because it was already in use in this repo. I agree it makes sense to fix it now.

openedx_tagging/core/tagging/models/base.py
bradenmacdonald: I don't think we really need to worry about id being zero, but it's better to be safe than sorry, so I've made that change :)

bradenmacdonald:
@rpenido Thanks for the great review! (and fast!) I agree that we should merge this ASAP because of the large number of possible conflicts.

> PS: Running make quality I got a warning on the console... Changing the pylintrc as suggested in the warning removed it.

The `pylintrc` is managed by [edx-lint](https://github.com/openedx/edx-lint), so they'll probably fix it "upstream" at some point. We don't need to fix it separately in this repo.

> Is there a chance that we break edx-platform (if the version is not pinned) because we forced the types here, and there is a wrong call on edx-platform side (or another package that imports this package)?

Nope. Python ignores the types at runtime, and on the edx-platform side they're not even being checked yet. Actually once this PR is merged, I'm going to modify edx-platform so that they are checked.

> Is there something that I can check on the `django-rules`PR? I'm not sure how to reproduce what are you experiencing.

Sorry, I didn't mention that. You need to edit `openedx_tagging/core/tagging/rules.py` to remove the `# type: ignore` from the line where we `import rules`, and then if you run `mypy` you'll see the error I mentioned.

bradenmacdonald:
@ormsbee Since Jill is on leave, would you be able to approve merging this PR?

openedx_learning/core/contents/api.py
ormsbee: Yup, ty!

openedx_learning/lib/fields.py
ormsbee: The only place it could be used is edx-platform, so I think it's okay to change if it's not already referenced there. I also don't think it's that important to do so though.

ormsbee (approved):
Looks good to me. I realize there are some follow-up items from this (like the test), but it's probably better to merge this ASAP given it's size and breadth.

Thanks @bradenmacdonald. This is one of those things that I really wanted to see happen but didn't get around to. It warms my heart to see this sort of thing revealing bugs.

openedx_learning/lib/fields.py
bradenmacdonald: Yeah honestly this is a very minor thing :p

openedx-webhooks:
@bradenmacdonald 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/73&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2023-08-16+21:04&entry.1917517419=2023-08-24+16:48&entry.2133058324=Yes) so we can improve your experience in the future.

tests/openedx_tagging/core/tagging/test_system_defined_models.py
ChrisChV: Ack!

tests/openedx_tagging/core/tagging/test_system_defined_models.py
ChrisChV: Ack!

PR #81, opened 2023-09-19
Enforce limit on number of tags per object

rpenido:
## Description

This PR enforces a 100 hardcoded limit on the number of ObjectTags applied on any object. It also removes the pagination of the tagging REST API for listing and updating object tags.

## Supporting Information

* Closes https://github.com/openedx/modular-learning/issues/99

## Testing instructions
* Please ensure that the tests cover the expected behavior of the view as described in the related issue.

## After merge
- [ ] Create a Tag with the version `v0.1.7`
____
Private-ref: [FAL-3501](https://tasks.opencraft.com/browse/FAL-3501)

### DIFF
--- a/openedx_tagging/core/tagging/models/base.py
+++ b/openedx_tagging/core/tagging/models/base.py
@@ -446,24 +446,25 @@ class Taxonomy(models.Model):
                 -1,
             )
 
-        def _check_current_tag_count() -> None:
+        def _check_new_tag_count(new_tag_count: int) -> None:
             """
-            Checks if the current count of tags for the object is less than 100
+            Checks if the new count of tags for the object is equal or less than 100
             """
             # Exclude self.id to avoid counting the tags that are going to be updated
             current_count = ObjectTag.objects.filter(object_id=object_id).exclude(taxonomy_id=self.id).count()
 
-            if current_count >= 100:
+            if current_count + new_tag_count > 100:
                 raise ValueError(
-                    _(f"Object ({object_id}) already have 100 or more tags.")
+                    _(f"Cannot add more than 100 tags to ({object_id}).")
                 )
 
-        _check_current_tag_count()
 
         if not isinstance(tags, list):
             raise ValueError(_(f"Tags must be a list, not {type(tags).__name__}."))
 
         tags = list(dict.fromkeys(tags))  # Remove duplicates preserving order
+
+        _check_new_tag_count(len(tags))
 
         if not self.allow_multiple and len(tags) > 1:
             raise ValueError(_(f"Taxonomy ({self.id}) only allows one tag per object."))
--- a/openedx_tagging/core/tagging/models/base.py
+++ b/openedx_tagging/core/tagging/models/base.py
@@ -458,7 +458,6 @@ class Taxonomy(models.Model):
                     _(f"Cannot add more than 100 tags to ({object_id}).")
                 )
 
-
         if not isinstance(tags, list):
             raise ValueError(_(f"Tags must be a list, not {type(tags).__name__}."))
 
--- a/openedx_tagging/core/tagging/models/base.py
+++ b/openedx_tagging/core/tagging/models/base.py
@@ -404,6 +404,19 @@ class Taxonomy(models.Model):
                 -1,
             )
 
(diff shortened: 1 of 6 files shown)


openedx-webhooks:
Thanks for the pull request, @rpenido!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

rpenido:
ToDo: bump version

openedx_tagging/core/tagging/models/base.py
```diff
@@ -404,6 +404,20 @@ def _find_object_tag_index(tag_ref, object_tags) -> int:
...
+        def _check_current_tag_count() -> None:
+            """
+            Checks if the current count of tags for the object is less than 100
+            """
+            # Exclude self.id to avoid counting the tags that are going to be updated
+            current_count = ObjectTag.objects.filter(object_id=object_id).exclude(taxonomy_id=self.id).count()
+
+            if current_count >= 100:
+                raise ValueError(
+                    _(f"Object ({object_id}) already have 100 or more tags.")
```
yusuf-musleh, line 416: nit: just a small spelling fix, "already have" -> "already has"

openedx_tagging/core/tagging/models/base.py
```diff
@@ -404,6 +404,20 @@ def _find_object_tag_index(tag_ref, object_tags) -> int:
...
+                )
+
+        _check_current_tag_count()
```
yusuf-musleh, line 419: Shouldn't this check also include the count of the new tags that will will be potentially be added, since this check is happening at the beginning of the function call? Eg: What happens when the `current_count` is say `99` but there are `2` _new_ tags in the `tags` list leading to `101`, would this check catch it?

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -782,3 +743,21 @@ def test_tag_object_without_permission(self, user_attr, expected_status):
 
         response = self.client.put(url, {"tags": ["Tag 1"]}, format="json")
         assert response.status_code == expected_status
+        assert not status.is_success(expected_status)  # No success cases here
+
+    def test_tag_object_count_limit(self):
```
yusuf-musleh, line 748: Would super helpful if we could also add a test case that covers the scenario described above.

yusuf-musleh (changes requested):
@rpenido Great work! I just had a few comments on a test case we are missing if you could take a look.

openedx_tagging/core/tagging/models/base.py
rpenido: You're right @yusuf-musleh! I totally missed the allow_multiple cases.
Thank you for pointing out!

openedx_tagging/core/tagging/models/base.py
rpenido: Changed here: https://github.com/openedx/openedx-learning/blob/432e4c22eb18a05eeafb5070a14666680f4c1484/openedx_tagging/core/tagging/models/base.py#L458

We are talking about the future state, not the current.

openedx_tagging/core/tagging/models/base.py
rpenido: Done 432e4c22eb18a05eeafb5070a14666680f4c1484

tests/openedx_tagging/core/tagging/test_views.py
rpenido: Agreed! Thank you!

Done here: 432e4c22eb18a05eeafb5070a14666680f4c1484

yusuf-musleh (approved):
:+1: Thanks for making the changes! Looks good to me, I believe @bradenmacdonald would need to do the final approval/merge for it.

- [X] I tested this: I read through the tests to make sure it covers the cases
- [X] I read through the code
- [X] ~I checked for accessibility issues~
- [X] Includes documentation
- [X] ~I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~

bradenmacdonald (approved):
Nice work! And I love the simplification of not needing pagination due to the new rule.

PR #87, opened 2023-09-29
Simplify Tag Models [WIP] [FC-0030]

bradenmacdonald:
This is my work on https://github.com/openedx/modular-learning/issues/85 , "clean up tagging models".

I have a few other cleanups in mind but I'm trying to keep this PR focused on one idea: Taxonomies are not aware of Objects. The Taxonomy cannot specify what types of objects it applies to, nor which ObjectTag subclass it works with. Instead that can be specified at a higher level in the API.

So the changes include:
* Removed `Taxonomy.object_tag_class` - taxonomies no longer specify what type of objects they can tag
* Removed `ModelObjectTag` - ObjectTag should only be subclassed to tag different types of objects, not to use different types of tags. It was confusing that we used `ContentObjectTag` when we want to tag content objects, but `ModelObjectTag` when we want to use Tags that are defined from another model.
* Moved `tag_object` and `autocomplete_tags` from `Taxonomy` to be python functions in the API. These were not actually being subclasses anyways so don't need to be instance methods.
* Removed `tag_ref` - **all tags are now set by value**. This makes things simpler and more consistent. (Values may change, it's true, but at any given point in time a value is a unique ID within a given taxonomy, and values are what users see. Tag IDs are still used internally and exposed via the API, but not used when setting tags. This lets us remove `tag_ref` and logic that was in `ObjectTag` is moved to more appropriate places.)

## TODO

This is still a work in progress. I have all the tests in `test_system_defined_models` working, but I have to update the other test cases and the corresponding edx-platform code. You can let me know your thoughts on this refactor now though!

Private-ref: [FAL-3477](https://tasks.opencraft.com/browse/FAL-3477)

### DIFF
--- a/openedx_tagging/core/tagging/models/base.py
+++ b/openedx_tagging/core/tagging/models/base.py
@@ -137,7 +137,7 @@ class Taxonomy(models.Model):
         ),
     )
     allow_multiple = models.BooleanField(
-        default=False,
+        default=False,  # TODO: This should be true, or perhaps remove this property altogether
         help_text=_(
             "Indicates that multiple tags from this taxonomy may be added to an object."
         ),
@@ -243,6 +243,13 @@ class Taxonomy(models.Model):
             )
 
         return self
+    
+    def check_casted(self):
+        """
+        Double-check that this taxonomy has been cast() to a subclass if needed.
+        """
+        if self.cast() is not self:
+            raise TypeError("Taxonomy was used incorrectly - without .cast()")
 
     def copy(self, taxonomy: Taxonomy) -> Taxonomy:
         """
@@ -341,75 +348,54 @@ class Taxonomy(models.Model):
 
         return tag_set.order_by("value", "id")
 
-    # TODO: replace this with just validate_value() and value_for_external_id()
-    def validate_object_tag(
-        self,
-        object_tag: "ObjectTag",
-        check_taxonomy=True,
-        check_tag=True,
-        check_object=True,
-    ) -> bool:
+    def validate_value(self, value: str) -> bool:
         """
-        Returns True if the given object tag is valid for the current Taxonomy.
-
-        Subclasses should override the internal _validate* methods to perform their own validation checks, e.g. against
-        dynamically generated tag lists.
-
-        If `check_taxonomy` is False, then we skip validating the object tag's taxonomy reference.
-        If `check_tag` is False, then we skip validating the object tag's tag reference.
-        If `check_object` is False, then we skip validating the object ID/type.
+        Check if 'value' is part of this Taxonomy.
+        A 'Tag' object may not exist for the value (e.g. if this is a free text
+        taxonomy, then any value is allowed but no Tags are created; if this is
+        a user taxonomy, Tag entries may only get created as needed.), but if
+        this returns True then the value conceptually exists in this taxonomy
+        and can be used to tag objects.
         """
-        if check_taxonomy and not self._check_taxonomy(object_tag):
-            return False
-
-        if check_tag and not self._check_tag(object_tag):
-            return False
-
-        if check_object and not self._check_object(object_tag):
-            return False
-
-        return True
-
-    def _check_taxonomy(
-        self,
-        object_tag: ObjectTag,
-    ) -> bool:
-        """
-        Returns True if the given object tag is valid for the current Taxonomy.
-
-        Subclasses can override this method to perform their own taxonomy validation checks.
-        """
-        # Must be linked to this taxonomy
-        return (
-            object_tag.taxonomy_id is not None
-        ) and object_tag.taxonomy_id == self.id
-
-    def _check_tag(
-        self,
-        object_tag: ObjectTag,
-    ) -> bool:
-        """
-        Returns True if the given object tag's value is valid for the current Taxonomy.
-
-        Subclasses can override this method to perform their own taxonomy validation checks.
-        """
-        # Open taxonomies only need a value.
+        self.check_casted()
         if self.allow_free_text:
-            return bool(object_tag.value)
+            return True
+        return self.tag_set.filter(value__iexact=value).exists()
 
-        # Closed taxonomies need an associated tag in this taxonomy
-        return (object_tag.tag is not None) and object_tag.tag.taxonomy_id == self.id
-
-    def _check_object(
-        self,
-        object_tag: ObjectTag,
-    ) -> bool:
+    def tag_for_value(self, value: str) -> Tag:
         """
-        Returns True if the given object tag's object is valid for the current Taxonomy.
+        Get the Tag object for the given value.
+        Some Taxonomies may auto-create the Tag at this point, e.g. a User
+        Taxonomy will create User Tags "just in time".
 
-        Subclasses can override this method to perform their own taxonomy validation checks.
+        Will raise Tag.DoesNotExist if the value is not valid for this taxonomy.
         """
-        return bool(object_tag.object_id)
+        self.check_casted()
+        if self.allow_free_text:
+            raise ValueError("tag_for_value() doesn't work for free text taxonomies. They don't use Tag instances.")
+        return self.tag_set.get(value__iexact=value)
+
+    def validate_external_id(self, external_id: str) -> bool:
+        """
+        Check if 'external_id' is part of this Taxonomy.
+        """
+        self.check_casted()
+        if self.allow_free_text:
+            return False  # Free text taxonomies don't use 'external_id' on their tags
+        return self.tag_set.filter(external_id__iexact=external_id).exists()
+
+    def tag_for_external_id(self, external_id: str) -> Tag:
+        """
+        Get the Tag object for the given external_id.
+        Some Taxonomies may auto-create the Tag at this point, e.g. a User
+        Taxonomy will create User Tags "just in time".
+
+        Will raise Tag.DoesNotExist if the tag is not valid for this taxonomy.
+        """
+        self.check_casted()
+        if self.allow_free_text:
+            raise ValueError("tag_for_external_id() doesn't work for free text taxonomies.")
+        return self.tag_set.get(external_id__iexact=external_id)
 
 
 class ObjectTag(models.Model):
@@ -527,42 +513,18 @@ class ObjectTag(models.Model):
         """
         self._value = value
 
-    @property
-    def tag_ref(self) -> str:
-        """
-        Returns this tag's reference string.
-
-        If tag is set, then returns its id.
-        Otherwise, returns the cached _value field.
-        """
-        return self.tag.id if self.tag else self._value
-
-    # TODO: remove tag_ref, always reference tags by 'value'
-    @tag_ref.setter
-    def tag_ref(self, tag_ref: str):
-        """
-        Sets the ObjectTag's Tag and/or value, depending on whether a valid Tag is found.
-
-        Subclasses may override this method to dynamically create Tags.
-        """
-        self.value = tag_ref
-
-        if self.taxonomy:
-            try:
-                self.tag = self.taxonomy.tag_set.get(pk=tag_ref)
-                self.value = self.tag.value
-            except (ValueError, Tag.DoesNotExist):
-                # This might be ok, e.g. if our taxonomy.allow_free_text, so we just pass through here.
-                # We rely on the caller to validate before saving.
-                pass
-
     def is_valid(self) -> bool:
         """
         Returns True if this ObjectTag represents a valid taxonomy tag.
 
         A valid ObjectTag must be linked to a Taxonomy, and be a valid tag in that taxonomy.
         """
-        return self.taxonomy.validate_object_tag(self) if self.taxonomy else False
+        if self.tag:
+            return True  # If the Tag still exists, this is a valid value in the taxonomy.
+        elif self.taxonomy:
+            # If this is not a free text value and it's no longer linked to a Tag, it's invalid.
+            # Running self.resync() _might_ fix this though.
+            return self.taxonomy.allow_free_text
 
     def get_lineage(self) -> Lineage:
         """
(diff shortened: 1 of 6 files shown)


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_tagging/core/tagging/models/base.py
```diff
@@ -137,7 +138,7 @@ class Taxonomy(models.Model):
         ),
     )
     allow_multiple = models.BooleanField(
-        default=False,
+        default=False,  # TODO: This should be true, or perhaps remove this property altogether
```
bradenmacdonald, line 141: @ChrisChV @pomegranited Do you know if we actually need this `allow_multiple` flag? I don't see any requirement for it in the UX mockups etc.

Same question about `required`

openedx_tagging/core/tagging/models/base.py
```diff
@@ -595,23 +436,22 @@ class ObjectTag(models.Model):
...
+        null=True,  # NULL in the case of free text taxonomies or when the Tag gets deleted.
+        blank=True,
         default=None,
         on_delete=models.SET_NULL,
         help_text=_(
             "Tag associated with this object tag. Provides the tag's 'value' if set."
         ),
     )
     _name = case_insensitive_char_field(
-        null=False,
```
bradenmacdonald, line 606: null=False is the default for char fields, so no need to specify it here.

openedx_tagging/core/tagging/models/base.py
pomegranited: I think we need `allow_multiple`, but `required` can be removed.
E.g. we only want one Language tag per content item, but we'll want to allow multiple LightCast tags to be added.

ChrisChV: I agree with @pomegranited :+1:

tests/openedx_tagging/core/tagging/test_system_defined_models.py
```diff
@@ -82,122 +78,171 @@ class TestModelSystemDefinedTaxonomy(TestTagTaxonomyMixin, TestCase):
...
-
-    # FIXME: something is wrong with this test case. It's setting the string
-    # "tag_id" as the primary key (integer) of the Tag instance, and it mentions
-    # "parent validation" but there is nothing to do with parents here.
-    #
-    # @ddt.data(
-    #     ("1", "tag_id", True),  # Valid
-    #     ("0", "tag_id", False),  # Invalid user
-    #     ("test_id", "tag_id", False),  # Invalid user id
-    #     ("1", None, False),  # Testing parent validations
```
ChrisChV, line 107: @bradenmacdonald  This case refers more to call the [super() functions](https://github.com/openedx/openedx-learning/pull/87/files#diff-8640a8393169a45e9926177fd6f2e8c7b0ad8439c96554334b95bbe912f52203L304) to call the parent class validations. Since that logic has been removed, I think it is okay to delete this test

tests/openedx_tagging/core/tagging/test_system_defined_models.py
```diff
@@ -207,37 +252,25 @@ class TestLanguageTaxonomy(TestTagTaxonomyMixin, TestCase):
     Test for Language taxonomy
     """
 
-    # FIXME: something is wrong with this test case. It's setting the string
```
ChrisChV, line 210: @bradenmacdonald  Same of [this comment](https://github.com/openedx/openedx-learning/pull/87/files#r1343052892)

tests/openedx_tagging/core/tagging/test_system_defined_models.py
bradenmacdonald: OK. Yeah for now I deleted it but put in some other new related tests.

ChrisChV (approved):
@bradenmacdonald Looks good to me :+1: I left some comments about the [commented tests](https://github.com/openedx/openedx-learning/pull/87#discussion_r1343052892)

- [X] I tested this runing the test.
- [X] I read through the code.
- [X] Includes documentation

openedx_tagging/core/tagging/models/base.py
```diff
@@ -625,7 +465,19 @@ class Meta:
...
+            ("object_id", "taxonomy", "tag_id"),
+            ("object_id", "taxonomy", "_value"),
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if not self.pk:  # This is a new instance:
+            # Set _name and _value automatically on creation, if they weren't set:
+            if self.taxonomy and not self._name:
+                self._name = self.taxonomy.name
```
pomegranited, line 478: nit (for symmetry with the stanza below, and to avoid fetching the taxonomy unless we need to):
```suggestion
            if not self._name and self.taxonomy:
                self._name = self.taxonomy.name
```

openedx_tagging/core/tagging/models/base.py
```diff
@@ -731,34 +567,8 @@ def resync(self) -> bool:
...
-                    self.tag = tag
-                    if taxonomy.validate_object_tag(self):
-                        changed = True
-                        break
-                    # If not, undo those changes and try the next one
-                    else:
-                        self.taxonomy = None
-                        self.tag = None
+        # We used to have code here that would try to find a new taxonomy if the current taxonomy has been deleted.
+        # But for now that's removed, as it risks things like linking a tag to the wrong org's taxonomy.
```
pomegranited, line 571: 💯

openedx_tagging/core/tagging/models/system_defined.py
```diff
@@ -288,17 +233,49 @@ def _get_available_languages(cls) -> set[str]:
...
+        raise Tag.DoesNotExist
 
-    def _check_tag(self, object_tag: ObjectTag) -> bool:
+    def validate_external_id(self, external_id: str):
         """
-        Returns True if the tag is on the available languages
+        Check if 'external_id' is part of this Taxonomy.
+        """
+        lang_code = external_id.lower()
+        LANGUAGE_DICT = getattr(settings, "LANGUAGE_DICT", dict(settings.LANGUAGES))  # setting may or may not exist.
```
pomegranited, line 259: Oh, I guess `settings.LANGUAGE_DICT` is an old name for `settings.LANGUAGES`? Didn't know that.

But if we need to check both of them, then don't we need to check them both [here](https://github.com/openedx/openedx-learning/pull/87/files#diff-8640a8393169a45e9926177fd6f2e8c7b0ad8439c96554334b95bbe912f52203R240) and [here](https://github.com/openedx/openedx-learning/pull/87/files#diff-8640a8393169a45e9926177fd6f2e8c7b0ad8439c96554334b95bbe912f52203R249) too?

openedx_tagging/core/tagging/models/system_defined.py
```diff
@@ -155,41 +58,101 @@ class ModelSystemDefinedTaxonomy(SystemDefinedTaxonomy):
...
+        external_id = getattr(instance, self.tag_class_key_field)
+        tag, _created = self.tag_set.get_or_create(external_id=external_id, defaults={"value": value})
+        if tag.value != value:
+            # Update the Tag to reflect the new cached 'value'
+            tag.value = value
+            tag.save()
+        return tag
 
 
-class UserModelObjectTag(ModelObjectTag):
```
pomegranited, line 190: Since we're not subclassing ObjectTag for system-defined taxonomies anymore (yay!), the [System Taxonomy creation ADR](https://github.com/openedx/openedx-learning/blob/braden/simplify-tag-models/docs/decisions/0012-system-taxonomy-creation.rst#system-tag-lists-and-validation) should be updated to describe the use of Taxonomy subclasses instead.

pomegranited (approved):
This is really great, @bradenmacdonald !
I left a few inline nits and minor questions, but overall it's so much better than what we had before. I'm happy for you to merge + tag the new version once these are addressed.

:+1:

- [x] I tested this by running this branch + https://github.com/openedx/edx-platform/pull/33378 in my devstack and playing with taxonomies in Django Admin.
- [x] I read through the code
- [x] Includes documentation

openedx_tagging/core/tagging/fixtures/language_taxonomy.yaml
```diff
@@ -1296,3 +1296,4 @@
     allow_multiple: false
     allow_free_text: false
     visible_to_authors: true
```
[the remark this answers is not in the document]
bradenmacdonald: Yes, sorry I wasn't clear. We can delete the languages, but we still need to create the taxonomy. However, if it's just creating a single object, I would prefer to move that to be a data migration instead of a fixture, to keep things simple.

openedx_tagging/core/tagging/models/system_defined.py
bradenmacdonald: Close... Actually settings.LANGUAGES is the normal setting, but it's a list of tuples. For convenience/performance, the `cms`/`lms` also [define](https://github.com/openedx/edx-platform/blob/d60cdc2305fc7ebb99fbffa5fad535a91f9f0be2/cms/envs/common.py#L1227) `LANGUAGE_DICT = dict(LANGUAGES)` which converts it into a dict.

I was already on the fence about whether to use LANGUAGE_DICT or not, but your comment shows to me that it's certainly not clear as it is. So I will add some comments in to clarify before merging.

openedx_tagging/core/tagging/fixtures/language_taxonomy.yaml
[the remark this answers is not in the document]
pomegranited: Ah gotcha.. data migration is fine with me.

openedx_tagging/core/tagging/models/system_defined.py
bradenmacdonald: Good point. I updated it. I'm going to merge now but feel free to leave comments on this diff if you have any, and I'll address them in future PRs.

openedx_tagging/core/tagging/models/base.py
```diff
@@ -674,40 +526,24 @@ def value(self, value: str):
...
+        if self.tag:
+            if self.tag.taxonomy_id != self.taxonomy_id:
+                raise ValidationError("ObjectTag's Taxonomy does not match Tag taxonomy")
+            elif self.tag.value != self._value:
+                raise ValidationError("ObjectTag's _value is out of sync with Tag.value")
+        else:
+            # Note: self.taxonomy and/or self.tag may be NULL which is OK, because it means the Tag/Taxonomy
+            # was deleted, but we still preserve this _value here in case the Taxonomy or Tag get re-created in future.
+            if self._value == "":
+                raise ValidationError("Invalid _value - empty string")
```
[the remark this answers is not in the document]
pomegranited: Oh man! Yes, better to fix that in another PR.

openedx-webhooks:
@bradenmacdonald 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/87&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2023-09-29+06:27&entry.1917517419=2023-10-03+21:36&entry.2133058324=Yes) so we can improve your experience in the future.

bradenmacdonald:
@ChrisChV @pomegranited Thank you for the very nice and timely reviews!

openedx_tagging/core/tagging/models/base.py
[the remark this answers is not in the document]
pomegranited: @bradenmacdonald 

> I'll have to fix that in a future PR because there are too many lint issues showing up now. 

Do you want me to submit a PR for this? I've got time, and there's another couple of minor things that should be fixed:

* [test coverage for files under `tests/` isn't 100%](https://github.com/openedx/openedx-learning/actions/runs/6398610152/job/17369107033#step:7:160); some code is now unused, and some parts of the tests aren't being run.
* [some warnings](https://github.com/openedx/openedx-learning/actions/runs/6398610152/job/17369107033#step:7:49) about the test Taxonomy subclass names -- these classes just need to be renamed.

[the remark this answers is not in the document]
bradenmacdonald: @pomegranited Yes please, that would be great! Thank you :)

openedx_tagging/core/tagging/fixtures/language_taxonomy.yaml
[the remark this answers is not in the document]
pomegranited: And if we remove the fixture, we can also remove `openedx_tagging/core/tagging/management/commands/`.

pomegranited:
@bradenmacdonald 

> TODO: Make Tag.lineage do one query only

I'm not sure this can be done without either a) using a proper tree-aware model, or b) using [recursive CTE which requires MySQL 8](https://dev.mysql.com/doc/refman/8.0/en/with.html#common-table-expressions-recursive). So we mitigated the performance risk by limiting the maximum tree depth to 3.

Also, the use case for `Tag.get_lineage()` has changed since the UI designs came in, so it might be better to return a full serializable list of Tags instead of just their values? Not sure, but I think that'll depend on how the frontend and search backend handle the implicit parent tags.

bradenmacdonald:
@pomegranited You can see what I'm thinking in https://github.com/openedx/openedx-learning/pull/92 . Any thoughts welcome, especially if you think it may not be an improvement in some way. I'm unsure about the performance because I'm doing a lot of JOINs in that version. Once @yusuf-musleh finishes the test data sets, I want to test all of these on some very large taxonomies and see how they perform.

PR #94, opened 2023-10-09
feat: add object tag view permissions [FC-0036]

rpenido:
## Description
Add view permission for ObjectTag

- The `oel_tagging.view_objecttag_objectid` rule is used in the rest API view to check if the user has permission on the object_id to see its tags

## Supporting Information
* https://github.com/openedx/modular-learning/issues/129

## Testing instructions
* Please ensure that the tests cover the expected behavior of the view
____
Private-ref: [FAL-3518](https://tasks.opencraft.com/browse/FAL-3518)

PR #96, opened 2023-10-11
feat: CRUD API for Taxonomy Tags [FC-0036]

yusuf-musleh:
### Description

This PR introduces the CRUD python and REST APIs for interacting with Taxonomy Tags. Creating, updating and deleting them.

### Supporting Information

Related Tickets:
- Closes https://github.com/openedx/modular-learning/issues/136

### Testing Instructions

Make sure all the tests pass, and they cover all the cases.

---
Private-ref: [FAL-3519](https://tasks.opencraft.com/browse/FAL-3519)

### DIFF
--- a/openedx_tagging/core/tagging/api.py
+++ b/openedx_tagging/core/tagging/api.py
@@ -329,7 +329,16 @@ def add_tag_to_taxonomy(
     Taxonomy, an exception is raised, otherwise the newly created
     Tag is returned
     """
-    return taxonomy.cast().add_tag(tag, parent_tag_value, external_id)
+    taxonomy = taxonomy.cast()
+    new_tag = taxonomy.add_tag(tag, parent_tag_value, external_id)
+
+    # Resync all related ObjectTags after creating new Tag to
+    # to ensure any existing ObjectTags with the same value will
+    # be linked to the new Tag
+    object_tags = taxonomy.objecttag_set.all()
+    resync_object_tags(object_tags)
+
+    return new_tag
 
 
 def update_tag_in_taxonomy(taxonomy: Taxonomy, tag: str, new_value: str):
@@ -361,7 +370,3 @@ def delete_tags_from_taxonomy(
     """
     taxonomy = taxonomy.cast()
     taxonomy.delete_tags(tags, with_subtags)
-
-    # Resync all related ObjectTags after deleting the Tag(s)
-    object_tags = taxonomy.objecttag_set.all()
-    resync_object_tags(object_tags)
--- a/openedx_tagging/core/tagging/api.py
+++ b/openedx_tagging/core/tagging/api.py
@@ -321,7 +321,7 @@ def autocomplete_tags(
 def add_tag_to_taxonomy(
     taxonomy: Taxonomy,
     tag: str,
-    parent_tag_id: int | None = None,
+    parent_tag_value: str | None = None,
     external_id: str | None = None
 ) -> Tag:
     """
@@ -329,18 +329,18 @@ def add_tag_to_taxonomy(
     Taxonomy, an exception is raised, otherwise the newly created
     Tag is returned
     """
-    return taxonomy.cast().add_tag(tag, parent_tag_id, external_id)
+    return taxonomy.cast().add_tag(tag, parent_tag_value, external_id)
 
 
-def update_tag_in_taxonomy(taxonomy: Taxonomy, tag: int, tag_value: str):
+def update_tag_in_taxonomy(taxonomy: Taxonomy, tag: str, new_value: str):
     """
     Update a Tag that belongs to a Taxonomy. The related ObjectTags are
     updated accordingly.
 
-    Currently only support updates the Tag value.
+    Currently only supports updating the Tag value.
     """
     taxonomy = taxonomy.cast()
-    updated_tag = taxonomy.update_tag(tag, tag_value)
+    updated_tag = taxonomy.update_tag(tag, new_value)
 
     # Resync all related ObjectTags to update to the new Tag value
     object_tags = taxonomy.objecttag_set.all()
@@ -351,7 +351,7 @@ def update_tag_in_taxonomy(taxonomy: Taxonomy, tag: int, tag_value: str):
 
 def delete_tags_from_taxonomy(
     taxonomy: Taxonomy,
-    tag_ids: list[int],
+    tags: list[str],
     with_subtags: bool
 ):
     """
@@ -360,7 +360,7 @@ def delete_tags_from_taxonomy(
     the sub-tags will be deleted as well.
     """
     taxonomy = taxonomy.cast()
-    taxonomy.delete_tags(tag_ids, with_subtags)
+    taxonomy.delete_tags(tags, with_subtags)
 
     # Resync all related ObjectTags after deleting the Tag(s)
     object_tags = taxonomy.objecttag_set.all()
--- a/openedx_tagging/core/tagging/api.py
+++ b/openedx_tagging/core/tagging/api.py
@@ -129,13 +129,7 @@ def resync_object_tags(object_tags: QuerySet | None = None) -> int:
     if not object_tags:
         object_tags = ObjectTag.objects.select_related("tag", "taxonomy")
 
-    num_changed = 0
-    for object_tag in object_tags:
-        changed = object_tag.resync()
-        if changed:
-            object_tag.save()
-            num_changed += 1
-    return num_changed
+    return ObjectTag.resync_object_tags(object_tags)
 
 
 def get_object_tags(
@@ -329,3 +323,13 @@ def add_tag_to_taxonomy(
     Tag is returned
     """
     return taxonomy.cast().add_tag(tag, parent_tag_id, external_id)
+
+
+def update_tag_in_taxonomy(taxonomy: Taxonomy, tag: int, tag_value: str):
+    """
+    Update a Tag that belongs to a Taxonomy. The related ObjectTags are
+    updated accordingly.
+
+    Currently only support updates the Tag value.
+    """
+    return taxonomy.cast().update_tag(tag, tag_value)
--- a/openedx_tagging/core/tagging/api.py
+++ b/openedx_tagging/core/tagging/api.py
@@ -315,3 +315,17 @@ def autocomplete_tags(
         # remove repeats
         .distinct()
     )
+
+
+def add_tag_to_taxonomy(
+    taxonomy: Taxonomy,
+    tag: str,
+    parent_tag_id: int | None = None,
+    external_id: str | None = None
+) -> Tag:
+    """
+    Adds a new Tag to provided Taxonomy. If a Tag already exists in the
+    Taxonomy, an exception is raised, otherwise the newly created
+    Tag is returned
+    """
+    return taxonomy.cast().add_tag(tag, parent_tag_id, external_id)
--- a/openedx_tagging/core/tagging/api.py
+++ b/openedx_tagging/core/tagging/api.py
@@ -333,3 +333,16 @@ def update_tag_in_taxonomy(taxonomy: Taxonomy, tag: int, tag_value: str):
     Currently only support updates the Tag value.
     """
     return taxonomy.cast().update_tag(tag, tag_value)
+
+
+def delete_tags_from_taxonomy(
+    taxonomy: Taxonomy,
+    tag_ids: list[Tag],
+    with_subtags: bool
+):
+    """
+    Delete Tags that belong to a Taxonomy. If any of the Tags have children and
+    the `with_subtags` is not set to `True` it will fail, otherwise
+    the sub-tags will be deleted as well.
+    """
+    return taxonomy.cast().delete_tags(tag_ids, with_subtags)
--- a/openedx_tagging/core/tagging/api.py
+++ b/openedx_tagging/core/tagging/api.py
@@ -129,7 +129,14 @@ def resync_object_tags(object_tags: QuerySet | None = None) -> int:
     if not object_tags:
         object_tags = ObjectTag.objects.select_related("tag", "taxonomy")
 
-    return ObjectTag.resync_object_tags(object_tags)
+    num_changed = 0
+    for object_tag in object_tags:
+        changed = object_tag.resync()
+        if changed:
+            object_tag.save()
+            num_changed += 1
+
+    return num_changed
 
 
 def get_object_tags(
@@ -332,12 +339,19 @@ def update_tag_in_taxonomy(taxonomy: Taxonomy, tag: int, tag_value: str):
 
     Currently only support updates the Tag value.
     """
-    return taxonomy.cast().update_tag(tag, tag_value)
+    taxonomy = taxonomy.cast()
+    updated_tag = taxonomy.update_tag(tag, tag_value)
+
+    # Resync all related ObjectTags to update to the new Tag value
+    object_tags = taxonomy.objecttag_set.all()
+    resync_object_tags(object_tags)
+
+    return updated_tag
 
 
 def delete_tags_from_taxonomy(
     taxonomy: Taxonomy,
-    tag_ids: list[Tag],
+    tag_ids: list[int],
     with_subtags: bool
 ):
     """
@@ -345,4 +359,9 @@ def delete_tags_from_taxonomy(
     the `with_subtags` is not set to `True` it will fail, otherwise
     the sub-tags will be deleted as well.
     """
-    return taxonomy.cast().delete_tags(tag_ids, with_subtags)
+    taxonomy = taxonomy.cast()
+    taxonomy.delete_tags(tag_ids, with_subtags)
+
+    # Resync all related ObjectTags after deleting the Tag(s)
+    object_tags = taxonomy.objecttag_set.all()
+    resync_object_tags(object_tags)
--- a/openedx_tagging/core/tagging/models/base.py
+++ b/openedx_tagging/core/tagging/models/base.py
@@ -399,10 +399,6 @@ class Taxonomy(models.Model):
         tag = self.tag_set.get(id=tag_id)
         tag.value = tag_value
         tag.save()
-
-        # Resync all related ObjectTags to update to the new Tag value
-        object_tags = self.objecttag_set.all()
-        ObjectTag.resync_object_tags(object_tags)
         return tag
 
     def delete_tags(self, tag_ids: List[int], with_subtags: bool = False):
@@ -709,18 +705,3 @@ class ObjectTag(models.Model):
         self._value = object_tag._value  # pylint: disable=protected-access
         self._name = object_tag._name  # pylint: disable=protected-access
         return self
-
-    @classmethod
-    def resync_object_tags(cls, object_tags: models.QuerySet[ObjectTag]) -> int:
-        """
-        Reconciles ObjectTag entries with any changes made to their associated
-        taxonomies and tags. Return the number of changes made.
-        """
-        num_changed = 0
-        for object_tag in object_tags:
-            changed = object_tag.resync()
-            if changed:
-                object_tag.save()
-                num_changed += 1
-
-        return num_changed
--- a/openedx_tagging/core/tagging/models/base.py
+++ b/openedx_tagging/core/tagging/models/base.py
@@ -345,7 +345,7 @@ class Taxonomy(models.Model):
     def add_tag(
         self,
         tag_value: str,
-        parent_tag_id: int | None = None,
+        parent_tag_value: str | None = None,
         external_id: str | None = None
     ) -> Tag:
         """
@@ -369,8 +369,8 @@ class Taxonomy(models.Model):
             raise ValueError(f"Tag with value '{tag_value}' already exists for taxonomy.")
 
         parent = None
-        if parent_tag_id:
-            parent = self.tag_set.get(id=parent_tag_id)
+        if parent_tag_value:
+            parent = self.tag_set.get(value__iexact=parent_tag_value)
 
         tag = Tag.objects.create(
             taxonomy=self, value=tag_value, parent=parent, external_id=external_id
@@ -378,7 +378,7 @@ class Taxonomy(models.Model):
 
         return tag
 
-    def update_tag(self, tag_id: int, tag_value: str) -> Tag:
+    def update_tag(self, tag: str, new_value: str) -> Tag:
         """
         Update an existing Tag in Taxonomy and return it. Currently only
         supports updating the Tag's value.
@@ -396,12 +396,12 @@ class Taxonomy(models.Model):
             )
 
         # Update Tag instance with new value
-        tag = self.tag_set.get(id=tag_id)
-        tag.value = tag_value
-        tag.save()
-        return tag
+        tag_to_update = self.tag_set.get(value__iexact=tag)
+        tag_to_update.value = new_value
+        tag_to_update.save()
+        return tag_to_update
 
-    def delete_tags(self, tag_ids: List[int], with_subtags: bool = False):
+    def delete_tags(self, tags: List[str], with_subtags: bool = False):
         """
         Delete the Taxonomy Tags provided. If any of them have children and
         the `with_subtags` is not set to `True` it will fail, otherwise
@@ -419,15 +419,15 @@ class Taxonomy(models.Model):
                 "delete_tags() doesn't work for system defined taxonomies. They cannot be modified."
             )
 
-        tags = self.tag_set.filter(id__in=tag_ids)
+        tags_to_delete = self.tag_set.filter(value__in=tags)
 
-        if tags.count() != len(tag_ids):
+        if tags_to_delete.count() != len(tags):
             # If they do not match that means there is a Tag ID in the provided
             # list that is either invalid or does not belong to this Taxonomy
             raise ValueError("Invalid tag id provided or tag id does not belong to taxonomy")
 
         # Check if any Tag contains subtags (children)
-        contains_children = tags.filter(children__isnull=False).distinct().exists()
+        contains_children = tags_to_delete.filter(children__isnull=False).distinct().exists()
 
         if contains_children and not with_subtags:
             raise ValueError(
@@ -436,7 +436,7 @@ class Taxonomy(models.Model):
             )
 
         # Delete the Tags with their subtags if any
-        tags.delete()
+        tags_to_delete.delete()
 
     def validate_value(self, value: str) -> bool:
         """
--- a/openedx_tagging/core/tagging/models/base.py
+++ b/openedx_tagging/core/tagging/models/base.py
@@ -370,10 +370,6 @@ class Taxonomy(models.Model):
 
         parent = None
         if parent_tag_value:
-            # Check if parent tag is valid
-            if not Tag.objects.filter(value__iexact=parent_tag_value).exists():
-                raise ValueError("Invalid `parent_tag_value` provided")
-
             # Get parent tag from taxonomy, raises Tag.DoesNotExist if doesn't
             # belong to taxonomy
             parent = self.tag_set.get(value__iexact=parent_tag_value)
@@ -401,10 +397,6 @@ class Taxonomy(models.Model):
                 "update_tag() doesn't work for system defined taxonomies. They cannot be modified."
             )
 
-        # Check if tag is valid
-        if not Tag.objects.filter(value__iexact=tag).exists():
-            raise ValueError("Invalid `tag` provided")
-
         # Update Tag instance with new value, raises Tag.DoesNotExist if
         # tag doesn't belong to taxonomy
         tag_to_update = self.tag_set.get(value__iexact=tag)
@@ -429,10 +421,6 @@ class Taxonomy(models.Model):
             raise ValueError(
                 "delete_tags() doesn't work for system defined taxonomies. They cannot be modified."
             )
-
-        # Check if tags provided are valid
-        if not Tag.objects.filter(value__in=tags).count() == len(tags):
-            raise ValueError("One or more tag in `tags` is invalid")
 
         tags_to_delete = self.tag_set.filter(value__in=tags)
 
--- a/openedx_tagging/core/tagging/models/base.py
+++ b/openedx_tagging/core/tagging/models/base.py
@@ -370,6 +370,12 @@ class Taxonomy(models.Model):
 
         parent = None
         if parent_tag_value:
+            # Check if parent tag is valid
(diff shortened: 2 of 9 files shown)


openedx-webhooks:
Thanks for the pull request, @yusuf-musleh!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_tagging/core/tagging/models/base.py
```diff
@@ -342,6 +342,70 @@ def get_filtered_tags(
...
+            raise ValueError(
+                "add_tag() doesn't work for system defined taxonomies. They cannot be modified."
+            )
+
+        if self.tag_set.filter(value__iexact=tag_value).exists():
+            raise ValueError(f"Tag with value '{tag_value}' already exists for taxonomy.")
+
+        parent = None
+        if parent_tag_id:
+            parent = Tag.objects.get(id=parent_tag_id)
```
pomegranited, line 374: We should ensure that the `parent` tag is in this taxonomy, e.g.

```suggestion
        if parent_tag_id:
            parent = self.tag_set.get(id=parent_tag_id)
```

openedx_tagging/core/tagging/models/base.py
```diff
@@ -342,6 +342,70 @@ def get_filtered_tags(
...
+                "update_tag() doesn't work for free text taxonomies. They don't use Tag instances."
+            )
+
+        if self.system_defined:
+            raise ValueError(
+                "update_tag() doesn't work for system defined taxonomies. They cannot be modified."
+            )
+
+        # Update Tag instance with new value
+        tag = self.tag_set.get(id=tag_id)
```
pomegranited, line 400: Similarly here, we should catch the possible `Tag.DoesNotExist`, and ensure the view responds with a 404.

openedx_tagging/core/tagging/models/base.py
pomegranited: Also, we should also catch the `Tag.DoesNotExist` that this might throw and ensure the explanation gets back to the user [when calling the view](https://github.com/openedx/openedx-learning/pull/96/files#diff-215d20c663dcd5727c14740a72d09fe6fd22e3a92fd08db46036e2e799699917R549).

pomegranited (changes requested):
This is looking good @yusuf-musleh !
Let me know when you're ready for a full review?

openedx_tagging/core/tagging/models/base.py
yusuf-musleh: Thanks for catching this! I updated it and added a test for this case, I initially was only testing for "invalid" tag ids forgot to check for the case of "valid but for other taxonomy".

openedx_tagging/core/tagging/models/base.py
yusuf-musleh: Addressed as well, thanks!

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -556,3 +634,81 @@ def get_queryset(self) -> list[Tag]:  # type: ignore[override]
...
+
+        tag = body.data.get("tag")
+        parent_tag_id = body.data.get("parent_tag_id", None)
+        external_id = body.data.get("external_id", None)
+
+        try:
+            new_tag = add_tag_to_taxonomy(
+                taxonomy, tag, parent_tag_id, external_id
+            )
+        except Tag.DoesNotExist as e:
```
pomegranited, line 656: I just noticed that the [api exports `TagDoesNotExist = Tag.DoesNotExist`](https://github.com/openedx/openedx-learning/blob/d44775d83e00e642c96077da93d6ed82c06204be/openedx_tagging/core/tagging/api.py#L22), so maybe we should use that instead?

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -399,6 +405,78 @@ class TaxonomyTagsView(ListAPIView):
...
+
+    **Delete Query Returns**
+        * 200 - Success
+        * 400 - Invalid parameters provided
+        * 403 - Permission denied
+        * 404 - Taxonomy not found
+
     """
 
     permission_classes = [TagListPermissions]
```
pomegranited, line 482: @yusuf-musleh I think your best approach would be to modify (and rename) [`TagListPermissions`](https://github.com/openedx/openedx-learning/blob/d44775d83e00e642c96077da93d6ed82c06204be/openedx_tagging/core/tagging/rest_api/v1/permissions.py#L38) to provide a `perm_map` like we did with [`ObjectTagObjectPermissions`](https://github.com/openedx/openedx-learning/blob/d44775d83e00e642c96077da93d6ed82c06204be/openedx_tagging/core/tagging/rest_api/v1/permissions.py#L23).

That will enforce the `oel_tagging.view_tag`, `oel_tagging.change_tag`, and `oel_tagging.delete_tag` rules before invoking the view methods.

But I think you're correct in your question -- these `oel_tagging.*_tag` rules need to be updated to reflect the changes to the taxonomy view/change rules made by https://github.com/openedx/openedx-learning/pull/94.

I think something like this would be sufficient?

```
def can_view_tag(user: UserType, tag: Tag | None = None) -> bool:
    """
    Users can view tags for any taxonomy they can view.
    """"
    taxonomy = tag.taxonomy.cast() if (tag and tag.taxonomy) else None
    return  user.has_perm(
        "oel_tagging.view_taxonomy",
        taxonomy,
    )

def can_change_tag(user: UserType, tag: Tag | None = None) -> bool:
    """
    Users can change tags for any taxonomy they can modify.
    """"
    taxonomy = tag.taxonomy.cast() if (tag and tag.taxonomy) else None
    return  user.has_perm(
        "oel_tagging.change_taxonomy",
        taxonomy,
    )

rules.add_perm("oel_tagging.add_tag", can_change_tag)
rules.add_perm("oel_tagging.change_tag", can_change_tag)
rules.add_perm("oel_tagging.delete_tag", can_change_tag)
rules.add_perm("oel_tagging.view_tag", can_view_tag)
rules.add_perm("oel_tagging.list_tag", can_view_tag)
```

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -1190,3 +1190,492 @@ def test_next_children(self):
...
+        }
+
+        response = self.client.post(
+            self.small_taxonomy_url, create_data, format="json"
+        )
+
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    def test_create_tag_in_taxonomy(self):
+        self.client.force_authenticate(user=self.user)
```
pomegranited, line 1208: Once you've fixed the `TaxonomyTagsView.permission_classes`, you'll need to change your authentication for these tests to `self.staff`. But it's worth leaving in a few checks to ensure that `self.user` is not allowed to create or update tags.

pomegranited (changes requested):
Your view changes are looking great @yusuf-musleh , and you're correct, there's some rules/permission changes needed.

openedx_tagging/core/tagging/rest_api/v1/views.py
yusuf-musleh: Sure thing, updated

openedx_tagging/core/tagging/rest_api/v1/views.py
yusuf-musleh: @pomegranited Thanks for the pointers! I went ahead with what you suggested, however I faced a few issues and took me quite some time to figure them out:
1. The `get_queryset` method in `TaxonomyTagsView` was initially returning a list, as it does some custom population of an attribute `subtags` and later serialized using the `TagsForSearchSerializer`. For the updated permissions (using `perm_map`) to work,`get_queryset` needs to return a queryset as it is used internally by DRF. The current solution I have (though not a huge fan of it) is getting the result list and constructing a queryset using `filter` and then repopulating the `subtags` again if applicable. Ideally, I think the approach should be to refactor the underlying implementation of [get_matching_tags](https://github.com/openedx/openedx-learning/blob/ca6bc859f69d5508d5457f42a61e8fd3263a7b53/openedx_tagging/core/tagging/rest_api/v1/views.py#L468) to use querysets instead. I couldn't get to that today, but I thought I'd mention it here to get your thoughts.
2. After adding the `perms_map` for `TagObjectPermissions`, I left the `has_object_permission` method in place, since it's a special case for the GET list endpoint. There is not `tag` object in that case, we have the Taxonomy object, so we utilize the `can_view_taxonomy` permission for the `oel_tagging.list_tag` case. It gets [called](https://github.com/openedx/openedx-learning/blob/ca6bc859f69d5508d5457f42a61e8fd3263a7b53/openedx_tagging/core/tagging/rest_api/v1/views.py#L429) in the `get_taxonomy` method to and handle whether the use can view the Tags under this Taxonomy or not.

openedx_tagging/core/tagging/api.py
```diff
@@ -129,13 +129,7 @@ def resync_object_tags(object_tags: QuerySet | None = None) -> int:
...
         object_tags = ObjectTag.objects.select_related("tag", "taxonomy")
 
-    num_changed = 0
-    for object_tag in object_tags:
-        changed = object_tag.resync()
-        if changed:
-            object_tag.save()
-            num_changed += 1
-    return num_changed
+    return ObjectTag.resync_object_tags(object_tags)
```
bradenmacdonald, line 132: What's the advantage of making this a static method on `ObjectTag` rather than leaving it here in the API? I kind of prefer having it here since it forces usage of `api`, and we don't want others directly calling methods on `ObjectTag`. Though perhaps we can use the [isolated apps linter](https://github.com/openedx/edx-platform/blob/febcccc14767378492a7a897b66023dafafda619/setup.cfg#L163-L173) to enforce that in the future as well.

I don't have a strong opinion on this, so no need to change it if you think this makes more sense.

openedx_tagging/core/tagging/rest_api/v1/views.py
pomegranited: @yusuf-musleh 

> 1... For the updated permissions (using `perm_map`) to work, `get_queryset` needs to return a queryset as it is used internally by DRF... Ideally, I think the approach should be to refactor the underlying implementation of [get_matching_tags](https://github.com/openedx/openedx-learning/blob/ca6bc859f69d5508d5457f42a61e8fd3263a7b53/openedx_tagging/core/tagging/rest_api/v1/views.py#L468) to use querysets instead.

I agree.. have you seen @bradenmacdonald 's  https://github.com/openedx/openedx-learning/pull/92 ? I think that PR does what you need already. We ran out of budget to complete it, but if it's needed for this work, then maybe we can complete it here?

And what you did for the permissions looks great, thank you.

openedx_tagging/core/tagging/api.py
```diff
@@ -315,3 +309,40 @@ def autocomplete_tags(
...
+) -> Tag:
+    """
+    Adds a new Tag to provided Taxonomy. If a Tag already exists in the
+    Taxonomy, an exception is raised, otherwise the newly created
+    Tag is returned
+    """
+    return taxonomy.cast().add_tag(tag, parent_tag_id, external_id)
+
+
+def update_tag_in_taxonomy(taxonomy: Taxonomy, tag: int, tag_value: str):
```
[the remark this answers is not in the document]
pomegranited: Ooh good points, @bradenmacdonald -- I agree that using `value` where we can makes sense.

openedx_tagging/core/tagging/api.py
pomegranited: I think @yusuf-musleh  did this so he could call `resync` from the `Taxonomy.delete_tags` and `Taxonomy.update_tags` model methods without creating a circular dependency.

But instead, we could leave `resync_object_tags` as-is, and call it from the api `update_tag_in_taxonomy` and `delete_tags_from_taxonomy` methods instead.

What do you think, @yusuf-musleh ?

pomegranited: Also, `resync_object_tags` could take a `taxonomy` as its first argument instead of the ObjectTags queryset, since we're currently only resyncing object tags for single taxonomy.

pomegranited (changes requested):
We're getting there.. I've left a couple of minor comments, but would like to re-review once you've addressed Braden's suggestions.

openedx_tagging/core/tagging/rest_api/v1/permissions.py
```diff
@@ -35,20 +35,21 @@ class ObjectTagObjectPermissions(DjangoObjectPermissions):
...
+        "POST": ["%(app_label)s.add_%(model_name)s"],
+        "PUT": ["%(app_label)s.change_%(model_name)s"],
+        "PATCH": ["%(app_label)s.change_%(model_name)s"],
+        "DELETE": ["%(app_label)s.delete_%(model_name)s"],
+    }
 
+    # This is to handle the special case for GET list of Taxonomy Tags
     def has_object_permission(self, request, view, obj):
         """
         Returns True if the user on the given request is allowed the given view for the given object.
```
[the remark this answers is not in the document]
yusuf-musleh: oh interesting! Thanks for catching this, I managed to reproduce and applied your suggested fix.

openedx_tagging/core/tagging/rules.py
```diff
@@ -51,17 +51,28 @@ def can_change_taxonomy(user: UserType, taxonomy: Taxonomy | None = None) -> boo
...
+def can_view_tag(user: UserType, tag: Tag | None = None) -> bool:
+    """
+    User can view tags for any taxonomy they can view.
+    """
+    taxonomy = tag.taxonomy.cast() if (tag and tag.taxonomy) else None
+    has_perm_thing = user.has_perm(
+        "oel_tagging.view_taxonomy",
+        taxonomy,
+    )
+    return has_perm_thing
```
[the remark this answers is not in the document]
yusuf-musleh: oops I forgot to change it back when debugging, updated it.

openedx_tagging/core/tagging/api.py
yusuf-musleh: I initially moved the logic of `resync_object_tags` to be a static method in the model to avoid duplicate code and to avoid calling api functions from inside the model's methods (like `update_tags`). However I like @pomegranited's suggestion,  makes more sense, I took out the resync logic and placed in in the api and brought back the `resync_object_tags` to the api, removing the static method: https://github.com/openedx/openedx-learning/commit/bdaffeb0d0e718ce56321be6db9a64b83e64bc47

PS: I actually forgot to do the resync of object tags in the delete method, I just added it to the `delete_tags_from_taxonomy`. Thanks for the reminder!

openedx_tagging/core/tagging/api.py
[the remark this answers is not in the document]
yusuf-musleh: I see, that makes sense, I updated the implementation to use values instead of IDs. https://github.com/openedx/openedx-learning/commit/3abce26a6fdd11f89dd4493db79a4d41a008c7c3

openedx_tagging/core/tagging/rest_api/v1/views.py
yusuf-musleh: @pomegranited I went through @bradenmacdonald 's https://github.com/openedx/openedx-learning/pull/92, lots of great work there, and already does alot of the heavy lifting for the refactoring, however since there are substantial changes made in that PR (and some more work remaining there) I feel it might get pretty messy to include it as part of this PR. Would it make sense to leave the current code in this PR as is and then continue working on https://github.com/openedx/openedx-learning/pull/92 and along with the refactoring needed here as part of a separate ticket? 

Alternatively, I could copy the relevant changes from that https://github.com/openedx/openedx-learning/pull/92 and include it as part of this PR? What do you guys think?

bradenmacdonald: > Would it make sense to leave the current code in this PR as is and then continue working on https://github.com/openedx/openedx-learning/pull/92 and along with the refactoring needed here as part of a separate ticket?

Yes, I think that makes sense :)

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -1209,3 +1209,543 @@ def test_next_children(self):
         assert data.get("count") == self.children_tags_count[0]
         assert data.get("num_pages") == 2
         assert data.get("current_page") == 2
+
```
pomegranited, line 1212: Your view tests are great @yusuf-musleh , thank you! Very readable.

pomegranited (changes requested):

openedx_tagging/core/tagging/rest_api/v1/views.py
pomegranited: Created https://github.com/openedx/modular-learning/issues/142 for this.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -192,3 +195,70 @@ def get_children_count(self, obj):
...
+
+    def validate_tags(self, value):
+        """
+        Check that the provided Tags exists based on the values
+        """
+
+        # Build query to check if case-insensitive value for Tags list provided
+        q_list = map(lambda n: Q(value__iexact=n), value)
+        q_list = reduce(lambda a, b: a | b, q_list)
+        valid = Tag.objects.filter(q_list).count() == len(value)
```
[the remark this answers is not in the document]
yusuf-musleh: The implementation in the `Taxonomy.delete_tags` is actually an incorrect one, since the simple `value__in` filter doesnt account for case-insensitve values, the current tests didn't catch it. Thanks for pointing it out, I'll update it.

> Also, do we need to this extra validation in these serializers, since the implementation also enforces the same rules?

My initial thought process was to add the validation in the api incase the functions were called directly rather than only in the rest_api/views (through serializers), however then I figured we could also make use of the validation that DRF provides in addition to that.

The rules they enforce are _slightly_ different, you'll notice the validation that happens in the serializer checks if the Tag values provided are valid (i.e exist in the DB regardless to which taxonomy they belong to) however the validation that happens in the api method checks if the **Taxonomy** has the provided (Tag) values, slight difference where the former would return a 400 and the later would return a 404. I guess the end result is technically the same (the request fails), just thought this give more granularity in checks and errors.

openedx_tagging/core/tagging/rest_api/v1/views.py
yusuf-musleh: Great, thanks for creating it @pomegranited !

openedx_tagging/core/tagging/rest_api/v1/serializers.py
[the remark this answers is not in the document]
yusuf-musleh: > The implementation in the Taxonomy.delete_tags is actually an incorrect one, since the simple value__in filter doesnt account for case-insensitve values, the current tests didn't catch it.

Actually, doing some tests, surprising the simple implementation works 😮, It might be because of the special custom `case_insensitive_char_field` we use in the model? Since that is generally not the default behavior for django's charfields. In any case, that was a pleasant surprise!

openedx_tagging/core/tagging/api.py
```diff
@@ -315,3 +316,52 @@ def autocomplete_tags(
...
+    Delete Tags that belong to a Taxonomy. If any of the Tags have children and
+    the `with_subtags` is not set to `True` it will fail, otherwise
+    the sub-tags will be deleted as well.
+    """
+    taxonomy = taxonomy.cast()
+    taxonomy.delete_tags(tags, with_subtags)
+
+    # Resync all related ObjectTags after deleting the Tag(s)
+    object_tags = taxonomy.objecttag_set.all()
+    resync_object_tags(object_tags)
```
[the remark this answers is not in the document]
yusuf-musleh: Got it, I knew there was a reason I didnt add the resync initially to the delete tags functionality 😅 . And thanks for pointing out the resync that needs to happen when creating new tag. I've updated it with the changes here: https://github.com/openedx/openedx-learning/pull/96/commits/dd35747d6f4d7f4c11057800a206a69186948890

openedx_tagging/core/tagging/rest_api/v1/serializers.py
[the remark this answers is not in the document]
pomegranited: Ah lovely, I'm glad to hear that case-insensitive fields aren't so hard to interact with!

I'm still not convinced that the extra Serializer validation is needed, since the model does what it does and more. But I'll leave that to you & upstream's discretion.

pomegranited (approved):
:+1: Looks great to me @yusuf-musleh ! Ready for your eyes @bradenmacdonald .

Note: needs a version bump before merging.

- [x] I tested this using the django devserver and the DRF REST API UI.
- [x] I read through the code -- great tests!
- [x] <strike>I checked for accessibility issues</strike>
- [x] Includes documentation

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -192,3 +192,67 @@ def get_children_count(self, obj):
         Returns the number of child tags of the given tag.
         """
         return len(obj.sub_tags)
+
+
+class TaxonomyTagCreateBodySerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    """
+    Serializer of the body for the Taxonomy Tags CREATE view
```
bradenmacdonald, line 199:
```suggestion
    Serializer of the body for the Taxonomy Tags CREATE request
```
If you say "view", it's unclear if this is for the request body or the response body.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -192,3 +192,67 @@ def get_children_count(self, obj):
...
+    """
+
+    tag = serializers.CharField(required=True)
+    parent_tag_value = serializers.CharField(
+        source='parent.value', required=False
```
bradenmacdonald, line 204: Is this `source` actually doing anything?

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -192,3 +192,67 @@ def get_children_count(self, obj):
...
+    )
+    external_id = serializers.CharField(required=False)
+
+    def validate_parent_tag_value(self, value):
+        """
+        Check that the provided parent Tag exists based on the value
+        """
+        valid = Tag.objects.filter(value__iexact=value).exists()
+        if not valid:
+            raise serializers.ValidationError("Invalid `parent_tag_value` provided")
```
bradenmacdonald, line 214: Personally I would put this validation in the `Taxonomy.add_tag` method, where most other validation is anyways. That way it gets used for both the python API and REST API. Currently, only the REST API has this validation. You can leave it as-is though.

Removing the validation method(s) from these three serializers would also make the serializer code much more compact.

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -423,9 +430,82 @@ class TaxonomyTagsView(ListAPIView):
         * 400 - Invalid query parameter
         * 403 - Permission denied
         * 404 - Taxonomy not found
+
+    **Create Query Parameters**
+        * pk (required) - The pk of the taxonomy to create a Tag for
```
bradenmacdonald, line 435:
```suggestion
        * id (required) - The ID of the taxonomy to create a Tag for
```

(very minor nit:) Using `pk` to refer to the primary key is kind of an internal django thing that REST API integration developers may not be familiar with. I would simply say `ID` in the documentation strings.

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -423,9 +430,82 @@ class TaxonomyTagsView(ListAPIView):
...
+
+    **Create Request Body**
+        * tag (required): The value of the Tag that should be added to
+          the Taxonomy
+        * parent_tag_value (optional): The value of the parent tag that the new
+          Tag should fall under
+        * extenal_id (optional): The external id for the new Tag
+
+    **Create Example Requests**
+        POST api/tagging/v1/taxonomy/:pk/tags                                       - Create a Tag in taxonomy
```
bradenmacdonald, line 445:
```suggestion
        POST api/tagging/v1/taxonomy/:id/tags                                       - Create a Tag in taxonomy
```

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -423,9 +430,82 @@ class TaxonomyTagsView(ListAPIView):
...
+        PUT api/tagging/v1/taxonomy/:pk/tags                                        - Update a Tag in Taxonomy
+        {
+            "tag": "Tag 1",
+            "updated_tag_value": "Updated Tag Value"
+        }
+        PATCH api/tagging/v1/taxonomy/:pk/tags                                      - Update a Tag in Taxonomy
+        {
+            "tag": "Tag 1",
+            "updated_tag_value": "Updated Tag Value"
+        }
```
bradenmacdonald, line 475: Nit: I think it would be a bit more RESTful to put the old value in the URL and just use `value` as the name of the new value field. Though since `value` is a mutable ID it's not a huge deal. So no need to change for now.

What I _would_ definitely recommend we do now is remove the `PUT` example from the docstring. Because if we add the ability to change other fields like `external_id` via this API endpoint in the future, anyone who is using `PUT` to change the name will now be accidentally overwriting/deleting the `external_id`, since PUT means "completely replace the tag with this new spec". But `PATCH` explicitly means "update the fields I specify to the new values and leave others unchanged".

bradenmacdonald (commented):
Almost there! You don't have to do all of these as changes, just let me know your thoughts.

See also my suggestion on the edx-platform PR about changing how we use `reverse` so that this API can be hosted in any namespace, not just `oel_tagging`

openedx_tagging/core/tagging/rest_api/v1/serializers.py
yusuf-musleh: Got it, updated.

openedx_tagging/core/tagging/rest_api/v1/views.py
yusuf-musleh: Makes sense, updated.

openedx_tagging/core/tagging/rest_api/v1/views.py
yusuf-musleh: I see, for sure, I removed the `PUT` example as suggested, and kept the endpoint as is for now.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
yusuf-musleh: After trying it out, it's definitely more concise and having the validations all in one place is cleaner, thanks @pomegranited and @bradenmacdonald for the suggestions! I updated it.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
yusuf-musleh: I removed it and it doesn't seem like it had any impact. I think this is used more for serializing model/response data rather than handing request data.

yusuf-musleh:
@bradenmacdonald Thanks for the review! I address/applied your suggestions, including the name space suggestion, I also bumped the version, it should hopefully be good to go now.

openedx_tagging/core/tagging/models/base.py
```diff
@@ -342,6 +342,117 @@ def get_filtered_tags(
...
+            )
+
+        if self.tag_set.filter(value__iexact=tag_value).exists():
+            raise ValueError(f"Tag with value '{tag_value}' already exists for taxonomy.")
+
+        parent = None
+        if parent_tag_value:
+            # Check if parent tag is valid
+            if not Tag.objects.filter(value__iexact=parent_tag_value).exists():
+                raise ValueError("Invalid `parent_tag_value` provided")
```
bradenmacdonald, line 375: I don't understand why we even need these lines ^^. Isn't the check below, which raises `Tag.DoesNotExist` sufficient? This just does an extra database query for no reason. We don't really care if the value exists already but in a different taxonomy - the check below already accounts for "value exists in another taxonomy" and "value doesn't exist at all"

bradenmacdonald (approved):
Thanks for those changes! This is good to merge, but I would like to see those duplicate checks removed first, unless there's some reason to keep them in that I'm missing.

openedx_tagging/core/tagging/models/base.py
yusuf-musleh: It is indeed sufficient, I just thought it adds a bit more granularity and description to the errors for the end user. However, you're right, it seems that the extra DB query for this might not be worthwhile. I've removed all of them and updated some test cases.

openedx_tagging/core/tagging/models/base.py
```diff
@@ -342,6 +342,117 @@ def get_filtered_tags(
...
+            )
+
+        if self.system_defined:
+            raise ValueError(
+                "update_tag() doesn't work for system defined taxonomies. They cannot be modified."
+            )
+
+        # Check if tag is valid
+        if not Tag.objects.filter(value__iexact=tag).exists():
+            raise ValueError("Invalid `tag` provided")
```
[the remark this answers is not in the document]
yusuf-musleh: Yup, addressed above.

openedx_tagging/core/tagging/models/base.py
```diff
@@ -342,6 +342,117 @@ def get_filtered_tags(
...
+            )
+
+        if self.system_defined:
+            raise ValueError(
+                "delete_tags() doesn't work for system defined taxonomies. They cannot be modified."
+            )
+
+        # Check if tags provided are valid
+        if not Tag.objects.filter(value__in=tags).count() == len(tags):
+            raise ValueError("One or more tag in `tags` is invalid")
```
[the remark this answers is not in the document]
yusuf-musleh: Similarly, removed and addressed above.

yusuf-musleh:
@bradenmacdonald Thanks for the review! I've updated the PR to remove the duplicate checks, and the version bump is already included.

openedx-webhooks:
@yusuf-musleh 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/96&entry.1484655318=https://github.com/yusuf-musleh&entry.752974735=2023-10-11+14:51&entry.1917517419=2023-10-22+23:07&entry.2133058324=Yes) so we can improve your experience in the future.

PR #105, opened 2023-10-23
Various fixes/enhancements to support v2 libraries import

ormsbee:
Namely:

* allows LearningPackages with data in them to be entirely deleted.
* admin helper for linking to higher level 1:1 models in Django admin–e.g. link from PublishableEntity to Component
* added get/set_draft_version to publishing API


ormsbee:
This is what the admin helper looks like in action:

example_admin_helper

Not having this was causing a surprising amount of friction while I was developing and checking the import command. 😛

kdmccormick:
@ormsbee I need to work on testing and merging a different Libraries V2 PR first, but I think can take a look within the next week.

bradenmacdonald (approved):
LGTM! I like your solution for the admin UI.

PR #110, opened 2023-10-31
docs: ADR for serving static assets

ormsbee:
A first stab at an asset serving ADR that could serve Studio and the LMS. I'm especially looking for feedback on how we could do the cookie auth handoff (or for someone to tell me why it's a terrible idea).

### DIFF
--- a/docs/decisions/0015-serving-static-assets.rst
+++ b/docs/decisions/0015-serving-static-assets.rst
@@ -4,7 +4,7 @@
 Context
 --------
 
-Both Studio and the LMS need to serve course team authored static assets as part of the authoring and learning experiences. These will most often be images, but may also include things like subtitles, audio files, and even JavaScript. It does NOT typically include video files, which are treated separately because of their size.
+Both Studio and the LMS need to serve course team authored static assets as part of the authoring and learning experiences. "Static assets" in the edx-platform context presently refers to: image files, audio files, text document files like PDFs, older video transcript files, and even JavaScript and Python files. It does NOT typically include video files, which are treated separately because of their large file size and complex workflows (processing for multiple resolutions, using third-party dictation services, etc.)
 
 This ADR is the synthesis of various ideas that were discussed across a handful of pull requests and issues. These links are provided for extra context, but they are not required to understand this ADR:
 
@@ -13,6 +13,9 @@ This ADR is the synthesis of various ideas that were discussed across a handful 
 * `Modeling Files and File Dependencies #70 <https://github.com/openedx/openedx-learning/issues/70>`_
 * `Serving static assets (disorganized thoughts) #108 <https://github.com/openedx/openedx-learning/issues/108>`_
 
+Data Storage Implementation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 The underlying data models live in the openedx-learning repo. The most relevant models are:
 
 * `RawContent in contents/models.py <https://github.com/openedx/openedx-learning/blob/main/openedx_learning/core/contents/models.py>`_
@@ -20,9 +23,24 @@ The underlying data models live in the openedx-learning repo. The most relevant 
 
 Key takeaways about how this data is stored:
 
-* Raw asset data is stored in django-storages using its hash value as the file name. This makes it cheap to make many references to the same asset data under different names and versions, but it means that we cannot simply give direct links to the raw file data to the browser.
 * Assets are associated and versioned with Components, where a Component is typically an XBlock. So you don't ask for "version 5 of /static/fig1.webp", you ask for "the /static/fig1.webp associated with version 5 of this block".
-* This initial MVP would be to serve assets for v2 content libraries, where all static assets are associated with a particular component XBlock. Later on, we'll want to allow courses to port their existing files and uploads into this system in a backwards compatible way. We will probably do this by creating a Component that treats the entire course as a Component. The specifics for how that is modeled on the backend are out of scope for this ADR, but this general approach is meant to work for both use cases.
+* This initial MVP would be to serve assets for v2 content libraries, where all static assets are associated with a particular component XBlock. Later on, we'll want to allow courses to port their existing files and uploads into this system in a backwards compatible way. We will probably do this by creating a non-XBlock, filesystem Component type that can treat the entire course as a Component. The specifics for how that is modeled on the backend are out of scope for this ADR, but this general approach is meant to work for both use cases.
+* The actual raw asset data is stored in django-storages using its hash value as the file name. This makes it cheap to make many references to the same asset data under different names and versions, but it means that we cannot simply give direct links to the raw file data to the browser (see the next section for details).
+
+The Difficulty with Direct Links to Raw Data Files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Since the raw data is stored as objects in an S3-like store, and the mapping of file names and versions to that raw data is stored in Django models, why not simply have a Django endpoint that redirects a request to the named asset to the hash-named raw data it corresponds to?
+
+**It will break relative links between assets.**
+  The raw data files exist in a flat space with hashes for names, meaning that any relative links between assets (e.g. a JavaScript file referencing an image) would break once a browser follows the redirect.
+
+**Setting Metadata: Content types, filenames, and caching.**
+  The assets won't generally "work" unless they're served with the correct Content-Type header. For users who want to download a file, it's quite inconvenient if the filename doesn't include the correct file extension (not to mention a friendly name instead of the hash). So we need to set the Content-Type and/or Content-Disposition: ; filename=... headers.
+
+  Setting these values for each request has proved problematic because some (but not all) S3-compatible storage services (including S3 itself) only support setting those headers for each request if you issue a signed GET request, which then gets in the way of caching and introduces the probability of browsers caching expired links, leading to all kinds of annoying cache invalidation issues.
+
+  Setting the filename value at upload time also doesn't work because the same data may be referenced under different filenames by different Components or even different versions of the same Component.
 
 Application Requirements
 ~~~~~~~~~~~~~~~~~~~~~~~~
@@ -51,10 +69,7 @@ Operational Requirements
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 **The asset server must be capable of handling high levels of traffic.**
-  Django views are poor choice for streaming files at scale, especially when deploying using WSGI (as Open edX does), since it will tie down a worker process for the entire duration of the response. While a Django-based streaming response is sufficient for small-to-medium sites, we should allow for a more scalable solution that fully takes advantage of modern CDN capabilities.
-
-**Serving assets should not *require* an S3-like object store.**
-  While S3-like object stores are often used in Open edX deployments at scale, our approach shouldn't *require* it. It is acceptable to require an S3-like store for high-scale traffic.
+  Django views are poor choice for streaming files at scale, especially when deploying using WSGI (as Open edX does), since it will tie down a worker process for the entire duration of the response. While a Django-based streaming response may sufficient for small-to-medium traffic sites, we should allow for a more scalable solution that fully takes advantage of modern CDN capabilities.
 
 **Serving assets should not *require* ASGI deployment.**
   Deploying the LMS and Studio using ASGI would likely substantially improve the scalability of a Django-based streaming solution, but migrating and testing this new deployment type for the entire stack is a large task and is considered out of scope for this project.
@@ -67,7 +82,7 @@ URLs
 
 The format will be: ``https://{asset_server}/assets/apps/{app}/{learning_package_key}/{component_key}/{version}/{filepath}``
 
-The asset server will have a completely different domain from the LMS and Studio, and will not be a subdomain.
+The assets will be served from a completely different domain from the LMS and Studio, and will not be a subdomain.
 
 A more concrete example: ``https://studio.assets.sandbox.openedx.io/apps/content_libraries/lib:Axim:200/xblock.v1:problem@826eb471-0db2-4943-b343-afa65a6fdeb5/v2/static/images/fig1.png``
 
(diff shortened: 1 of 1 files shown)


docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,125 @@
...
+For instance, Studio would likely implement a simple permissions checking model that only examines the learning context and restricts access to course staff. LMS might eventually use a much more sophisticated model that looks at the individual Component that an asset belongs to.
+
+Cookie Authentication
+~~~~~~~~~~~~~~~~~~~~~
+
+Authentication will use a domain cookie on the root assets domain (e.g. ``*.assets.sandbox.openedx.io``). The cookie will have the ``Secure`` and ``HttpOnly`` attributes.
+
+OPEN QUESTIONS:
+
+* What's the best way to do handoff between LMS/Studio and get that cookie information over to the asset domains?
```
ormsbee, line 108: One idea would be to have a one-time key that we store in cache and redirect the user's browser with it.  So it could be something like:

1. Browser makes a request to a new endpoint in Studio/LMS.
2. Studio/LMS generates a random token, and stores user information associated with it in our backend  cache (redis or memecached).
3. Studio/LMS then redirects the request to an auth URL on the static assets server with that random token in the querystring.
4. Asset server finds that entry in the cache, logs that user in, and removes the cache entry.
5. Asset server is now using session auth for that user.

It's cheating in a way because it relies on the asset server and the Studio or LMS instance being on the same service backend. But we're making that the case anyway because of permissions checking integration. And it's simpler than having signed requests (though it could ultimately be extended to go in that direction if we ever want to separate it into an entirely different service).

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,125 @@
...
+* ``draft`` indicating the latest draft version (viewed by authors in Studio).
+* ``published`` indicating the latest published version (viewed by students in the LMS)
+* ``v{num}`` meaning a specific version–e.g. ``v20`` for version 20.
+
+Site Separation
+~~~~~~~~~~~~~~~
+
+The asset-serving site will run in the same process as the LMS or Studio process the links, so ``studio.assets.sandbox.openedx.io`` would point to the same place that ``studio.sandbox.openedx.org``. Being in the same process will make it easier to implement permissions checks and URL generation code.
+
+The switching between the two would happen with a new Middleware class that is implemented in openedx-learning. We would use this Middleware to make sure that assets are never served through the Studio and LMS URLs, and that no other Studio or LMS views are served through the asset server URL. This will require new configuration settings.
```
bradenmacdonald, line 85: I'm not sold on the idea of running the asset server in the same process as the LMS and CMS.

First of all, it is neither specific to the LMS nor the CMS so it doesn't make sense to tie it to one, or the other, or both. Second, in production it should preferably be deployed at the edge. By starting with an in-process implementation, we end up having to re-implement everything for the scalable edge version. The edge version can't do in-process permission checks for example.

I think that eventually, the "asset server" should be a really small, lean JavaScript function that uses web standard APIs ([`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL), `fetch`, `Request`, `Response`), and which makes REST API calls to the LMS as needed to verify authentication and get metadata like content-type etc. This approach can be deployed to almost any modern CDN - Lambda@Edge/CloudFront, CloudFlare Workers, Deno Deploy, Fastly, etc. And those CDNs also each provide a KV store that you can use to cache metadata and permissions.

So for the MVP version, sure, implement it in process in python to keep it simple for now, but maybe put it only in the LMS and make it use a REST API over HTTP to fetch metadata and permissions, so that it will be aligned with the future edge version?

docs/decisions/0015-serving-static-assets.rst
bradenmacdonald, line 108: Maybe:

1. Browser requests a file without any auth headers
2. Edge function requests the metadata from the LMS API (if not already cached) to determine if auth is needed.
3. If no auth is needed, serve the file directly.
4. If auth is needed, return a 302 redirect to a special LMS endpoint. That endpoint redirects back to the CDN but includes an auth header in the query string.
5. Edge function serves the file along with a `Set-Cookie` header, moving the auth token from query string to cookie.

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,125 @@
...
+We could theoretically add masquerading information to the cookie, but we would not implement in the first iteration.
+
+Performance at Scale
+~~~~~~~~~~~~~~~~~~~~
+
+The performance of this system for any particular asset view request should be similar to what already exists today. The difference in the long term is that by tying permissions to individual Components, we will in practice be caching far less than we do today, which will increase the load on the LMS.
+
+The high scale version of this approach will require having a CDN with programmable workers and a scalable object store that supports signed URLs (such as S3). The main objective would be to shift the file streaming burden out of Django and onto the CDN and object store.
+
+In Learning Core, we would implement an APIView (or possibly extend the asset-serving one) to return a JSON response of file metadata. This would include things like size, MIME type, last modified date, cache expiration policy, etc. The response would also contain a signed URL pointing to a object store resource, like S3. The CDN worker then does the fetch on that resource. We would create an example worker for at least CloudFlare.
```
bradenmacdonald, line 122: As mentioned above, I think we should just go with this approach right away, though for now we can substitute a minimal in-process python django app in lieu of an edge function.

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Application Requirements
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Relative links between assets must be preserved.**
+  Assets may reference each other in relative links, e.g. a JavaScript file that references images or other JavaScript files. That means that our solution cannot require querystring-based authorization tokens in the style of S3 signed URLs, since asset files would have no way to encode those into their relative links.
+
+**Multiple versions of the asset should be available at the same time.**
+  Our system should be able to serve at minimum the current draft and published versions of an asset. Ideally, it should be able to serve any version of an asset. This is a departure from the way Studio and the LMS currently handle files and uploads, since there is currently no versioning at all–assets exist in a flat namespace at the course level and are immediately published.
```
connorhaugh, line 34: I see why this is useful, but I want to dig into this a little bit, because this is inherently a costly choice, right? I am curious: what is the process by which an author augments a file to create a new version? Is it re-upload with the same name?

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Security Requirements
+~~~~~~~~~~~~~~~~~~~~~
+
+**Assets must enforce user+file read permissions at the Learning Context level.**
```
connorhaugh, line 39: +1 to this as being a major req.

connorhaugh (commented):
Intriguing. I have some questions above and then a document level question.

I think an implicit choice made is that there is a 1-1 mapping of assets to xblocks. We, however, know that reuse of assets is very common across blocks. How does that square with permissions and versioning?

docs/decisions/0015-serving-static-assets.rst
ormsbee: Yeah, basically. If you have a Problem and `/static/figure1.webp` is reference, and you upload a new `/static/figure1.webp`, then you're creating a new version of that XBlock with that file updated.

It is somewhat costly, since we're holding onto both the old and new images in this case. But at the same time, it gives us the ability to actually batch changes in XBlock content and files together, e.g. for publishing them both at the same time.

ormsbee:
@connorhaugh:

> I think an implicit choice made is that there is a 1-1 mapping of assets to xblocks. We, however, know that reuse of assets is very common across blocks. How does that square with permissions and versioning?

The mapping isn't really 1:1 though. The raw data for the asset lives in a [file referenced](https://github.com/openedx/openedx-learning/blob/51e27248d51be9c3554f3734428e550748b206da/openedx_learning/core/contents/models.py#L102-L108) by a `RawContent` model. But that same asset can exist in many different XBlock components. It can even exist as multiple different file names within the same XBlock if we really want it to, though I can't think of a plausible use case for that.

In terms of re-use, I think there are two scenarios that can play out:

### Incidental re-use, with component-local assets.

We upload a cute cat image in one problem, and then upload it (and a few others) in a second problem in the same library.  We're effectively "re-using" the image data, but not in any way that requires centralized management. The system saves the raw data once, but there are two different rows of [`ComponentVersionRawContent`](https://github.com/openedx/openedx-learning/blob/51e27248d51be9c3554f3734428e550748b206da/openedx_learning/core/components/models.py#L171) that reference it. The two images will be available at two different URLs and can have separate permissions checks applied, since they are addressed as part of two different XBlock components, and the system may give a 403 for one you're not yet allowed to see.

### Intentional re-use, with centralized management.

I think this one's trickier, because there are a couple flavors.

The first kind is linking to the assets of other Components directly. Under the covers, this is how existing Files and Uploads in Studio would be handled (with a Component that represents all the current files and uploads stored for a course). Things could be linked directly, and anyone enrolled in the course would have access to it.

There's also linking to a shared asset where the dependency is more explicitly tracked. Say we have a library of custom grader code, that includes both Python code that can be executed, as well as JavaScript code for certain problem types. Our component XBlock (a ProblemBlock) uses these assets. There are a few broad approaches I can think of:

1. Optimize for the browser by downloading things from the original component location, even if that's in another Learning Package. This is the best for browser caching, especially if we're using it across many different ProblemBlocks in our course, but it can make permissions checking difficult.
2. Make the download URL of the shared content appear to be nested inside the borrowing component–i.e. make it look like the asset belongs to the ProblemBlock and not wherever it originally came from. This simplifies permissions checking because you're only ever checking that you have permissions to the ProblemBlock, not to anything it might be using.
3. Something in between, possibly at the LearningPackage level.

docs/decisions/0015-serving-static-assets.rst
ormsbee: I really think Open edX should be deployable without a CDN.

What would you think about implementing this at the nginx layer [using njs](https://github.com/nginx/njs-examples/#fetch) and/or Caddy layer (with a module)? The CDN would still be very useful for edge caching, but asset serving would still be scalable out of the box. Then there's no need for my middleware proposal, and we can stick with the simple permissions+metadata REST endpoint.

I think we would be _requiring_ an object store at that point, because we can't safely expose the storages media files directly on the LMS or Studio domains... unless we want to build in some kind of auth system that only exposes those URLs to the reverse proxy (e.g. token auth).

> First of all, it is neither specific to the LMS nor the CMS so it doesn't make sense to tie it to one, or the other, or both.

I'm not really sure about that. The content libraries will have assets associated with them, and despite the current configuration, the libraries themselves shouldn't be an LMS thing. We might keep a single asset server instead of my proposed subdomains, but I suspect we'll still want to allow for proxying to an endpoint in either Studio for LMS for the permissions check.

bradenmacdonald: Yes, something like that would be great, and I think would cover a nice range of dev and small instance setups. I agree CDN shouldn't be a requirement, but I would like to see a fairly scalable default with the ability to add a CDN on top when needed. Maybe I'm overcomplicating it though. It's starting to sound like a lot of layers to develop, deploy, and maintain. And anywhere that we create differences between deployment methodologies (object store vs. local, nginx vs. caddy, CDN vs. not) rather than optional layers that are either present or not, we create potential for bugs and confusion where there's differences in the implementation.

> I suspect we'll still want to allow for proxying to an endpoint in either Studio for LMS for the permissions check.

I guess it could also make sense to disallow access to the draft version from the LMS endpoint, so there could be some differences in how they're configured.

ormsbee: > Maybe I'm overcomplicating it though. It's starting to sound like a lot of layers to develop, deploy, and maintain. And anywhere that we create differences between deployment methodologies (object store vs. local, nginx vs. caddy, CDN vs. not) rather than optional layers that are either present or not, we create potential for bugs and confusion where there's differences in the implementation.

So last night I was thinking that maybe we could use a Python service, but a new repo/app that's in an async-based framework that has better streaming performance characteristics than Django. 

But then this morning, I remembered `X-Accel-Redirect`, which I think can cover our needs, and is supported by both nginx and caddy. Instead of a REST response, we'd have Django return the basic content type information for the asset file in the headers, and then set an `X-Accel-Redirect` to the actual file that needs to be done. Then nginx or caddy take things from there with an internal (not browser) redirect.

bradenmacdonald: Oooh, that sounds cool!

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,125 @@
...
+* `Serving static assets (disorganized thoughts) #108 <https://github.com/openedx/openedx-learning/issues/108>`_
+
+The underlying data models live in the openedx-learning repo. The most relevant models are:
+
+* `RawContent in contents/models.py <https://github.com/openedx/openedx-learning/blob/main/openedx_learning/core/contents/models.py>`_
+* `Component and ComponentVersion in components/models.py <https://github.com/openedx/openedx-learning/blob/main/openedx_learning/core/components/models.py>`_
+
+Key takeaways about how this data is stored:
+
+* Raw asset data is stored in django-storages using its hash value as the file name. This makes it cheap to make many references to the same asset data under different names and versions, but it means that we cannot simply give direct links to the raw file data to the browser.
```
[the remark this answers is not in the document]
ormsbee: Added this background info.

docs/decisions/0015-serving-static-assets.rst
ormsbee: Revised the proposal to use `X-Accel-Redirect`.

ormsbee:
Made some major revisions. I'll probably ping the forums for more eyes this coming week.

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,165 @@
...
+The Django view will also be responsible for setting other important header information, such as size, content type, and caching information.
+
+Object Store Requirement
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+**This is the first major piece of courseware functionality that would *require* an object store to work correctly.** It does not require that you configure *all* your media files to go to an object store, but it would require that all new assets in this system go there.
+
+It is possible to add support for a file-backed storages option for this, but there are major drawbacks:
+
+#. To be secure, it would either require sharing the asset raw data files between Studio/LMS and the asset server containers, or it would require serving the assets through Django itself (significantly adding to server load).
```
bradenmacdonald, line 123: > between Studio/LMS and the asset server containers

Is this out of date? Since above it says that the "asset server" is just a different domain name that is served by the LMS/Studio (in the same container).

> it would require serving the assets through Django itself

Doesn't X-accel-redirect solve this problem? You can redirect to a static file on the filesystem, in the media folder.

ormsbee: Actually, maybe this is just me not understanding how we store media files without Minio today. Do you know how that works?  Where do those files actually live, and how does Studio/LMS access them?

bradenmacdonald: On tutor, they're just files in `/openedx/media` and they're served [by uwsgi](https://github.com/overhangio/tutor/blob/10c988127df5bccc37f9b769088bac0c8ebdcff7/tutor/templates/build/openedx/settings/uwsgi.ini#L1-L3).

ormsbee:
@bradenmacdonald: Finally getting back to this. Addressed your last comment about how an object store should no longer be a requirement. Honestly relieved to not have to try to push that change through just yet. 😛

bradenmacdonald (approved):
We'll learn more as we go but this seems like a good direction to me.

regisb (commented):
I learned about this ADR via [your recent comment](https://github.com/openedx/blockstore/issues/314#issuecomment-1924458461) on the Blockstore DEPR issue. I realize I'm very late to the game but I think very strongly that we must reconsider the new requirement of a second top-level domain.

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,154 @@
...
+
+**Assets should enforce more granular permissions at the individual Component level.**
+  An important distinction between ContentStore and v2 Content Library assets is that the latter can be directly associated with a Component. As a long term goal, we should be able to make permissions check on per-Component basis. So if a student does not have permission to view a Component for whatever reason (wrong content group, exam hasn't started, etc.), then they should also not have permission to see static assets associated with that component.
+
+  The further implication of this requirement is that *permissions checking must be extensible*. The openedx-learning repo will implement the details of how to serve an asset, but it will not have the necessary models and logic to determine whether it is allowed to.
+
+**Assets must be served from an entirely different domain than the LMS and Studio instances.**
```
regisb, line 65: Does this requirement apply to all Open edX platforms, and not just edX.org? We can't possibly expect Open edX users to register a second domain name to host their platforms. At the very least, this item should be discussed with the BTR. Personally, I would oppose such a requirement.

regisb (commented):
I learned about this ADR via https://github.com/openedx/blockstore/issues/314#issuecomment-1924458461 on the Blockstore DEPR issue. I realize I'm very late to the game but I think very strongly that we must reconsider the new requirement of a second top-level domain.

docs/decisions/0015-serving-static-assets.rst
ormsbee: @regisb: It would apply to all Open edX platforms.

For now, we'll use a different sub-domain for development purposes (so it would come across like a new service). There are still some long term security risks with that, but it's better than the current state of things. I'll put a writeup together on security tradeoffs before bringing this to the BTR and likely the security working group. We won't add a hard requirement of a new domain for assets before doing that feedback cycle.

That being said, right now I'm really trying to unblock some folks on libraries work, and then I'm taking time off to go back to my hometown for the first time since the pandemic. It's likely that I won't have the time to go through this process until next month.

Thank you for your feedback.

ormsbee: FYI @bradenmacdonald, @kdmccormick: ^

kdmccormick: > I'll put a writeup together on security tradeoffs before bringing this to the BTR and likely the security working group. 

Thanks Dave, I'll keep an eye out for that.

> We can't possibly expect Open edX users to register a second domain name to host their platforms. 

This is a good point, and it makes me wonder if a sibling subdomain would be sufficient for security. For example, if a site operator today had:

* example.org  # lms
* studio.example.org  # cms
* apps.example.org  # mfes

could we move them to:

* assets.example.org  # assets
* lms.example.org   # lms
* studio.example.org  # cms
* apps.example.org  # mfes
* example.org/*  # caddy wildcard redirect to lms.example.org/*, just to preserve existing LMS URLs.

bradenmacdonald: A sibling subdomain only works if you are extremely careful to never set cookies onto the root domain. If you look at tutor today, for example, many cookies like `lms_sessionid` get set onto the root `.local.edly.io` domain. While it may be possible to configure Open edX to work without root domain cookies, there is not really any secure way to enforce this "don't set cookies on the root domain" as a policy, so it remains a potential security issue. Plus it works both ways - if you use a subdomain domain for untrusted content, that untrusted content can _set_ cookies onto your root domain. (Unless it's a [public suffix](https://publicsuffix.org/) like `opencraft.hosting`.)

It's much safer to use a completely unrelated domain.

> We can't possibly expect Open edX users to register a second domain name to host their platforms.

To be clear, _registering_ a second domain name is usually not going to be necessary except for aesthetic reasons.

Almost all production deployments happen on the cloud and virtually all cloud providers provide free domain names on public suffixes, e.g. `d111111abcdef8.cloudfront.net` or `x.s3.us-west-2.amazonaws.com` (AWS), `x.r2.dev` or `x.workers.dev` (CloudFlare), etc. Any of these free domains will suffice.

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,154 @@
...
+  To reduce our chance of maliciously uploaded JavaScript compromising LMS and Studio users, user-uploaded assets must live on an entirely different domain from LMS and Studio (i.e. not just another subdomain). So if our LMS is located at ``sandbox.openedx.org``, the files should be accessed at a URL like ``assets.sandbox.openedx.io``.
```
regisb, line 66: I fail to see how this security issue would be mitigated by hosting the javascript files on a different domain. It seems to me that the cookie access capability of custom javascript depends on the runtime context, not the hosting domain.

For instance, javascript loaded from abc.com, but executed on def.com, would be able to access the cookies from abc.com. What am I missing?

bradenmacdonald: Good point. Actually this is only required assuming that content is being run in an iframe, _and_ the main source code file (HTML) for the iframe is being hosted among the static assets. It's true that as long as the iframe has a different origin than the LMS, it doesn't matter where the actual asset files are stored. I think we need to think this through a bit more. CC @ormsbee

ormsbee: FWIW, people can and occasionally do upload HTML files as static assets.

regisb: Right. I probably don't have the whole picture, but it seems to me that this is a very niche use case. It would be a shame if this specific scenario by itself forced us to setup another domain name.

Is this a use case that we want to preserve, or it a new one? In its current form, isn't it already a security liability, as it allows course staff to run arbitrary scripts?

docs/decisions/0015-serving-static-assets.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Operational Requirements
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+**The asset server must be capable of handling high levels of traffic.**
+  Django views are poor choice for streaming files at scale, especially when deploying using WSGI (as Open edX does), since it will tie down a worker process for the entire duration of the response. While a Django-based streaming response may sufficient for small-to-medium traffic sites, we should allow for a more scalable solution that fully takes advantage of modern CDN capabilities.
```
regisb, line 72: As a side note, this is not the case for uwsgi, which can serve static assets without pausing workers. This is the primary reason why tutor uses uwsgi. But on the other hand uwsgi is no longer maintained, and thus we find ourselves shopping for a replacement, hoping that we can find one with the same feature set.

docs/decisions/0015-serving-static-assets.rst
ormsbee: > Is this a use case that we want to preserve, or it a new one? In its current form, isn't it already a security liability, as it allows course staff to run arbitrary scripts?

It is a liability. One we're hoping to close here.

Whether these use cases should be preserved is a product call. I've seen HTML uploads used by folks who have some fancy JS simulation already built out from some other project, and they want to run it in their course. I've also seen it used by people who have some kind of syllabus from a Word export, where all the images are base64-encoded data urls embedded into the HTML itself, leading to a 20 MB file that would choke the HTMLBlock editor.

I'd be delighted to kick those cases to the curb, but I don't know what other vectors there are. PDFs run JavaScript for form validation these days. I think that's locked down fairly tight and wouldn't have access to cookies, but I'm not sure. A more serious vulnerability is that [JS can be embedded into an SVG file](https://digi.ninja/blog/svg_xss.php). Browsers won't run it when you include SVGs using the `` tag, but if you send someone a link to look at this cool image and they click it, their session could be compromised.

Now there are ways to sanitize it, set up proper content security policies, etc. But I guess my point is that I don't know what else is out there, or will be out there five years from now. Using a separate domain helps to remove this class of vulnerability, which is probably why places like MDN [recommend storing it on a different domain](https://developer.mozilla.org/en-US/docs/Learn/Forms/Sending_and_retrieving_form_data#be_paranoid_never_trust_your_users):

> **Sandbox uploaded files.** Store them on a different server and allow access to the file only through a different subdomain or even better through a completely different domain.

docs/decisions/0015-serving-static-assets.rst
ormsbee: Sure. But even if uwsgi can send files efficiently when its file serving subsystem is invoked, a Django view is still going to tie down the worker process because it's Django doing the streaming in that case. I don't know the specifics of how you'd offload the file serving from the Django view to uwsgi (I'm sure it's possible), but at that point, it's functionally equivalent to using the `X-Sendfile` header and letting caddy do it, isn't it? In either case, the Django view is saying, "I can't send this efficiently, so I'm just going to send the headers for where to find it, and delegate the sending to something with better concurrency/performance than I have."

docs/decisions/0015-serving-static-assets.rst
regisb: Sorry for the slow answer...

Now I understand the general recommendation about hosting uploaded assets on a different subdomain or domain. But that should be a recommendation, not a requirement. It's a whole different thing to say "we recommend you host your uploaded assets on a different domain, but if you understand the risks you can use a subdomain of your LMS, or even the same domain" -- and to actually support those use cases.

feanil: If we do end up going the route of having this be a recommendation instead of a hard requirement, I think we would still want the implementation with two domains to still be the default from a "security by default" perspective.  I don't mind people choosing to not do this but I don't want the software to make it easy to do the "wrong thing."

PR #111, opened 2023-10-31
Improve the "get object tags" API [FC-0036]

bradenmacdonald:
https://github.com/openedx/modular-learning/issues/133

This PR updates the "get object tags" API so that it returns the results grouped by taxonomy, and sorted into "tree order" (just like the "get tags" API does).

I've also added a new API that can retrieve the # of tags applied to each unit for every unit on the course, in a single database query.

### TODO
Before merging, check how this integrates into edx-platform.

### Follow up
I will do a separate PR to allow including "available but unused" taxonomies in the result - those which don't yet have any tags applied to the object, but which the user _could_ use to add additional tags. The permissions of that are somewhat complicated though so it's best as its own PR.

Private ref: [FAL-3547](https://tasks.opencraft.com/browse/FAL-3547)


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -625,7 +565,42 @@ def test_retrieve_object_tags(self, user_attr, expected_status, expected_count):
...
+                                {
+                                    "name": "User Authors",
+                                    "value": "test_user_1",
+                                    "lineage": ["test_user_1"],
+                                },
+                            ],
+                        }
+                    ],
+                },
+            }
```
bradenmacdonald, line 603: @yusuf-musleh If I make the "get object tags" API return data like this, it should make the unit tags overlay much simpler, right?

```python
            assert response.data == {
                "problem15": {
                    "taxonomies": [
                        {
                            "name": "Life on Earth",
                            "taxonomy_id": 1,
                            "editable": True,
                            "tags": [
                                {
                                    "name": "Life on Earth",
                                    "value": "Mammalia",
                                    "lineage": ["Eukaryota", "Animalia", "Chordata", "Mammalia"],
                                },
                                {
                                    "name": "Life on Earth",
                                    "value": "Fungi",
                                    "lineage": ["Eukaryota", "Fungi"],
                                },
                            ]
                        },
                        {
                            "name": "user AUthors",
                            "taxonomy_id": 3,
                            "editable": False,
                            "tags": [
                                {
                                    "name": "User Authors",
                                    "value": "test_user_1",
                                    "lineage": ["test_user_1"],
                                },
                            ],
                        }
                    ],
                },
            }
```

yusuf-musleh: @bradenmacdonald Absolutely! This would make things much simpler. In the currently implementation there's a bunch of stitching stuff together, having this would reduce the API calls needed, especially if the taxonomies returned included taxonomies that are "available" and not only limited to ones that have their tags applied to the object (unit).

bradenmacdonald:
@pomegranited I think this is ready for review, though I may need to make some changes once I try to integrate it into edx-platform. Haven't done that side of it yet.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -61,22 +63,60 @@ class ObjectTagListQueryParamsSerializer(serializers.Serializer):  # pylint: dis
...
+        fields = ["value", "lineage"]
+
+    lineage = serializers.ListField(child=serializers.CharField(), source="get_lineage")
+
+
+class ObjectTagSerializer(ObjectTagMinimalSerializer):
+    """
+    Serializer for the ObjectTag model.
+    """
+    class Meta:
```
pomegranited, line 82: Need to provide Meta.model, otherwise visiting http://127.0.0.1:8000/tagging/rest_api/v1/object_tags/abc/ throws an exception `Class ObjectTagSerializer missing "Meta.model" attribute`

```suggestion
    class Meta:
        model = ObjectTag
```

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -637,34 +697,56 @@ def test_retrieve_object_tags_unauthorized(self):
...
     )
     @ddt.unpack
     def test_retrieve_object_tags_taxonomy_queryparam(
-        self, user_attr, object_id, expected_status, expected_count
+        self, user_attr, object_id, expected_status,
     ):
         """
         Test retrieving object tags for specific taxonomies provided
         """
+        object_id = "html7"
```
pomegranited, line 711: nit: don't need the `object_id` as a param to this test if we're just going to hardcode it here.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
pomegranited: FYI even with the `Meta.model` set, I still get an error in the DRF UI.
But I'm unable to reproduce this issue in tests, and so maybe we don't care?

It happens regardless of whether there are any tags applied to the requesting object.

```
AttributeError at /tagging/rest_api/v1/object_tags/abc/

Got AttributeError when attempting to get a value for field `lineage` on serializer `ObjectTagSerializer`.
The serializer field might be named incorrectly and not match any attribute or key on the `list` instance.
Original exception text was: 'list' object has no attribute 'get_lineage'.
```

pomegranited (approved):
Couple of nits, but this looks and works great @bradenmacdonald ! :+1:

- [x] I tested this using the Django runserver
- [x] I read through the code
- [x] Includes documentation

openedx_tagging/core/tagging/rest_api/v1/serializers.py
bradenmacdonald: > FYI even with the Meta.model set, I still get an error in the DRF UI.
> But I'm unable to reproduce this issue in tests, and so maybe we don't care?

Ah, that was weird but I figured out the error occurs when the UI tries to render the _form_ for this API endpoint, not when using the endpoint itself. To fix it, I just had to mark the `lineage` field of the serializer as `read_only=True`.

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -136,11 +139,11 @@ def test_list_taxonomy(self, user_attr: str | None, expected_status: int):
...
-        Taxonomy.objects.create(name="T1", enabled=True).save()
-        Taxonomy.objects.create(name="T2", enabled=True).save()
-        Taxonomy.objects.create(name="T3", enabled=False).save()
-        Taxonomy.objects.create(name="T4", enabled=False).save()
-        Taxonomy.objects.create(name="T5", enabled=False).save()
+        api.create_taxonomy(name="T1", enabled=True).save()
+        api.create_taxonomy(name="T2", enabled=True).save()
+        api.create_taxonomy(name="T3", enabled=False).save()
+        api.create_taxonomy(name="T4", enabled=False).save()
+        api.create_taxonomy(name="T5", enabled=False).save()
```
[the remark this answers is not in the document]
bradenmacdonald: Oops! Thanks for catching that :)

openedx-webhooks:
@bradenmacdonald 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/111&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2023-10-31+21:56&entry.1917517419=2023-11-02+18:22&entry.2133058324=Yes) so we can improve your experience in the future.

PR #112, opened 2023-11-01
feat: add import taxonomy endpoint

rpenido:
## Description
Add a rest api endpoint to import taxonomies

## Supporting Information
* Part of https://github.com/openedx/modular-learning/issues/115

## Testing instructions
* Please ensure that the tests cover the expected behavior of the view
____
Private-ref: [FAL-3532](https://tasks.opencraft.com/browse/FAL-3532)


openedx-webhooks:
Thanks for the pull request, @rpenido! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -175,3 +176,33 @@ class TaxonomyTagDeleteBodySerializer(serializers.Serializer):  # pylint: disabl
...
+    def get_parser_format(self, obj):
+        """
+        Returns the ParserFormat based on the file extension
+        """
+        filename = obj["file"].name
+        ext = filename.split('.')[-1]
+        if ext.lower() == 'csv':
+            return ParserFormat.CSV
+        elif ext.lower() == 'json':
+            return ParserFormat.JSON
```
pomegranited, line 198: nit: this can be made more general:

```suggestion
        parser_format = getattr(ParserFormat, ext.upper(), None)
        if parser_format:
            return parser_format
```

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -175,3 +176,33 @@ class TaxonomyTagDeleteBodySerializer(serializers.Serializer):  # pylint: disabl
...
+        else:
+            raise serializers.ValidationError(f'File type not supported: ${ext.lower()}')
+
+    def to_internal_value(self, data):
+        """
+        Adds the parser_format to the validated data
+        """
+        validated_data = super().to_internal_value(data)
+        validated_data['parser_format'] = self.get_parser_format(data)
+        return validated_data
```
pomegranited, line 208: You can avoid having to override `to_internal_value` this by adding a `SerializerMethodField` for the `parser_format` field.

ref https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield

```suggestion
```

openedx_tagging/core/tagging/rest_api/v1/views_import.py
```diff
@@ -49,3 +52,47 @@ def get(self, request: Request, file_ext: str, *args, **kwargs) -> FileResponse:
...
+        taxonomy_description = body.validated_data["taxonomy_description"]
+        file = body.validated_data["file"].file
+        parser_format = body.validated_data["parser_format"]
+
+        import_success = api.create_taxonomy_and_import_tags(
+            taxonomy_name=taxonomy_name,
+            taxonomy_description=taxonomy_description,
+            file=file,
+            parser_format=parser_format,
+        )
```
pomegranited, line 93: Hmm.. the import/export tags API was designed to [use async celery tasks](https://github.com/openedx/openedx-learning/blob/05105bcc9e733619650c5041caa91215d9fa36d0/openedx_tagging/core/tagging/import_export/tasks.py) so it could run in the background (in case there's a lot of tags to import).

@bradenmacdonald do you want us to use these async tasks here, or are we doing all imports synchronously?

bradenmacdonald: I suspect synchronous will be totally fine for the short term, but we might as well do it async since we've done most of the work already. How much work will that add to make the REST API layer async too?

openedx_tagging/core/tagging/rest_api/v1/serializers.py
rpenido: I tried this, but it didn't work. I think it works when serializing, not deserializing. The value is filled in the `to_representation` [method](https://github.com/encode/django-rest-framework/blob/f56b85b7dd7e4f786e0769bba6b7609d4507da83/rest_framework/fields.py#L1869-L1871), not in the `to_internal_value`.

openedx_tagging/core/tagging/rest_api/v1/views_import.py
bradenmacdonald: I meant (1). But with simple polling for now, no need for websockets or anything fancy. For (2) it actually ties up much _more_ server resources than just directly doing the import synchronously.

If that's going to take a while to implement though we can just to sync for now.

rpenido: Got it!
There is only one problem in our flow:
We need to create the Taxonomy before calling the import.

If the import fails, it is easy to remove the created taxonomy in the sync call.

Doing async, we can create the taxonomy, return the id to the client, and make calls waiting for the results (the result is associated with the taxonomy). But I need to figure out how to delete the freshly created taxonomy after a failed attempt. Would you happen to have any ideas on this?

This is not urgent for this task, but we will need to handle it if we want to go async sometime in the future. One approach is to change the import flow in the front-end: we first create the taxonomy and always call import inside it (then we don't need to handle the delete). This will impact our designs and this task a bit.

Another approach is to handle the create+import sync and use async for importing on an already created taxonomy. But it will also impact the UX, having different flows.

bradenmacdonald: I think the right approach is to create an additional async task which creates the taxonomy, then synchronously calls the existing import task (which is a celery task so can be called either sync or async, but since we're already in an async task, we call it sync). If that succeeds, it returns the taxonomy ID, and if it fails, it deletes the taxonomy and returns an error message.

The main thing is that this "wrapper" task which does create+import returns an _import ID_ not a taxonomy ID. And then the frontend polls the import ID until it gets either a success or failure message. Only then does it get the taxonomy ID.

--------

But that's all getting too complicated. Just do it synchronously for now an we'll make it async if it's too slow in practice. Explicitly put a comment in the REST API for the import that "this is an unstable API and may change if we make it async."

rpenido:
I think this is ready for a final review @pomegranited !

openedx_tagging/core/tagging/rest_api/v1/serializers.py
pomegranited: Ach, bummer.

But can we avoid the duplicated code in `validate_file` and `get_parser_format`?

openedx_tagging/core/tagging/import_export/api.py
```diff
@@ -48,12 +48,32 @@
...
+    Create a taxonomy and import the tags from `file`
+    """
+    taxonomy = taxonomy_api.create_taxonomy(taxonomy_name, taxonomy_description)
+
+    import_success = import_tags(taxonomy, file, parser_format)
+
+    if not import_success:
+        taxonomy.delete()
+
+    return import_success
```
pomegranited, line 74: The [issue](https://github.com/openedx/modular-learning/issues/115) says:

> 7. If an error occurs, display an alert/toast and put the details in the browser console.

So we need to send any errors logged during import back to the frontend so it can `console.error` it.

I believe that means changing `import_tags` so that it returns the task or the error log as well as the status.

openedx_tagging/core/tagging/rest_api/v1/views_import.py
```diff
@@ -49,3 +53,51 @@ def get(self, request: Request, file_ext: str, *args, **kwargs) -> FileResponse:
...
+            "taxonomy_description": "This is a description",
+            "file": <file>,
+        }
+
+    **Query Returns**
+        * 200 - Success
+        * 400 - Bad request
+        * 405 - Method not allowed
+    """
+    http_method_names = ['post']
```
pomegranited, line 75: We're going to need to the REST views to support both:

* POST - create new taxonomy + import tags
* PUT - update existing taxonomy + import tags

So I thought that the `POST` view could call both `create_taxonomy` and `import_tags` (and `delete_taxonomy` if it failed), and the `PUT` view could just call `import_tags`.

And then we don't need the `create_taxonomy_and_import_tags` API method here or on the platform side.

openedx_tagging/core/tagging/import_export/api.py
```diff
@@ -48,12 +48,32 @@
...
 from django.utils.translation import gettext as _
 
+from .. import api as taxonomy_api
 from ..models import TagImportTask, TagImportTaskState, Taxonomy
 from .exceptions import TagImportError
 from .import_plan import TagImportPlan, TagImportTask
 from .parsers import ParserFormat, get_parser
 
 
+def create_taxonomy_and_import_tags(
```
pomegranited, line 58: See [note](https://github.com/openedx/openedx-learning/pull/112/files#diff-cf610e6b1cd41175eba13b86d368c8d680949a471fbb6d42aaa17e0a852a6764R75) below -- I don't think we need this API method here or on the platform PR.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
rpenido: I'm cheating here or is it okay to do it on the `validate`? d400a5418557b3622d9178be54c4de75372a90f6

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -239,6 +264,57 @@ def export(self, request, **_kwargs) -> HttpResponse:
...
         return HttpResponse(tags, content_type=content_type)
 
+    @action(detail=False, url_path="import", methods=["post"])
+    def create_import(self, request: Request, **_kwargs) -> HttpResponse:
+        """
+        Create a new taxonomy and imports the tags from the uploaded file.
+        """
+        perm = "oel_tagging.import_taxonomy"
+        if not request.user.has_perm(perm):
+            raise PermissionDenied("You do not have permission to import taxonomies")
```
rpenido, line 274: I think we can assume that if a user can create a taxonomy, he also can create a taxonomy via import.
If this is true, we don't need the `import_taxonomy` permission because the POST permission will be called anyway in this endpoint.

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -239,6 +264,57 @@ def export(self, request, **_kwargs) -> HttpResponse:
...
+            return HttpResponseBadRequest(import_error)
+
+    @action(detail=True, url_path="tags/import", methods=["put"])
+    def update_import(self, request: Request, **_kwargs) -> HttpResponse:
+        """
+        Imports tags from the uploaded file to an already created taxonomy.
+        """
+        perm = "oel_tagging.import_taxonomy"
+        if not request.user.has_perm(perm):
+            raise PermissionDenied("You do not have permission to import taxonomies")
```
rpenido, line 301: Same here. If a user can update a taxonomy, he can import tags to this taxonomy.

openedx_tagging/core/tagging/rest_api/v1/views_import.py
rpenido: Made a refactor here: a50e08f1cf1efd3bc92078a6d29a8decc91bc929
I moved the View class to be actions inside our TaxonomyView. That way we can inherit some permissions [here](https://github.com/openedx/openedx-learning/pull/112/files#r1384924676)

openedx_tagging/core/tagging/rest_api/v1/views.py
pomegranited: @rpenido I agree -- simpler is better, so it's ok to rely on the `create` and `change` taxonomy permissions that are checked by the view.

openedx_tagging/core/tagging/rest_api/v1/views.py
pomegranited: Agreed.

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -239,6 +264,63 @@ def export(self, request, **_kwargs) -> HttpResponse:
...
+        taxonomy_description = body.validated_data["taxonomy_description"]
+        file = body.validated_data["file"].file
+        parser_format = body.validated_data["parser_format"]
+
+        taxonomy = create_taxonomy(taxonomy_name, taxonomy_description)
+        try:
+            import_success = import_tags(taxonomy, file, parser_format)
+
+            if import_success:
+                return HttpResponse(status=200)
```
pomegranited, line 289: We should return the created Taxonomy here (serialized with `TaxonomySerializer`) on success.
Then you can also test this response in your view tests, instead of fetching the created instance by taxonomy name.

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -239,6 +264,63 @@ def export(self, request, **_kwargs) -> HttpResponse:
...
+
+        file = body.validated_data["file"].file
+        parser_format = body.validated_data["parser_format"]
+
+        taxonomy = self.get_object()
+        try:
+            import_success = import_tags(taxonomy, file, parser_format)
+
+            if import_success:
+                return HttpResponse(status=200)
```
pomegranited, line 317: Ditto here, can we return the Taxonomy on success?

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -239,6 +264,63 @@ def export(self, request, **_kwargs) -> HttpResponse:
...
+        parser_format = body.validated_data["parser_format"]
+
+        taxonomy = create_taxonomy(taxonomy_name, taxonomy_description)
+        try:
+            import_success = import_tags(taxonomy, file, parser_format)
+
+            if import_success:
+                return HttpResponse(status=200)
+            else:
+                import_error = get_last_import_log(taxonomy)
```
pomegranited, line 291: Hmm.. This works fine in a serial environment, but if we've got multiple imports happening at the same time, this won't necessarily be the import log we're interested in here. Really, the only way to get that is if we have the task ID after import.

But I don't know if it's worth refactoring `import_tags` to support this, so: could you add a comment here to warn us? Then we'll know where to look if users report problems.

openedx_tagging/core/tagging/rules.py
```diff
@@ -172,6 +172,7 @@ def can_change_object_tag(
 rules.add_perm("oel_tagging.delete_taxonomy", can_change_taxonomy)
 rules.add_perm("oel_tagging.view_taxonomy", can_view_taxonomy)
 rules.add_perm("oel_tagging.export_taxonomy", can_view_taxonomy)
+rules.add_perm("oel_tagging.import_taxonomy", can_change_taxonomy)
```
pomegranited, line 175: I agree, this can be removed.
```suggestion
```

pomegranited (changes requested):
This is looking good, and the tests look great @rpenido! Comments inline.

openedx_tagging/core/tagging/rest_api/v1/views.py
rpenido: Done: ae764004976bcc8801b86a409046c843ae3b4721
I went out of scope and did the same for the `export`. If you can view a taxonomy, you can export it. Let me know if it is not the right time to do that, and I revert it!

openedx_tagging/core/tagging/rest_api/v1/views.py
rpenido: This is not as alarming as I made it sound:
1- It only happens on `update`. If we do a mass import/create we will have new ids, so no problem.
2- A did another look at the code and realized that we prevent multiple imports of the same taxonomy simultaneously. So, the concurrent problem may not occur at all, right?

openedx_tagging/core/tagging/rest_api/v1/views.py
rpenido: Yes, but I think we will need to revisit this in the future. 

I think we should return the `Taxonomy` for now, but we will probably want to return the tag list for this taxonomy (at least the first page of it) after the import.

We will have a cleaner view when implementing the Import inside the Taxonomy Details.

And just to make sure: the import inside the Taxonomy Details is not part of this task, right?

openedx_tagging/core/tagging/rest_api/v1/views.py
rpenido: Good catch @pomegranited!
Done: 62e2593e176d2aa4468c49047c0853322e0ffe48

PS: The `Tag` list is not serialized with the `Taxonomy` object. We need a different endpoint for that. I updated the tests to use this endpoint  (as we discussed in another PR) here: a827178ccc74d834ed4dbe085da83d04bcd6b598

rpenido:
ToDo: Bump version before merge

openedx_tagging/core/tagging/rest_api/v1/views.py
pomegranited: Cool, thank you for doing the same for export! That's cleaner all around.

pomegranited (approved):
:+1: This looks great @rpenido ! Ready for your review @bradenmacdonald .

Note: needs a version bump before merging.

- [x] I tested this 
- [x] I read through the code
- [x] Includes documentation

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -1949,3 +1953,392 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+
+@ddt.ddt
+class TestImportTagsView(ImportTaxonomyMixin, APITestCase):
+    """
+    Tests the taxonomy import tags action.
+    """
+    def setUp(self):
+        ImportTaxonomyMixin.setUp(self)
+
+        self.taxonomy = Taxonomy.objects.create(
```
bradenmacdonald, line 2153: Nit: it's not a big deal, but I generally would prefer that we use the API methods `api.create_taxonomy(...)` and `api.add_tag_to_taxonomy(...)` instead of using the models directly. In this case for example, `add_tag_to_taxonomy` does [a lot of additional validation](https://github.com/openedx/openedx-learning/blob/7afd44d0e6fd78d2ab0dfa1124f670a7f23c846b/openedx_tagging/core/tagging/models/base.py#L529-L566) of the created objects, which is good to have, but isn't happening here if we just create the models directly.

The only major exception is in `test_models.py` where we're actually testing the models; in that case obviously we should use the models directly.

bradenmacdonald:
@pomegranited From a very quick look, this seems good to me. You're a CC for this repo so I'm happy for you to approve and merge it, unless you want me to do a second review?

openedx_tagging/core/tagging/rest_api/v1/views.py
pomegranited: @rpenido 

> We will have a cleaner view when implementing the Import inside the Taxonomy Details.

Agreed -- ok to just return Taxonomy for now.

> And just to make sure: the import inside the Taxonomy Details is not part of this task, right?

Correct, importing tags to an existing taxonomy is covered by https://github.com/openedx/modular-learning/issues/140

pomegranited:
@bradenmacdonald Could you do a 2nd review? I'm happy to merge this when it's ready.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -216,3 +217,31 @@ class TaxonomyTagDeleteBodySerializer(serializers.Serializer):  # pylint: disabl
         child=serializers.CharField(), required=True
     )
     with_subtags = serializers.BooleanField(required=False)
+
+
+class TaxonomyImportBodySerializer(serializers.Serializer):  # pylint: disable=abstract-method
+    """
+    Serializer of the body for the Taxonomy Import action
```
bradenmacdonald, line 224:
```suggestion
    Serializer of the body for the Taxonomy Import request
```
I think it's important to distinguish between request and response serializers, when it's only one or the other.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -216,3 +217,31 @@ class TaxonomyTagDeleteBodySerializer(serializers.Serializer):  # pylint: disabl
...
+        if not parser_format:
+            raise serializers.ValidationError({"file": f'File type not supported: {ext.lower()}'}, 'file')
+
+        attrs['parser_format'] = parser_format
+        return attrs
+
+
+class TaxonomyImportNewBodySerializer(TaxonomyImportBodySerializer):  # pylint: disable=abstract-method
+    """
+    Serializer of the body for the Taxonomy Create and Import action
```
bradenmacdonald, line 244:
```suggestion
    Serializer of the body for the Taxonomy Create and Import request
```

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -215,9 +240,6 @@ def export(self, request, **_kwargs) -> HttpResponse:
         Export a taxonomy.
         """
         taxonomy = self.get_object()
-        perm = "oel_tagging.export_taxonomy"
-        if not request.user.has_perm(perm, taxonomy):
-            raise PermissionDenied("You do not have permission to export this taxonomy.")
```
bradenmacdonald, line 220: It's true that "if you can view the taxonomy, you can export it", but are we still checking that the user has "view" permission here? I don't see it.

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -1949,3 +1953,392 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+        for i, tag in enumerate(tags):
+            assert tag["value"] == self.old_tags[i].value
+
+    @ddt.data(
+        "csv",
+        "json",
+    )
+    def test_import_free_text(self, file_format) -> None:
+        """
+        Tests importing a taxonomy with an invalid file content.
```
bradenmacdonald, line 2284:
```suggestion
        Tests that importing tags into a free text taxonomy is not allowed.
```

bradenmacdonald (approved):

bradenmacdonald:
@pomegranited  It looks good to me! Just saw some minor things to clean up.

openedx_tagging/core/tagging/rest_api/v1/views.py
rpenido: The `get_object()` function handles that:

https://github.com/encode/django-rest-framework/blob/1db19f4b2d1500894448634852e02f73043382e4/rest_framework/generics.py#L103

tests/openedx_tagging/core/tagging/test_views.py
rpenido: Even in the `setUp` @bradenmacdonald?

A breaking change will prevent the tests from running and be trickier to diagnose, don't you think?

bradenmacdonald: Well that's the thing - the API is not supposed to have breaking changes in general. But we may change the internal models. So if this test isn't specifically testing the internal models, it should use the API which should be more stable. But again, don't worry about it for now. What you have is fine.

pomegranited:
Thanks for fixing those docstrings @rpenido ! Could you push a version bump when you're done addressing @bradenmacdonald 's review, and I'll get this merged.

rpenido:
> Thanks for fixing those docstrings @rpenido ! Could you push a version bump when you're done addressing @bradenmacdonald 's review, and I'll get this merged.

Done c4fe95d!

openedx-webhooks:
@rpenido 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/112&entry.1484655318=https://github.com/rpenido&entry.752974735=2023-11-01+20:51&entry.1917517419=2023-11-10+02:09&entry.2133058324=Yes) so we can improve your experience in the future.

pomegranited:
[Version 0.3.3 released](https://pypi.org/project/openedx-learning/0.3.3/).

PR #129, opened 2023-12-04
feat: convert mime_type field to its own model.

ormsbee:
This commit normalizes the Media/MIME type in RawContent to be a foriegn key to a new table rather than a text string. This was done for a couple of reasons:

1. Recent problems with edx-platform's courseware_studentmodule table have driven home the point that we should try to reduce index sizes where possible, and RawContent is likely to grow into a very large table in time.
2. Media type designations can sometimes change, and that's much easier to do as a data migration if it's in its own lookup table, rather than affecting millions of rows in RawContent.

In order to support these changes, I've also added LRU caching for MediaType lookup... which in turn required a test helper that will clear out the values between test calls. That's the rationale behind our new openedx_learning.lib.test_utils.TestCase.

One major thing that this PR does that is a big unusual at this stage is that it rewrites the database migrations for `oel_contents` and `oel_components`. I thought this was safe because we haven't actually included these apps into edx-platform yet, and it was cleaner than writing migration code on a non-nullable field.


ormsbee:
@kdmccormick, @bradenmacdonald: The MIME type refactoring I mentioned earlier.

kdmccormick (approved):

PR #135, opened 2023-12-20
fix: Issue when delete all tags on import [FC-0036]

ChrisChV:
## Description
This fix an error when it comes to performing actions on a taxonomy with tags that have `external_id=None`

## More info
Part of: https://github.com/openedx/modular-learning/issues/126

## Testing instructions
Please ensure that the tests cover the expected behavior

## After merge
- [ ] Create a tag to upload the new version

### DIFF
--- a/openedx_tagging/core/tagging/import_export/actions.py
+++ b/openedx_tagging/core/tagging/import_export/actions.py
@@ -259,6 +259,16 @@ class UpdateParentTag(ImportAction):
     def __str__(self) -> str:
         taxonomy_tag = self._get_tag()
 
+        if taxonomy_tag.external_id:
+            prefix_str = _("Update the parent of tag (external_id={external_id})").format(
+                external_id=taxonomy_tag.external_id
+            )
+        else:
+            prefix_str = ""
+            prefix_str = _("Update the parent of tag (value={value})").format(
+                value=taxonomy_tag.value
+            )
+
         if not taxonomy_tag.parent:
             from_str = _("from empty parent")
         else:
@@ -270,17 +280,6 @@ class UpdateParentTag(ImportAction):
                 from_str = _("from parent (value={value})").format(
                     value=taxonomy_tag.parent.value
                 )
-                from_str = ""
-
-        if taxonomy_tag.external_id:
-            prefix_str = _("Update the parent of tag (external_id={external_id})").format(
-                external_id=taxonomy_tag.external_id
-            )
-        else:
-            prefix_str = ""
-            prefix_str = _("Update the parent of tag (value={value})").format(
-                value=taxonomy_tag.value
-            )
 
         return str(
             _(
--- a/openedx_tagging/core/tagging/import_export/actions.py
+++ b/openedx_tagging/core/tagging/import_export/actions.py
@@ -72,7 +72,9 @@ class ImportAction:
         """
         Returns the respective tag of this actions
         """
-        return self.taxonomy.tag_set.get(external_id=self.tag.id)
+        if self.tag.id:
+            return self.taxonomy.tag_set.get(external_id=self.tag.id)
+        return self.taxonomy.tag_set.get(value=self.tag.value)
 
     def _search_action(
         self,
@@ -256,16 +258,35 @@ class UpdateParentTag(ImportAction):
 
     def __str__(self) -> str:
         taxonomy_tag = self._get_tag()
+
         if not taxonomy_tag.parent:
             from_str = _("from empty parent")
         else:
-            from_str = _("from parent (external_id={external_id})").format(external_id=taxonomy_tag.parent.external_id)
+            if taxonomy_tag.parent.external_id:
+                from_str = _("from parent (external_id={external_id})").format(
+                    external_id=taxonomy_tag.parent.external_id
+                )
+            else:
+                from_str = _("from parent (value={value})").format(
+                    value=taxonomy_tag.parent.value
+                )
+                from_str = ""
+
+        if taxonomy_tag.external_id:
+            prefix_str = _("Update the parent of tag (external_id={external_id})").format(
+                external_id=taxonomy_tag.external_id
+            )
+        else:
+            prefix_str = ""
+            prefix_str = _("Update the parent of tag (value={value})").format(
+                value=taxonomy_tag.value
+            )
 
         return str(
             _(
-                "Update the parent of tag (external_id={external_id}) "
+                "{prefix_str} "
                 "{from_str} to parent (external_id={parent_id})."
-            ).format(external_id=taxonomy_tag.external_id, from_str=from_str, parent_id=self.tag.parent_id)
+            ).format(prefix_str=prefix_str, from_str=from_str, parent_id=self.tag.parent_id)
         )
 
     @classmethod
@@ -323,11 +344,18 @@ class RenameTag(ImportAction):
 
     def __str__(self) -> str:
         taxonomy_tag = self._get_tag()
+        if taxonomy_tag.external_id:
+            prefix_str = _("Rename tag value of tag (external_id={external_id})").format(
+                external_id=taxonomy_tag.external_id
+            )
+        else:
+            prefix_str = _("Rename tag value of tag (id={id})").format(id=taxonomy_tag.id)
+
         return str(
             _(
-                "Rename tag value of tag (external_id={external_id}) "
+                "{prefix_str} "
                 "from '{from_value}' to '{to_value}'"
-            ).format(external_id=taxonomy_tag.external_id, from_value=taxonomy_tag.value, to_value=self.tag.value)
+            ).format(prefix_str=prefix_str, from_value=taxonomy_tag.value, to_value=self.tag.value)
         )
 
     @classmethod
@@ -373,7 +401,9 @@ class DeleteTag(ImportAction):
     """
 
     def __str__(self) -> str:
-        return str(_("Delete tag (external_id={external_id})").format(external_id=self.tag.id))
+        if self.tag.id:
+            return str(_("Delete tag (external_id={external_id})").format(external_id=self.tag.id))
+        return str(_("Delete tag (value={value})").format(value=self.tag.value))
 
     name = "delete"
 
@@ -413,7 +443,9 @@ class WithoutChanges(ImportAction):
     name = "without_changes"
 
     def __str__(self) -> str:
-        return str(_("No changes needed for tag (external_id={external_id})").format(external_id=self.tag.id))
+        if self.tag.id:
+            return str(_("No changes needed for tag (external_id={external_id})").format(external_id=self.tag.id))
+        return str(_("No changes needed for tag (value={value})").format(value=self.tag.value))
 
     @classmethod
     def applies_for(cls, taxonomy: Taxonomy, tag) -> bool:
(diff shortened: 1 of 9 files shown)


openedx-webhooks:
Thanks for the pull request, @ChrisChV! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_tagging/core/tagging/import_export/import_plan.py
```diff
@@ -136,7 +136,7 @@ def generate_actions(
 
         if replace:
             tags_for_delete = {
-                tag.external_id: tag for tag in self.taxonomy.tag_set.all()
+                tag.external_id or tag.id: tag for tag in self.taxonomy.tag_set.all()
```
rpenido, line 139: Because of this change, I think we also need to change the code below:
https://github.com/openedx/openedx-learning/blob/fb8d16112fa4705fda8f8ac44cd3401cfb08619e/openedx_tagging/core/tagging/import_export/import_plan.py#L96

Context: this is to prevent updating a tag that could already be deleted

The following test (to be put in `import_export/test_api.pu` reproduce the issue:
```
    def test_import_removing_with_childs_no_external_id(self) -> None:
        """
        Test import need to remove childs with parents that will also be removed,
        using tags without external_id
        """
        new_taxonomy = Taxonomy(name="New taxonomy")
        new_taxonomy.save()
        level2 = Tag.objects.create(
            id=1000,
            value="Tag 2",
            taxonomy=new_taxonomy,
        )
        level1 = Tag.objects.create(
            id=1001,
            value="Tag 1",
            taxonomy=new_taxonomy,
        )
        level3 = Tag.objects.create(
            id=1002,
            value="Tag 3",
            taxonomy=new_taxonomy,
        )
        level2.parent = level1
        level2.save()

        level3.parent = level3
        level3.save()

        # Import with empty tags, to remove all tags
        importFile = BytesIO(json.dumps({"tags": []}).encode())
        result, _tasks, _plan =  import_export_api.import_tags(
            new_taxonomy,
            importFile,
            ParserFormat.JSON,
            replace=True,
        )
        breakpoint()
        assert result
```

ChrisChV: Good catch! Thanks!

ChrisChV: @rpenido  Updated here: https://github.com/openedx/openedx-learning/pull/135/commits/b1c4cc902182b48bf902241f69148693d81422d2

rpenido:
LGTM @ChrisChV! :+1:

- [x] I read through the code
- [x] I checkout the branch and run the tests
- [x] I imported tags over a Taxonomy that originated this bug

@ormsbee If you have time, could you please review/merge this PR while @bradenmacdonald is off?

Thank you!

ormsbee:
@rpenido: I'm also technically off until Jan 2. Is this blocking your work?

rpenido:
> @rpenido: I'm also technically off until Jan 2. Is this blocking your work?

No @ormsbee. This is not urgent and could wait. Sorry to bother you!

Thank you for your response and have a good holiday! :smiley:

openedx_tagging/core/tagging/import_export/actions.py
```diff
@@ -72,7 +72,9 @@ def _get_tag(self) -> Tag:
         """
         Returns the respective tag of this actions
         """
-        return self.taxonomy.tag_set.get(external_id=self.tag.id)
+        if self.tag.id:
+            return self.taxonomy.tag_set.get(external_id=self.tag.id)
```
bradenmacdonald, line 76:
```suggestion
            try:
                return self.taxonomy.tag_set.get(external_id=self.tag.id)
            except Tag.DoesNotExist:
                return self.taxonomy.tag_set.get(value=self.tag.value, external_id=None)
```
What about if the tags were imported without an external ID, but then external IDs were added to the data and the import happens again? I think something like this ^ would handle that situation nicely.

openedx_tagging/core/tagging/import_export/actions.py
```diff
@@ -256,16 +258,34 @@ class UpdateParentTag(ImportAction):
...
     def __str__(self) -> str:
         taxonomy_tag = self._get_tag()
+
+        if taxonomy_tag.external_id:
+            prefix_str = _("Update the parent of tag (external_id={external_id})").format(
+                external_id=taxonomy_tag.external_id
+            )
+        else:
+            prefix_str = ""
+            prefix_str = _("Update the parent of tag (value={value})").format(
```
bradenmacdonald, line 268: Instead of these big `if` statements, I think it would be a much nicer approach to do this:

```
description_str = _("Update the parent of {tag} from parent {old_parent} to {new_parent}").format(
    tag=taxonomy_tag,
    old_parent=taxonomy_tag.parent,
    new_parent=self.tag.parent_id,
)
```

And just make sure that both `Tag` models have a __str__ like:

```
def __str__(self):
    if self.external_id:
        return f'Tag<"{self.value}" ({self.external_id})>'
    return f'Tag<"{self.value}">'
```

What do you think?

openedx_tagging/core/tagging/import_export/actions.py
ChrisChV: Yes you are right, I updated this [here](https://github.com/openedx/openedx-learning/pull/135/files#diff-caeae510c03c9354f48c9cb75e95fc0fb5322180c1b874622c237a461091b6f7R75)

openedx_tagging/core/tagging/import_export/actions.py
ChrisChV: Yes, is cleaner. I updated that here https://github.com/openedx/openedx-learning/pull/135/commits/6181f7c06f6cbcbcdcb3649f96e7ebc0a93cfba3

openedx_tagging/core/tagging/models/base.py
```diff
@@ -94,7 +94,9 @@ def __str__(self):
         """
         User-facing string representation of a Tag.
         """
-        return f"<{self.__class__.__name__}> ({self.id}) {self.value}"
+        if self.external_id:
+            return f"<{self.__class__.__name__}> ({self.id} / {self.external_id} / {self.value})"
+        return f"<{self.__class__.__name__}> ({self.id} / {self.value})"
```
bradenmacdonald, line 99: The `id` is not necessarily going to be stable, so I think this may make the tests flaky. And it doesn't mean anything to users. What do you think about just using `value` and `external_id` in these __str__ and leaving out `id` ?

ChrisChV: The problem is that this function is also used by developers when printing tags in the console. I originally had the id, I wouldn't be too sure if it's a good idea to remove the id, What do you think?

bradenmacdonald: I don't think we should display the ID to users. You could add a separate function like `.display_str()` that doesn't have the ID and use that instead of `__str__()` if you want to keep the ID in `__str__`.

ChrisChV: Thanks, Done. Udpated here https://github.com/openedx/openedx-learning/pull/135/commits/9030b1e8d56fddb7d997dd4b14c4fec4975528d9

ChrisChV:
@bradenmacdonald It's ready for another review

tests/openedx_tagging/core/tagging/import_export/test_import_plan.py
```diff
@@ -340,6 +336,7 @@ def test_plan(self, tags, replace, expected):
         tags = [TagItem(**tag) for tag in tags]
         self.import_plan.generate_actions(tags=tags, replace=replace)
         plan = self.import_plan.plan()
+        print(plan)
```
bradenmacdonald, line 339:
```suggestion
```

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -2457,12 +2457,13 @@ def test_import_plan(self, file_format: str) -> None:
...
             + "--------------------------------\n" \
-            + "#1: Delete tag (external_id=old_tag_1)\n" \
-            + "#2: Delete tag (external_id=old_tag_2)\n" \
+            + "#1: Delete tag <TagItem> (old_tag_1 / Old tag 1)\n" \
+            + "#2: Delete tag <TagItem> (old_tag_2 / Old tag 2)\n" \
             + "#3: Create a new tag with values (external_id=tag_1, value=Tag 1, parent_id=None).\n" \
             + "#4: Create a new tag with values (external_id=tag_2, value=Tag 2, parent_id=None).\n" \
             + "#5: Create a new tag with values (external_id=tag_3, value=Tag 3, parent_id=None).\n" \
             + "#6: Create a new tag with values (external_id=tag_4, value=Tag 4, parent_id=None).\n"
+        print(response.data["plan"])
```
bradenmacdonald, line 2466:
```suggestion
```

tests/openedx_tagging/core/tagging/import_export/test_import_plan.py
```diff
@@ -299,11 +299,11 @@ def test_generate_actions(self, tags, replace, expected_errors, expected_actions
...
-            "#3: Rename tag value of tag (external_id=tag_2) from 'Tag 2' to 'Tag 2 v2'\n"
-            "#4: Update the parent of tag (external_id=tag_4) from parent (external_id=tag_3) "
-            "to parent (external_id=tag_1).\n"
-            "#5: Rename tag value of tag (external_id=tag_4) from 'Tag 4' to 'Tag 4 v2'\n"
-            "#6: No changes needed for tag (external_id=tag_1)\n"
+            "#3: Rename tag value of <Tag> (tag_2 / Tag 2) to 'Tag 2 v2'\n"
+            "#4: Update the parent of <Tag> (tag_4 / Tag 4) from parent "
+            "<Tag> (tag_3 / Tag 3) to tag_1\n"
+            "#5: Rename tag value of <Tag> (tag_4 / Tag 4) to 'Tag 4 v2'\n"
+            "#6: No changes needed for <TagItem> (tag_1 / Tag 1)\n"
```
bradenmacdonald, line 306: From a user perspective, it's confusing that this sometimes says `<TagItem>` and other times says `<Tag>`. We may need to tweak this in the future, especially if we're trying to display this nicely in the frontend. (Though maybe we also need a JSON version of the plan for that.) I think this is OK for now though.

bradenmacdonald (changes requested):
Almost ready to merge. Just some stray `print` calls. Thanks.

tests/openedx_tagging/core/tagging/import_export/test_import_plan.py
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/135/commits/fa42366a09a3d244bcb24403aa916fe52ac75971

tests/openedx_tagging/core/tagging/test_views.py
ChrisChV: Updated here: https://github.com/openedx/openedx-learning/pull/135/commits/fa42366a09a3d244bcb24403aa916fe52ac75971

bradenmacdonald (approved):

openedx-webhooks:
@ChrisChV 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/135&entry.1484655318=https://github.com/ChrisChV&entry.752974735=2023-12-20+22:09&entry.1917517419=2024-01-12+17:46&entry.2133058324=Yes) so we can improve your experience in the future.

PR #138, opened 2024-01-04
Tagging: serialize object permissions to REST API [FC-0036]

pomegranited:
## Description

Adds "user permissions" to the Tagging REST API to show what permissions the current request.user has on the objects serialized in the response.

## Supporting information

Part of https://github.com/openedx/modular-learning/issues/160

## Testing instructions

See https://github.com/openedx/frontend-app-course-authoring/pull/787 for manual testing instructions.

## Deadline

None

## Other information

TO DO (Re [comment](https://github.com/openedx/openedx-learning/pull/138#issuecomment-1879332200):)

- [x] _Please include a test that loads a bunch of taxonomies/tags/objecttags as a non-superuser and makes sure the query count is only increased by at most 1 when permissions are included in the response._

   Done, see [test_views.py](https://github.com/openedx/openedx-learning/pull/138/files#diff-2f11bfdf39bf0bee3b4778c0d84478fa6afd170e57245f804c77b2e9b4d865a4).

- [x] _Do we want to allow API consumers to choose whether the perms info is included in the REST API responses or not?
   It's probably simpler to just always include them, unless there is some performance or response size penalty._
   
   I decided to add an `include_perms` parameter which triggers the permission checks. These checks are cheap to run in this library, but users of the library may customize their permissions (like we did with content_tagging) and so the checks may not always be this quick to run.
   
- [x] bump version

Private ref: FAL-3577


openedx-webhooks:
Thanks for the pull request, @pomegranited! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

bradenmacdonald:
A few thoughts:
* please include a test that loads a bunch of taxonomies/tags/objecttags as a non-superuser and makes sure the query count is only increased by at most 1 when permissions are included in the response
* The permissions for all tags in a taxonomy are always the same, right? And the permissions for an ObjectTag always are the same as the permissions for the Object itself? So I'm tempted to say don't include permissions for them - just include perms at the taxonomy level or course component level. But I realize that's making an assumption that this won't change in the future. If we include the perms now with every Taxonomy/Tag/ObjectTag as you're proposing, and make the frontend work accordingly, then we can be super flexible about how permissions work in the future if we want to, and won't need to make any changes to the frontend. So that makes sense.
* Do we want to allow API consumers to choose whether the perms info is included in the REST API responses or not? It's probably simpler to just always include them, unless there is some performance or response size penalty.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -30,11 +30,71 @@ class TaxonomyExportQueryParamsSerializer(serializers.Serializer):  # pylint: di
...
+    can_change = serializers.SerializerMethodField()
+    can_delete = serializers.SerializerMethodField()
+
+    def _check_permission(self, action, instance) -> bool:
+        """
+        Returns True if the current `request.user` may perform the given `action` on the `instance` object.
+
+        Uses the current request as passed into the serializer context.
+        """
+        assert action in ("add", "change", "view", "delete")
```
rpenido, line 50: nit:
The `add` action differs slightly from the others: we don't have the object instance to check against.

We will probably need a specific endpoint to check if the user can create/import a taxonomy. And we could use an attribute in the Taxonomy instance to check if a user can add tags to it.

pomegranited: >  The add action differs slightly from the others: we don't have the object instance to check against.

Addressed with 8d619c5, thanks!

> We will probably need a specific endpoint to check if the user can create/import a taxonomy. And we could use an attribute in the Taxonomy instance to check if a user can add tags to it.

@bradenmacdonald suggested [doing it this way](https://github.com/openedx/modular-learning/issues/160#issuecomment-1881617717), so that we can see the model level permissions even if an empty list is returned.

I chose to implement it using a custom Paginator class, as the only way I could figure out how to inject this data alongside the [paginated response](https://github.com/encode/django-rest-framework/blob/047bec1288040b9f387e4977bb6cb3fd00d8ea40/rest_framework/pagination.py#L231-L237).

What do you both think of this approach? 23789cc

(To do: add model level permissions to views that use ObjectTagsByTaxonomySerializer, just not sure where yet.)

bradenmacdonald: @pomegranited I agree with @rpenido that it makes more sense to put only the "add" permission in the paginator and the view/change/delete permissions in each object. (Maybe we don't need `view` actually, because if you can't view it, it won't be included in the list. Right?)

@rpenido Good point about ObjectTags permissions depending on both the taxonomy and the object.

@rpenido We don't need to prevent the user from opening the drawer. Studio doesn't currently have a "read-only" view, so if you can see some library content or course content, you can edit it. But yes in the future if you can read tags but not edit them, you should still be able to open the drawer, just not make changes once it's opened.

pomegranited: Thank you for your feedback @rpenido and @bradenmacdonald !

see 634168a for:

* only "can_add" returned at the top level.
* only "can_change" and "can_delete" returned for each object.

Will finish the rest of the feature and ping you for a full review.

bradenmacdonald: Nice @pomegranited , I think that's looking great.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -184,6 +250,33 @@ def get_sub_tags_url(self, obj: TagData | Tag):
...
+
+    def get_can_change(self, instance) -> Optional[bool]:
+        """
+        Returns True if the current user is allowed to edit/change this Tag instance.
+
+        Returns False for all TagData instances.
+        """
+        if isinstance(instance, Tag):
+            return super()._can('change', instance)
+        return None
```
pomegranited, line 268: from [comment](https://github.com/openedx/openedx-learning/pull/138#issuecomment-1879332200):

> The permissions for all tags in a taxonomy are always the same, right? ...  So I'm tempted to say don't include permissions for them - just include perms at the taxonomy level or course component level. But I realize that's making an assumption that this won't change in the future. If we include the perms now with every Taxonomy/Tag/ObjectTag as you're proposing, and make the frontend work accordingly, then we can be super flexible about how permissions work in the future if we want to, and won't need to make any changes to the frontend. So that makes sense.

Here is where that plan breaks down, because we don't currently have a way to check use permissions for TagData instances (rather than Tag instances). I could create a dummy Tag here from the TagData instance and use that to check the permissions, but that will likely trigger another database query (for the `Tag.taxonomy` field).

But we could use the UserPermissionsPagination class to put permissions for the taxonomy itself into the top level like we did for `can_add`?

bradenmacdonald: @pomegranited For now, the permissions of every tag in the taxonomy are the same. But we want to design the REST API to be flexible in the future, as I mentioned. So I would suggest this:

When you initialize a `TagDataSerializer`, pass in the `can_change` and `can_delete` permissions as `context`, just the same way that we currently pass in `taxonomy_id` and `request` as context. Then just change `get_can_change` and `get_can_delete` to return the value from the context, or omit those fields if the data is not available in the context. This will be simple, complete, and works with both `TagData` and `Tag`. It also maintains a flexible API so that if we need perms that vary per-tag in the future, we can change this to do that somehow, without needing to change the REST API or the frontend.

rpenido: I'm failing to find a hypothetical case where we need different Tag and Taxonomy permissions. Do you see any @pomegranited and @bradenmacdonald?

If we can't foresee a use for different permissions, I think we could move closer to removing the Tag permissions and keeping it simple instead of making it more flexible. Of course, only if we can't see a use for it in the future.

So, now, I'm more inclined to use the Taxonomy `canTagObject` permission for Tagging and the Taxonomy `canChange` for Tag editing (that we don't have frontend yet).

pomegranited: @rpenido 

> So, now, I'm more inclined to use the Taxonomy canTagObject permission for Tagging and the Taxonomy canChange for Tag editing (that we don't have frontend yet), skipping serializing permission at TagDataSerializer.

I ended up going with @bradenmacdonald 's suggestion for the TagData:

> When you initialize a TagDataSerializer, pass in the can_change and can_delete permissions as context, just the same way that we currently pass in taxonomy_id and request as context. 

...so we have the permissions data in the REST API whenever we start editing taxonomy tags in the UI.

But I'm equally ok with removing this work if we feel like it's more unused code we shouldn't be maintaining yet.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -60,6 +102,23 @@ def to_representation(self, instance):
...
+        assert request and request.user
+        if not self._include_perms:
+            return None
+
+        model = self._model
+        app_label = model._meta.app_label
+        perm_name = f'{app_label}.add_objecttag'
+        perm_object = ObjectTagPermissionItem(taxonomy=instance, object_id="")
+        return request.user.has_perm(perm_name, perm_object)  # type: ignore[arg-type]
+
```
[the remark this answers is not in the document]
pomegranited: @rpenido 

Since people may override these rules when they use the library, I'd rather call the canonical permission, and let the rule itself [call `change_objecttag_taxonomy`](https://github.com/openedx/openedx-learning/blob/main/openedx_tagging/core/tagging/rules.py#L152-L155).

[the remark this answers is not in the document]
rpenido: Sure! Either way is acceptable to me.

openedx_tagging/core/tagging/rest_api/v1/serializers.py
rpenido: Hi, @pomegranited!
While reviewing, it took me some time to get the permission context again, and I felt we were overengineering a bit with some "to be used in the future" features.
 
If you (who wrote this change) didn't feel this way, we probably don't need to simplify it. If this is the case, I'm ok with the current solutions!

pomegranited: Thank you for your thoughts here @rpenido ! I'll leave this comment thread unresolved so upstream/CCs can see it too, and will see what they say.

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -488,22 +508,24 @@ def update(self, request, *args, **kwargs) -> Response:
...
-            # The obj arg expects a model, but we are passing an object
-            perm_obj,  # type: ignore[arg-type]
-        ):
-            raise PermissionDenied(
-                "You do not have permission to change object tags for this taxonomy or object_id."
-            )
+        perms = [
+            "oel_tagging.add_objecttag",
+            "oel_tagging.delete_objecttag"
+        ]
```
rpenido, line 519: I would prefer if we create an `oel_tagging.can_tag_object` to replace `add_objecttag`, `delete_objecttag`and `change_objecttag`, but I'm fine with this approach.

bradenmacdonald: I agree. `add_objecttag` is ambiguous, because does it mean you can tag a specific object or use a specific taxonomy to tag any object, or you can add object tags in general but maybe not in particular?

`can_tag_object` is very clear what it means - you can tag that object. (add/change/remove tags)

pomegranited: Makes sense to me too, thank you for pushing for consensus here, @rpenido !

See 3dfea40

pomegranited:
@bradenmacdonald  could I get your review here?

bradenmacdonald (commented):
Just a couple minor things to clarify...

openedx_tagging/core/tagging/rest_api/v1/views.py
```diff
@@ -586,6 +606,8 @@ class TaxonomyTagsView(ListAPIView, RetrieveUpdateDestroyAPIView):
           Using full_depth_threshold=1000 is recommended in general, but use lower values during development to ensure
           compatibility with both large and small result sizes.
         * include_counts (optional) - Include the count of how many times each tag has been used.
+        * include_perms (optional) - Include permission checks in response.
+          If omitted, permission fields (can_add_tag, can_change_tag, can_delete_tag) will be None.
```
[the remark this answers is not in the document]
pomegranited: @bradenmacdonald I could do that by changing the Serializer depending on whether `include_perms` is present.

But since we've demonstrated that adding these permissions doesn't add any extra queries, I'm more inclined to remove the `include_perms` parameter and always include permissions in the response.

What would you prefer?

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -174,8 +184,13 @@ def test_list_taxonomy(self, user_attr: str | None, expected_status: int, tags_c
                     "system_defined": False,
                     "visible_to_authors": True,
                     "tags_count": tags_count,
+                    # Enabled taxonomy can be modified by taxonomy admins
+                    "can_change_taxonomy": is_admin,
+                    "can_delete_taxonomy": is_admin,
+                    "can_tag_object": False,
```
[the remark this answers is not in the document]
pomegranited: You'd think so, wouldn't you? But the base [`oel_tagging.can_change_object_tag_objectid`](https://github.com/openedx/openedx-learning/blob/jill/serialize-permissions/openedx_tagging/core/tagging/rules.py#L130-L137) always returns `False`, and so nobody has permission to add base ObjectTags.

I'm more than happy to change this, though, I think it's over cautious.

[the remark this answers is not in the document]
bradenmacdonald: So it's False because the admin could add a ContentObjectTag but not an ObjectTag? Is that what you mean?

If so, is there a similar API that includes `can_tag_content_object` or something? This field just doesn't seem useful at the moment.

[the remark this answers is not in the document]
rpenido: More context here: https://github.com/openedx/openedx-learning/pull/74#discussion_r1313168598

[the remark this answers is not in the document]
pomegranited: @bradenmacdonald 

> So it's False because the admin could add a ContentObjectTag but not an ObjectTag? Is that what you mean?

Yes.

> If so, is there a similar API that includes `can_tag_content_object` or something? This field just doesn't seem useful at the moment.

https://github.com/openedx/edx-platform/pull/34004 shows where `can_tag_object` is True in the content_tagging context, does that help?

openedx_tagging/core/tagging/rest_api/v1/views.py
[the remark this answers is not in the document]
pomegranited: Please see [786ef8b](https://github.com/openedx/openedx-learning/pull/138/commits/786ef8b137ddc93d5b6d6871316f70b56343db5d)

tests/openedx_tagging/core/tagging/test_views.py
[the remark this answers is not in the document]
bradenmacdonald: Yes, that helps :) 
Could you please just add a comment here along those lines? Something like "This is False because we default to not allowing users to tag arbitrary objects. But specific uses of this code like content_tagging in edx-platform will override this perm to be true for their use cases"

bradenmacdonald:
@rpenido @pomegranited Are you both happy with this PR and ready for final review + merge from me?

openedx_tagging/core/tagging/rest_api/v1/serializers.py
```diff
@@ -60,6 +100,16 @@ def to_representation(self, instance):
     def get_tags_count(self, instance):
         return instance.tag_set.count()
 
+    def get_can_tag_object(self, instance) -> Optional[bool]:
```
bradenmacdonald, line 103: Nit: Since we have the "annotations" future import, you can also just write `bool | None` which I find simpler and which doesn't require importing `Optional`

tests/openedx_tagging/core/tagging/test_views.py
```diff
@@ -640,36 +722,42 @@ def test_retrieve_object_tags(self, user_attr, expected_status):
...
                         {
                             "name": "User Authors",
                             "taxonomy_id": 3,
-                            "editable": False,
+                            "can_tag_object": True,
                             "tags": [
                                 {
                                     "value": "test_user_1",
                                     "lineage": ["test_user_1"],
+                                    "can_change_objecttag": True,
```
bradenmacdonald, line 751: I should have asked this earlier but what does it even mean to change an ObjectTag? Aren't they kind of immutable? (I know we should update `_name` and `_value` if the taxonomy/tag itself has changed, but that's an internal synchronization thing, not a user-triggered request).

I guess for free-text tags it could be possible to edit the value but most UX I've seen for free text tagging makes them immutable in the UI and you just have to delete and re-create the tag to do so.

So I'm wondering if we just drop the `can_change_objecttag`. I noticed it wasn't in this doc for example:

Screenshot 2024-01-23 at 11 12 12 AM

bradenmacdonald (approved):
I had a question here but I'm fine to merge either way. Code LGTM.

tests/openedx_tagging/core/tagging/test_views.py
pomegranited: No worries, I've removed `can_change_objecttag`.
One day when we add free-text tags, we may want to resurrect this permission.

pomegranited:
Thanks for your reviews @bradenmacdonald ! I'm going to merge & tag this and get https://github.com/openedx/edx-platform/pull/34004 ready for merge too.

openedx-webhooks:
@pomegranited 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/138&entry.1484655318=https://github.com/pomegranited&entry.752974735=2024-01-04+06:20&entry.1917517419=2024-01-24+00:27&entry.2133058324=Yes) so we can improve your experience in the future.

PR #141, opened 2024-01-09
fix: bug importing tags with the same value as an existing one

rpenido:
## Description
If we try to import tags over a taxonomy that shares the same Tag Value, an error is thrown about the conflict tag value.

It can be reproduced by importing the Template Taxonomy over the "Lightcast Open Skills Taxonomy".
Both taxonomies have a tag with the value "piano". In the plan, it tries to create the tag piano for the Template before deleting the one from Lightcast, raising an error.

This PR changes the order of the import actions, always running the DeleteTag actions before others to avoid this.

## Testing Instructions
Please ensure that the tests cover the expected behavior.

---
Private-ref:
 - [FAL-3603](https://tasks.opencraft.com/browse/FAL-3603)

### DIFF
--- a/openedx_tagging/core/tagging/import_export/import_plan.py
+++ b/openedx_tagging/core/tagging/import_export/import_plan.py
@@ -49,7 +49,7 @@ class TagImportPlan:
         for action in available_actions:
             self.indexed_actions[action.name] = []
 
-    def _build_action(self, action_cls, tag: TagItem):
+    def _build_action(self, action_cls: type[ImportAction], tag: TagItem):
         """
         Build an action with `tag`.
 
@@ -139,6 +139,13 @@ class TagImportPlan:
                 tag.external_id: tag for tag in self.taxonomy.tag_set.all()
             }
 
+            for tag in tags:
+                if tag.id in tags_for_delete:
+                    tags_for_delete.pop(tag.id)
+
+            # Delete all not readed tags
+            self._build_delete_actions(tags_for_delete)
+
         for tag in tags:
             has_action = False
 
@@ -151,13 +158,6 @@ class TagImportPlan:
             if not has_action:
                 # If it doesn't find an action, a "without changes" is added
                 self._build_action(WithoutChanges, tag)
-
-            if replace and tag.id in tags_for_delete:
-                tags_for_delete.pop(tag.id)
-
-        if replace:
-            # Delete all not readed tags
-            self._build_delete_actions(tags_for_delete)
 
     def plan(self) -> str:
         """
--- a/openedx_tagging/core/tagging/import_export/actions.py
+++ b/openedx_tagging/core/tagging/import_export/actions.py
@@ -117,38 +117,48 @@ class ImportAction:
         actions
         """
         try:
-            # Validates if exists a tag with the same value on the Taxonomy
-            taxonomy_tag = self.taxonomy.tag_set.get(value=self.tag.value)
-            return ImportActionError(
-                action=self,
-                message=_(
-                    "Duplicated tag value with tag in database (external_id={external_id})."
-                ).format(external_id=taxonomy_tag.external_id)
-            )
+            is_deleted_tag_value = any(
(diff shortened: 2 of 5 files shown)


openedx-webhooks:
Thanks for the pull request, @rpenido! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_tagging/core/tagging/import_export/import_plan.py
```diff
@@ -139,6 +139,13 @@ def generate_actions(
                 tag.external_id: tag for tag in self.taxonomy.tag_set.all()
             }
 
+            for tag in tags:
```
ChrisChV, line 142: @rpenido Maybe to avoid this loop, you can add the delete actions first using the `_build_action()` func, adding a new param `add_first`. What do you think?

ChrisChV (approved):
@rpenido Looks good :+1:, only one suggestion
- [X] I tested this: I tested that the bug is gone
- [X] I read through the code and considered the security, stability and performance implications of the changes.
- [X] Includes tests for bugfixes and/or features added.
- [X] Includes documentation

openedx_tagging/core/tagging/import_export/import_plan.py
rpenido: My first attempt was just to change the `self.actions.append(action)` to `self.actions.insert(0, action)` for the DeleteTag here:
https://github.com/openedx/openedx-learning/blob/707203268e252a05b2d3454b188cd3f628fc15e6/openedx_tagging/core/tagging/import_export/import_plan.py#L64-L65

This didn't solve the problem because we ran the validation below (for the CreateTag action) in the plan stage (before actually running the delete action).
https://github.com/openedx/openedx-learning/blob/707203268e252a05b2d3454b188cd3f628fc15e6/openedx_tagging/core/tagging/import_export/actions.py#L120-L127

I need to skip this validation if the tag value has a DeleteTag action, so I need to build all the delete actions (calling `_build_delete_actions`) before validating the new tags.

I also tried to avoid this loop filtering while we constructed the dict here:
https://github.com/openedx/openedx-learning/blob/707203268e252a05b2d3454b188cd3f628fc15e6/openedx_tagging/core/tagging/import_export/import_plan.py#L138-L140

But because we have tags (from the DB) and tags (from the importing param), the filtering code became nearly intelligible, so I opted for the loop outside (since it is not a nested loop).

ChrisChV: > My first attempt was just to change the self.actions.append(action) to self.actions.insert(0, action) for the DeleteTag here (this was your suggestion?):

Yes, it is my suggestion.

> But because we have tags (from the DB) and tags (from the importing param), the filtering code became nearly intelligible, so I opted for the loop outside (since it is not a nested loop).

I see, in that case the loop is fine. Thanks for the explanation!

bradenmacdonald (approved):
I would prefer a few more comments to explain what the code is doing, as it's a little hard to follow in some places, but overall this seems fine to me. Thanks!

Should we merge #135 first?

rpenido:
> Should we merge #135 first?

They are independent. You could merge this first.

We will have a minor conflict anyway in the tests because:
- This PR changes the order of the actions
- #135 changes the taxonomy string representation

I added a comment in the code that will conflict.

openedx-webhooks:
@rpenido 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/openedx-learning&entry.867055334=https://github.com/openedx/openedx-learning/pull/141&entry.1484655318=https://github.com/rpenido&entry.752974735=2024-01-09+17:33&entry.1917517419=2024-01-10+22:01&entry.2133058324=Yes) so we can improve your experience in the future.

PR #143, opened 2024-01-18
Changes for v2 content libs

ormsbee:
Work to support https://github.com/openedx/edx-platform/pull/34066

This bumps the version for openedx-learning to 0.5.0.

Some highlights:

* new API functions and queries needed for draft/publish comparisons, fetching & creating component versions, etc.
* re-created the migrations yet again, since this is the last time I should be able to get away with it before it gets installed into edx-platform.
* made some fixes that I realized would bite us later (like making the text collation text insensitive for search purposes later).
* new custom manager to  make it easier to specify common "select_related" queries on models, first used on `Component`.
* a few performance tweaks based on looking at the queries that show up when using this in edx-platform in https://github.com/openedx/edx-platform/pull/34066 (mostly getting rid of n+1 queries)

Squashed commit message:

```
This commit adds a number of features needed to accomodate content
libraries. These are all collected together because this work was done
while iteratively making more and more of libraries functional, and it
was difficult to properly separate out later.

This bumps the version to 0.5.0.

New publishing API functions:

* Learning Packages
    * get_learning_package
    * get_learning_package_by_key,
    * update_learning_package

* Publishable Entities
    * get_publishable_entity and
    * get_publishable_entity_by_key

* Draft/Publishing
    * get_last_publish
    * get_all_drafts
    * get_entities_with_unpublished_changes
    * get_entities_with_unpublished_deletes
    * get_published_version
    * set_draft_version
    * soft_delete_draft
    * reset_drafts_to_published

New Components API functions:

* create_next_version
* get_component_by_key
* component_exists_by_key
* get_components

PublishableEntityMixin improvements:

* New property methods: latest, has_unpublished_changes
* Existing methods were improved to use model-cached values.

Encoding fixes:

* collations for multiple fields were changed to be case insensitive in
  order to better support simple text searches.

Model optimizations:

* created the WithRelationsManager to more easily define querysets that
  involve adding select_related fields. This was used to elininate a
  number of n+1 queries that came up during libraries development.
* PublishableEntityVersion.version_num was reduced to a 4-byte
  PositiveIntegerField instead of using the 8-byte default for
  openedx_learning apps (it's assumed we will never make more than 2
  billion versions of a particular PublishableEntity). This was done to
  help reduce the size of indexes.

Typing:

* type annotations were added to many functions and test functions.
```


ormsbee:
Okay, I think I've gotten all the basic changes in there. Now to fix old tests and add new ones for the new functionality. Then a final re-base and version bump.

ormsbee:
Old tests are fixed, added a few new ones, still some more to go. I'll deal with mypy last.

ormsbee:
@kdmccormick, @bradenmacdonald: If you have some time, please give this a review pass. The tests aren't fully there yet, but I'd appreciate your thoughts on the API additions especially.

openedx_learning/core/publishing/api.py
```diff
@@ -107,6 +162,37 @@ def learning_package_exists(key: str) -> bool:
     return LearningPackage.objects.filter(key=key).exists()
 
 
+def get_last_publish(learning_package_id: int) -> PublishLog | None:
+    return PublishLog.objects \
+                     .filter(learning_package_id=learning_package_id) \
+                     .order_by('-id') \
+                     .first()
+
+def get_all_drafts(learning_package_id: int):
```
bradenmacdonald, line 171: Nit: I think you can set these return types to `QuerySet[Draft]` etc., as long as you have `from __future__ import annotations`

openedx_learning/core/publishing/api.py
```diff
@@ -107,6 +162,37 @@ def learning_package_exists(key: str) -> bool:
...
+    return Draft.objects.filter(
+        entity__learning_package_id=learning_package_id,
+        version__isnull=False,
+    )
+
+def get_entities_with_unpublished_changes(learning_package_id: int):
+    return PublishableEntity.objects \
+               .filter(learning_package_id=learning_package_id) \
+               .exclude(draft__version=F('published__version'))
```
bradenmacdonald, line 180: Nit: super duper minor, but the 11-space indentation of these lines seems totally random. I guess it's four spaces from the start of `PublishableEntity` ? Which is inconsistent with `get_last_publish`.

openedx_learning/core/publishing/api.py
```diff
@@ -107,6 +162,37 @@ def learning_package_exists(key: str) -> bool:
...
+
+def get_entities_with_unpublished_deletes(learning_package_id: int):
```
bradenmacdonald, line 182: What's the intended use case of this API method? Perhaps a widget in the UI that says something like "These things have been deleted since your last publish: ____" ?

openedx_learning/core/publishing/model_mixins.py
```diff
@@ -129,59 +134,110 @@ def __init__(self, content_obj):
...
             Return the content version object that is the current draft.
+
+            So if you mix ``PublishableEntityMixin`` into ``Component``, then
+            ``component.versioning.draft`` will return you the
+            ``ComponentVersion`` that is the current draft (not the underlying
+            ``PublishableEntityVersion``).
+
+            If this is causing many queries, it might be the case that you need
+            to add ``select_related('publishable_entity__draft__version')`` to
+            the queryset.
```
bradenmacdonald, line 160: If you want, [you can use `Field.is_cached` to check this in code with zero db overhead](https://stackoverflow.com/a/68200269), and display a warning to the developer at runtime. But though it's not a private method, it's technically undocumented functionality.

tests/openedx_learning/core/components/test_api.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+class TestPerformance(TestCase):
+    """
+    Performance related tests to make sure we don't get n + 1 queries.
+    """
+    @classmethod
+    def setUpTestData(cls) -> None:
+        """
+        Initialize our content data (all our tests are read only).
```
bradenmacdonald, line 21: Nit: I don't think the tests are actually read only? They're doing stuff like `create_component_and_version`. Maybe you mean "None of the tests modify the Learning Package itself" ?

tests/openedx_learning/core/components/test_api.py
```diff
@@ -0,0 +1,439 @@
...
+        )
+        publishing_api.publish_all_drafts(
+            self.learning_package.pk,
+            published_at=self.now
+        )
+
+        # We should be fetching all of this with a select-related, so only one
+        # database query should happen here.
+        with self.assertNumQueries(1):
+            component = components_api.get_component(component.pk)
```
bradenmacdonald, line 54: Nit: the docstring of this class says it's "to make sure we don't get n + 1 queries", but calling `get_component` for a single component is not a situation in which you're going to see n + 1 queries. I guess you're still fleshing this out though. I'll look at the rest of the tests later when this is ready :)

bradenmacdonald (commented):
These API changes look great @ormsbee ! All my notes here are very minor nits, so feel free to ignore.

tests/openedx_learning/core/components/test_api.py
ormsbee: Oops. Yeah, sloppy wording on my part, will fix, ty.

tests/openedx_learning/core/components/test_api.py
ormsbee: I meant that there aren't additional queries to get the draft and published versions of things, but yeah, that's not n+1. Will re-phrase.

openedx_learning/core/components/api.py
```diff
@@ -98,10 +166,84 @@ def create_component_and_version(
...
+    if draft is not None:
+        qset = qset.filter(publishable_entity__draft__version__isnull = not draft)
+    if published is not None:
+        qset = qset.filter(publishable_entity__published__version__isnull = not published)
+    if namespace is not None:
+        qset = qset.filter(namespace=namespace)
+    if types is not None:
+        qset = qset.filter(type__in=types)
+    if title is not None:
+        qset = qset.filter(publishable_entity__draft__version__title__icontains=title)
```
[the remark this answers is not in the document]
ormsbee: Good point. I'll make `draft_title` and `published_title` as separate fields that can be passed in.

openedx_learning/core/components/api.py
```diff
@@ -98,10 +166,84 @@ def create_component_and_version(
...
+    Get Component by its primary key.
+
+    This is the same as the PublishableEntity's ID primary key.
+    """
+    return Component.with_publishing_relations.get(pk=component_pk)
 
-def get_component_by_pk(component_pk: int) -> Component:
-    return Component.objects.get(pk=component_pk)
+def get_component_by_key(
+    learning_package_id: int,
```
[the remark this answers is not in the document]
ormsbee: Good point. Maybe I should stick with `_pk` for everything?

The reason why it's not `component_id` is because there is no `Component.id` field–Components have a `OneToOneField` relationship to `PublishableEntity` with `primary_key=True`. So the actual database field is `oel_components_component.publishable_entity_id`.

So maybe I should just convert everything to be `_pk` for consistency instead, since `learning_package.pk` will just be an alias to `learning_package.id` anyway. But one of the things that I didn't think through at the time is that if you have a foreign key relation to `Component` from another model, the db field will be `component_id`, and `component_version.component_pk` doesn't work.

openedx_learning/core/components/api.py
```diff
@@ -73,14 +73,82 @@ def create_component_version(
...
+            component_id=component_pk,
+        )
+        # First copy the new stuff over...
+        for key, raw_content_pk in content_to_replace.items():
+            ComponentVersionRawContent.objects.create(
+                raw_content_id=raw_content_pk,
+                component_version=component_version,
+                key=key,
+                learner_downloadable=False,
+            )
```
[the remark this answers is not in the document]
ormsbee: Yeah. I was first punting just because I was trying to implement only exactly what I needed to get v2 libraries over, but I should just add it. Thank you.

openedx_learning/core/components/api.py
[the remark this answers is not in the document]
bradenmacdonald: Personally I don't think it's a big deal either way, but I'd probably lean toward calling it `component_id` for consistency, even though `Component.id` doesn't exist.

openedx_learning/core/publishing/api.py
ormsbee: I'm not entirely sure, actually. It's returned as [part of the metadata in the `get_library` call](https://github.com/openedx/edx-platform/blob/5838d68efc1bd29d9dda593d1d0a1306654e58ba/openedx/core/djangoapps/content_libraries/api.py#L393-L394).

ormsbee: In the [new code](https://github.com/openedx/edx-platform/pull/34066/files#diff-a760c3e87f2f671ac2453abbbaf6f10ca052c7acf1c46c1d36f24646e1d52a00R334-R339), I just run `.exists()` on the queryset returned here.

openedx_learning/core/components/api.py
```diff
@@ -73,14 +73,82 @@ def create_component_version(
     return component_version
 
 
+def create_next_version(
+    component_pk: int,
+    title: str,
+    content_to_replace: dict[str: int],
```
kdmccormick, line 79:
```suggestion
    content_to_replace: dict[str, int],
```

openedx_learning/core/publishing/api.py
ormsbee: Yeah, I was trying out a few different things. I think establishing a decent convention is actually going to matter for this API because it has long function names and encourages chaining query sets.

How would you feel about this alternative?

```python
    return (
        PublishableEntity
            .objects
            .filter(
                draft__version__isnull=True,
            )
            .exclude(published__version__isnull=True)
    )
```

I think it looks a little funny at first, but it gives the most space for chaining purposes and long variable assignment, it doesn't require using backslashes.

openedx_learning/core/publishing/model_mixins.py
ormsbee: Thanks. I think I'll avoid it for now, but I'll keep it in mind if performance regressions along these lines starts to become a problem.

openedx_learning/core/components/api.py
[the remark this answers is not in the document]
ormsbee: Implementing this now. FWIW, I want to avoid any inference here (e.g. `published=True` implying that a title check should happen against the published version only). I figure any smarts like that should be in the layer above, and these query fields will just map one-to-one to know fields–hence separating `title` into separate parameters for draft vs. published titles.

openedx_learning/core/publishing/api.py
ormsbee: Turns out that this offends the linter. I guess I'll go with:

```python
def get_entities_with_unpublished_changes(learning_package_id: int):
    return PublishableEntity.objects \
                            .filter(learning_package_id=learning_package_id) \
                            .exclude(draft__version=F('published__version'))
```

openedx_learning/core/components/api.py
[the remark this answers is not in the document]
kdmccormick: @ormsbee @bradenmacdonald Consider me a soft vote for `_pk`. In edx-platform, `_id` could indicate all sorts of things: primary db keys, opaque keys (course_id, library_id, usage_id), and other things (block_id, user_id). `_pk` is unambiguously a primary db key.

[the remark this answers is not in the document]
ormsbee: Yeah, I'm going to go with `_pk` everywhere and do that refactoring now.

ormsbee:
Okay, I wrote the tests, made changes based on the review comments, and the linters are appeased. I'm going to do the `_pk` refactoring now, but with any luck I'm hoping this can merge today.

ormsbee:
Okay, slight delay because I notice that it's always doing the work of loading all components no matter the pagination params in the REST API. I think this is a problem on the edx-platform side, but I'm looking into it to be sure.

openedx_learning/core/components/api.py
```diff
@@ -73,15 +73,98 @@ def create_component_version(
...
+            # we add our key->raw_content_pk mapping to the next version.
+            if raw_content_pk is not None:
+                ComponentVersionRawContent.objects.create(
+                    raw_content_id=raw_content_pk,
+                    component_version=component_version,
+                    key=key,
+                    learner_downloadable=False,
+                )
+        # Now copy any old associations that existed, as long as they don't
+        # conflict with the new stuff.
```
kdmccormick, line 144:
```suggestion
        # Now copy any old associations that existed, as long as they aren't
        # in conflict with the new stuff or marked for deletion.
```
non-blocking suggestion -- this might add a bit more clarity

openedx_learning/core/components/api.py
```diff
@@ -73,15 +73,98 @@ def create_component_version(
     return component_version
 
 
+def create_next_version(
+    component_pk: int,
+    title: str,
+    content_to_replace: dict[str, int | None],
```
kdmccormick, line 79:
```suggestion
    content_to_replace: dict[str, int],
    content_to_remove: set[str],
```
non-blocking--

`None` is acting a sentinel value in `content_to_replace`, and I tend to be wary of sentinels, especially when they feel like special cases both on the caller's side and the implementer's side. In this case, I think I'd split into a dict for replacements and a set for deletions.

That said, I think you have a better idea of the calling patterns here, so maybe this will be more elegant to callers than I'm imagining. Regardless, it's a minor thing, and you've explained it very clearly in the docstring, so no sweat if you decide to keep it as-is.

kdmccormick (commented):
I reviewed all the apis, but had to sign off before getting to the mixins and model managers. Looking great! No blockers from me.

openedx_learning/core/components/api.py
ormsbee: I think the big thing that using `None` has going for it is that it makes it harder to call with inconsistent arguments. Say someone passed in:

```python
content_to_replace = {'block.xml': 201, 'hello.png': 928}
content_to_delete = {'hello.png'}
```

If those are both sent in the same call, we have to either let one or the other directive take precedence, or raise an error with the inconsistency, and callers have to be aware of that. In the current implementation, the semantics of a dict make it so that you can't represent that particular inconsistency.

openedx_learning/core/components/api.py
ormsbee: Changed, ty!

ormsbee:
> Okay, I wrote the tests, made changes based on the review comments, and the linters are appeased. I'm going to do the _pk refactoring now, but with any luck I'm hoping this can merge today.

I went back and forth on `_id` vs. `_pk` for a bit and then decided to punt entirely by making these arguments positional-only for the moment. >_<

ormsbee:
Also, I tracked down the query issue to something in edx-platform, and I'll have an update for that PR with the fix later. Part of it points to an issue that our internal APIs have with QuerySets and pagination. Namely, the pattern for DRF and Django in general is QuerySet -> Pagination -> Serialization. So it gets awkward when we're returning something from an API that we want to be QuerySet-like in that it's filterable and pageable, but also presents something that is not actually a model. I hacked up something before that wrapped the `__iter()__` call and did serialization that seemed to work, but I feel like that's probably too much hackery to want to support in the longer term.

Anyhow, I'll discuss it more on the other PR.

ormsbee:
It occurred to me while looking at some recent comments in the wiki w.r.t. library content needs that we probably want to have `ComponentType` represented as its own model. Partly for saving space, but mostly because it gives us something to later hang policy off of on a per-type basis, e.g. "videos are okay to show in libraries, SGA blocks are experimental for library use".

openedx_learning/core/components/api.py
kdmccormick: I hadn't thought of that. Sounds good to me!

ormsbee:
BTW, the only reason this build isn't green is because I was a doofus and mis-spelled a couple of conventional commit messages. Please know that it's ready for final review, and I'll squash it at the end.

openedx_learning/core/publishing/api.py
bradenmacdonald: Darn, I did like that alternative. Too bad the linter didn't.

bradenmacdonald (approved):
These are nice API changes and the code looks good to me!

I didn't run/test the code - let me know if you'd like me to.

bradenmacdonald:
Question that's not directly related: Can you use `add_content_to_component_version` to modify a `ComponentVersion` that has already been published?

kdmccormick (approved):

ormsbee:
Thanks folks! I squashed my commits and made a decent commit message out of it. Will merge once things are green.

ormsbee:
> Question that's not directly related: Can you use `add_content_to_component_version` to modify a `ComponentVersion` that has already been published?

Yes, you can. 95% of the time you shouldn't, but some real use cases came up where doing so seems like it would be beneficial:

* **Bug fixes**: We expect people to attached data to ComponentVersions, but that code will have bugs at some point, and people will want to be able to correct that programmatically without creating a new "version" of that thing.
* **Data upgrade/migration**: Say we have tens of thousands of Videos with SRT files, and we want to bulk upgrade everything to VTT files instead. We'd probably update the player to recognize both (with a preference for VTT), and then bulk convert SRT to VTT where VTT is absent.

PR #148, opened 2024-01-31
refactor: normalize component types with ComponentType

ormsbee:
Prior to this commit, Components encoded their namespace + type info as
two columns on the Component model itself. This wasted a lot of space on
a very common index lookup–many rows of ('xblock.v1', 'video'),
('xblock.v1', 'problem'), and ('xblock.v1', 'text'). But worse, the lack
of a separate ComponentType model meant that there was no first class
entity to store per-type policy data against. For example, we might want
to say "the following types are supported by libraries, while these
other types are experimental", or "these types are enabled for these
orgs", etc.

Components are required to have a ComponentType. We're rewriting the
first migration for the components app here, since this app hasn't been
added to edx-platform yet.


ormsbee:
Supersedes #147

openedx_learning/core/components/models.py
```diff
@@ -28,6 +28,45 @@
...
+    # namespace and name work together to help figure out what Component needs
+    # to handle this data. A namespace is *required*. The namespace for XBlocks
+    # is "xblock.v1" (to match the setup.py entrypoint naming scheme).
+    namespace = case_sensitive_char_field(max_length=100, blank=False)
+
+    # name is a way to help sub-divide namespace if that's convenient. This
+    # field cannot be null, but it can be blank if it's not necessary. For an
+    # XBlock, this corresponds to tag, e.g. "video". It's also the block_type in
+    # the UsageKey.
+    name = case_sensitive_char_field(max_length=100, blank=True)
```
kdmccormick, line 54: Should we set `null=True`, so that we can't have multiple ComponentTypes with the same namespace and and empty names?

openedx_learning/core/components/api.py
```diff
@@ -18,32 +18,44 @@
...
 from django.db.transaction import atomic
 
+from ...lib.cache import lru_cache
 from ..publishing import api as publishing_api
-from .models import Component, ComponentVersion, ComponentVersionRawContent
+from .models import Component, ComponentType, ComponentVersion, ComponentVersionRawContent
+
+
+@lru_cache(maxsize=128)
+def get_or_create_component_type_id(namespace: str, name: str) -> int:
```
kdmccormick, line 27: Even though it returns a pk, I think I would mildly prefer `get_or_create_component_type`, since you're not going to create a "compoment type", not a "component type id".

kdmccormick (approved):
Just a couple Qs, otherwise looks great.

openedx_learning/core/components/models.py
ormsbee: I'm not sure if I understood the question correctly. The way the code is now (with `null=False`), the uniqueness constraint on `(namespace, name)` means that there can't be multiple entries for the same namespace with blank names. But turning `null=True` would allow that to happen, since nulls don't evaluate as equal to each other, so the uniqueness constraint doesn't help.

kdmccormick: Sorry, I misread https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.Field.null. You're right--this looks good.

openedx_learning/core/components/api.py
kdmccormick: Ah. Would the rolled-back id be mis-cached in the `lru_cache`, or somewhere else?

ormsbee: Yeah, in `lru_cache`.

bradenmacdonald: I don't expect that _creating_ `ComponentType` instances will occur outside of `create_component`. So I suggest that you just put the `ComponentType.objects.get_or_create(` call inline (without a cache) inside `create_component`. Then if it gets rolled back, no problem.

If this gets simplified to `@lru_cache def get_component_type_id()` then you have two nice simplifications:
1. If the `ComponentType` is not found, it won't be cached, due to the `ComponentType.NotFound` exception
2. You can optimize lookups like this:
    ```python
     Component.with_publishing_relations \
                        .get(
                            component_type__namespace=namespace,
                            component_type__name=type_name,
                            local_key=local_key,
                        )
    ```

    becomes

    ```python
     Component.with_publishing_relations \
                        .get(
                            component_type_id=get_component_type_id(namespace, type_name),
                            local_key=local_key,
                        )
    ```
which will then cause your function to throw differentiated exceptions: `Component.NotFound` OR `ComponentType.NotFound` vs. only throwing `Component.NotFound`. Arguably more explicit.

kdmccormick: ^ Nice, I like that.

ormsbee: I think that I'm going to just kick the `lru` to the curb entirely for now. The worry I have is that the `lru` bleeds across transactions, and we don't know what combination of reads and writes of Components people are going to do in a big transaction (e.g. import). So it's possible that a process will add a bunch of stuff (including new ContentTypes), then do some reads of Components because they're grouping them somehow (thus loading things into the LRU cache). Then their process dies, and gets rolled back.

I guess the semantics I'm really looking for are transaction-local caches. Which actually sounds really cool now that I think about it... but I'm definitely not going there this week. A much simpler version of that would be a local object that proxies the requests through and goes out of scope at the end of whatever import operation happens. But things will probably be fine if I just punt on caching for a while.

PR #149, opened 2024-02-01
Combine RawContent and TextContent into Content

ormsbee:
This is mostly pulling together `RawContent` and `TextContent` into a unified `Content` model in order to reduce confusion and not force text to always go to a file-based storage backend.

Other things that may also be a part of this PR as issues I've noticed along the way:

* reduced the primary key for LearningPackage to 4-bytes to reduce the size of indexes.
* maybe eliminate the `FileField` to save space (and use the low level storages API instead)
* fix a potential cache corruption issue with `lru` in rollback situations.

### DIFF
--- a/openedx_learning/core/components/api.py
+++ b/openedx_learning/core/components/api.py
@@ -285,17 +285,19 @@ def get_components(
         qset = qset.filter(component_type__name__in=type_names)
     if draft_title is not None:
         qset = qset.filter(
-            publishable_entity__draft__version__title__icontains=draft_title
+            Q(publishable_entity__draft__version__title__icontains=draft_title) |
+            Q(local_key__icontains=draft_title)
         )
     if published_title is not None:
         qset = qset.filter(
-            publishable_entity__published__version__title__icontains=published_title
+            Q(publishable_entity__published__version__title__icontains=published_title) |
+            Q(local_key__icontains=published_title)
         )
 
     return qset
 
 
-def get_component_version_content(
+def look_up_component_version_content(
     learning_package_key: str,
     component_key: str,
     version_num: int,
@@ -323,10 +325,10 @@ def get_component_version_content(
                                   ).get(queries)
 
 
-def add_content_to_component_version(
+def create_component_version_content(
     component_version_id: int,
-    /,
     content_id: int,
+    /,
     key: str,
     learner_downloadable=False,
 ) -> ComponentVersionContent:
--- a/openedx_learning/core/contents/api.py
+++ b/openedx_learning/core/contents/api.py
@@ -64,8 +64,8 @@ def get_content(content_id: int, /) -> Content:
 
 def get_or_create_text_content(
     learning_package_id: int,
-    /,
     media_type_id: int,
+    /,
     text: str,
     created: datetime,
     create_file: bool = False,
@@ -86,10 +86,6 @@ def get_or_create_text_content(
     If you want to create a large text file, or want to create a text file that
     doesn't need to be stored in the database, call ``create_file_content``
     instead of this function.
-
-    The `hash_digest` is included as an optional argument because a very common
-    pattern is going to be "I have this data, let's see if a Content already
-    exists for it."
     """
     text_as_bytes = text.encode('utf-8')
     hash_digest = create_hash_digest(text_as_bytes)
@@ -122,8 +118,8 @@ def get_or_create_text_content(
 
 def get_or_create_file_content(
     learning_package_id: int,
-    /,
     media_type_id: int,
+    /,
     data: bytes,
     created: datetime,
 ) -> Content:
@@ -131,9 +127,10 @@ def get_or_create_file_content(
     Get or create a Content with data stored in a file storage backend.
 
     Use this function to store non-text data, large data, or data where low
-    latency access is not necessary. Also use this function to store any Content
-    that you want to be downloadable by browsers in the LMS since the static
-    asset serving system will only work with file-backed Content.
+    latency access is not necessary. Also use this function (or
+    ``get_or_create_text_content`` with ``create_file=True``) to store any
+    Content that you want to be downloadable by browsers in the LMS, since the
+    static asset serving system will only work with file-backed Content.
     """
     hash_digest = create_hash_digest(data)
     with atomic():
--- a/openedx_learning/core/contents/models.py
+++ b/openedx_learning/core/contents/models.py
@@ -18,7 +18,7 @@ from ...lib.managers import WithRelationsManager
 from ..publishing.models import LearningPackage
 
 
-def get_storage_class() -> Storage:
+def get_storage() -> Storage:
     """
     Return the Storage instance for our Content file persistence.
 
@@ -227,7 +227,7 @@ class Content(models.Model):
     # This hash value may be calculated using create_hash_digest from the
     # openedx.lib.fields module. When storing text, we hash the UTF-8
     # encoding of that text value, regardless of whether we also write it to a
-    # file or not.
+    # file or not. When storing just a file, we hash the bytes in the file.
     hash_digest = hash_field()
 
     # Do we have file data stored for this Content in our file storage backend?
@@ -263,17 +263,28 @@ class Content(models.Model):
     created = manual_date_time_field()
 
     @cached_property
-    def mime_type(self):
+    def mime_type(self) -> str:
+        """
+        The IANA media type (a.k.a. MIME type) of the Content, in string form.
+
+        MIME types reference:
+          https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types
+        """
         return str(self.media_type)
 
     def file_path(self):
+        """
+        Path at which this content is stored (or would be stored).
+
+        This path is relative to configured storage root.
+        """
         return f"{self.learning_package.uuid}/{self.hash_digest}"
 
-    def write_file(self, file: File):
+    def write_file(self, file: File) -> None:
         """
         Write file contents to the file storage backend.
         """
-        storage = get_storage_class()
+        storage = get_storage()
         file_path = self.file_path()
 
         # There are two reasons why a file might already exist even if the the
@@ -290,11 +301,11 @@ class Content(models.Model):
         if not storage.exists(file_path):
             storage.save(file_path, file)
 
-    def file_url(self):
+    def file_url(self) -> str:
         """
         This will sometimes be a time-limited signed URL.
         """
-        return get_storage_class().url(self.file_path())
+        return get_storage().url(self.file_path())
 
     def clean(self):
         """
--- a/openedx_learning/core/publishing/models.py
+++ b/openedx_learning/core/publishing/models.py
@@ -30,10 +30,11 @@ class LearningPackage(models.Model):  # type: ignore[django-manager-missing]
 
     Each PublishableEntity belongs to exactly one LearningPackage.
     """
+    # Explictly declare a 4-byte ID instead of using the app-default 8-byte ID.
     # We do not expect to have more than 2 billion LearningPackages on a given
-    # site, but many, many things have foreign keys to this model and uniqueness
-    # indexes on those foreign keys + their own fields, so going from the
-    # app-default of 8-byte IDs to 4-byte IDs will add up over time.
+    # site. Furthermore, many, many things have foreign keys to this model and
+    # uniqueness indexes on those foreign keys + their own fields, so the 4
+    # bytes saved will add up over time.
     id = models.AutoField(primary_key=True)
 
     uuid = immutable_uuid_field()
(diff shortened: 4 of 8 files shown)


ormsbee:
Okay, thinking on this a little more–the FileField is pretty much entirely redundant. The storage location is based entirely on the LearningPackage UUID + the `Content` hash digest. I'm going to see if it's cumbersome to make it a boolean flag + accessor method.

ormsbee:
Other self notes before I forget:

* The text should be null, not blank, when the data is only in the file.
* In the future, we might want to allow a blob field in `Content` for compressed text (e.g. using `zlib` compression). This could save us a lot of space, at the expense of making it more difficult to query. MySQL has a compressed row type sounds perfect for the usage pattern `Content` has, but we can't use it in RDS Aurora.

ormsbee:
The compression thing came to mind because I was looking through some example course data and there are a handful of Capa problems that weigh in at ~13-14 KB, which is higher than the 10K limit we currently place on `Content` text length. But when compressed with `zlib`, that goes down to about 2K–the larger problems tend to be that way because they have a lot of Python code and HTML table markup, both of which compress really well.

There are HTML blocks that are dramatically larger than this, but that's because they're encoding images into the raw HTML using base64-encoded data URLs (`"`). We are definitely not supporting that.

ormsbee:
Okay, I poked into the compression thing just a little bit further. I'm going to stop now because it's not critical to get in for the short term, but I have a general plan for it:

1. Rename the `text` field to `uncompressed_text`.
2. Create a new `BinaryField` for `compressed_text`.
3. Create a cached property `text` that knows how to switch between the two.

At the time of write, we run zlib compression on the text and decide whether to use the compressed or uncompressed field for this row. The other field is left null. When we first introduce this feature, we can run it as a data migration, though that wouldn't be a requirement.

Pruning is still the more important feature for controlling the content size growth.

ormsbee:
@bradenmacdonald, @kdmccormick: A little later than I had hoped, but it's ready for real review.

ormsbee:
@bradenmacdonald, @kdmccormick: Do you folks know what this mypy error is about by any chance?

> openedx_learning/core/contents/api.py:152: error: Incompatible type for "text" of "Content" (got "None", expected "Union[str, Combinable]")  [misc]

kdmccormick:
@ormsbee It's telling you that when constructing a `Content` model instance, you can't set `text=None`; it wants you to set it to an actual str instance. Would it make sense to do `text=""` here, or is that fundamentally different?

You can ignore the `Combinable` thing--that's like `F` and `Q` objects.

ormsbee:
> @ormsbee It's telling you that when constructing a `Content` model instance, you can't set `text=None`; it wants you to set it to an actual str instance. Would it make sense to do `text=""` here, or is that fundamentally different?

It's fundamentally different in this case. `text=None` means "there is no text representation in the database for this Content–it only exists in the file store". `text=""` means "there is a text representation in the database for this Content, and that happens to be an empty string".

But it's a nullable field, so it should be permitted. Does the type-checker just not accept nullable text fields?

kdmccormick:
> But it's a nullable field, so it should be permitted. Does the type-checker just not accept nullable text fields?

That would surprise me, so I tested it out by changing the definition of `text` to a regular TextField:

```python
#    text = MultiCollationTextField(
    text = models.TextField(
        blank=True,
        null=True,
        max_length=MAX_TEXT_LENGTH,
        # We don't really expect to ever sort by the text column, but we may
        # want to do case-insensitive searches, so it's useful to have a case
        # and accent insensitive collation.
#        db_collations={
#            "sqlite": "NOCASE",
#            "mysql": "utf8mb4_unicode_ci",
#        }
    )
```

and that type-checked fine (except for a new error that pops up on openedx_learning/core/components/admin.py:167, where I think you need to change `content_obj.text` to `context_obj.text or ""`).

So, I think the issue that the field-value type argument (specifically, `str|None`) isn't getting passed through MultiCollationTextField into TextField for some reason. It's possible that we need to define MultiCollationTextField as a generic class, although I'll need to play with that a bit to figure out the right syntax.

If that ends up blocking this PR, you could hack around the error for now by adding a type annotation directly to `text`:

```python
    # TextField type args are [TypeForGetting,TypeForSetting]
    text: models.TextField[str|None, str|None] = MultiCollationTextField(
        blank=True,
        null=True,
        max_length=MAX_TEXT_LENGTH,
        # We don't really expect to ever sort by the text column, but we may
        # want to do case-insensitive searches, so it's useful to have a case
        # and accent insensitive collation.
        db_collations={
            "sqlite": "NOCASE",
            "mysql": "utf8mb4_unicode_ci",
        }
    )

**Comment by kdmccormick** [comment]
@ormsbee if you want to just work around this for now using `text: models.TextField[str|None, str|None]`, I opened https://github.com/openedx/openedx-learning/issues/152, which the type-checking nerd in me would be happy to work through at some later date.

**Comment by ormsbee** [comment]
@kdmccormick: I respect and appreciate your inner type-checking nerd. I'll use the workaround you suggested. Thank you.

File: openedx_learning/core/contents/api.py [hunk 1]
```diff
@@ -80,69 +40,122 @@ def get_or_create_media_type_id(mime_type: str) -> int:
...
+    return Content.objects.get(id=content_id)
 
 
-def get_or_create_raw_content(
+def get_or_create_text_content(
     learning_package_id: int,
     /,
-    data_bytes: bytes,
-    mime_type: str,
+    media_type_id: int,
```
kdmccormick, line 68: Consider putting `media_type_id` above the `/` to avoid committing to `_id` in the kwarg name.

openedx_learning/core/contents/api.py
```diff
@@ -80,69 +40,122 @@ def get_or_create_media_type_id(mime_type: str) -> int:
...
+    that text data to the file storage backend. This is useful if we want to let
+    that file be downloadable by browsers in the LMS at some point.
+
+    If you want to create a large text file, or want to create a text file that
+    doesn't need to be stored in the database, call ``create_file_content``
+    instead of this function.
+
+    The `hash_digest` is included as an optional argument because a very common
+    pattern is going to be "I have this data, let's see if a Content already
+    exists for it."
```
kdmccormick, line 92:
```suggestion
```

openedx_learning/core/contents/api.py
```diff
@@ -80,69 +40,122 @@ def get_or_create_media_type_id(mime_type: str) -> int:
...
+
+        return content
+
+
+def get_or_create_file_content(
     learning_package_id: int,
     /,
-    data_bytes: bytes,
-    mime_type: str,
+    media_type_id: int,
```
kdmccormick, line 126: Same comment about moving `media_type_id` above `/`

openedx_learning/core/contents/api.py
```diff
@@ -80,69 +40,122 @@ def get_or_create_media_type_id(mime_type: str) -> int:
...
+) -> Content:
     """
-    Get the TextContent in the given learning package with the specified data,
-    or create it if it doesn't exist.
+    Get or create a Content with data stored in a file storage backend.
+
+    Use this function to store non-text data, large data, or data where low
+    latency access is not necessary. Also use this function to store any Content
+    that you want to be downloadable by browsers in the LMS since the static
+    asset serving system will only work with file-backed Content.
```
kdmccormick, line 136: suggested clarification, if you agree that this is the case:
```suggestion
    latency access is not necessary. Also use this function (or get_or_create_text_content
    with create_file=True) to store any Content that you want to be downloadable by
    browsers in the LMS since the static asset serving system will only work with file-backed Content.
```

kdmccormick (commented):
The general shape of the refactoring looks great.

I'm about 2/3rds through; I'll leave the rest of my review after lunch.

openedx_learning/core/contents/models.py
```diff
@@ -95,92 +105,211 @@ def __str__(self):
...
     )
 
+    # This should be manually set so that multiple Content rows being set in
+    # the same transaction are created with the same timestamp. The timestamp
+    # should be UTC.
+    created = manual_date_time_field()
+
     @cached_property
     def mime_type(self):
         return str(self.media_type)
```
kdmccormick, line 267: We should have a type hint here. Also, optionally, here's a docstring you could add.

```suggestion
    @cached_property
    """
    The IANA media type (a.k.a. MIME type) of the content stored in `text` and/or `file`.
    
    MIME types reference:
    https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types
    """
    def mime_type(self) -> str:
        return str(self.media_type)
```

openedx_learning/core/contents/models.py
```diff
@@ -95,92 +105,211 @@ def __str__(self):
...
 
+    def file_path(self):
+        return f"{self.learning_package.uuid}/{self.hash_digest}"
+
+    def write_file(self, file: File):
```
kdmccormick, line 272:
```suggestion
    def write_file(self, file: File) -> None:
```

openedx_learning/core/contents/models.py
```diff
@@ -95,92 +105,211 @@ def __str__(self):
...
+    # the same transaction are created with the same timestamp. The timestamp
+    # should be UTC.
+    created = manual_date_time_field()
+
     @cached_property
     def mime_type(self):
         return str(self.media_type)
 
+    def file_path(self):
+        return f"{self.learning_package.uuid}/{self.hash_digest}"
```
kdmccormick, line 270: type hint & docstring (optional)
```suggestion
    def file_path(self) -> str:
        """
        Path at which this content is stored (or would be stored), relative to configured storage root.
        """
        return f"{self.learning_package.uuid}/{self.hash_digest}"
```

openedx_learning/core/contents/models.py
```diff
@@ -95,92 +105,211 @@ def __str__(self):
...
+        # sort of transaction semantics, so it won't get rolled back.
+        #
+        # 2. The Content is of a different MediaType. The same exact bytes can
+        # be two logically separate Content entries if they are different file
+        # types. This lets other models add data to Content via 1:1 relations by
+        # ContentType (e.g. all SRT files). This is definitely an edge case.
+        if not storage.exists(file_path):
+            storage.save(file_path, file)
+
+    def file_url(self):
```
kdmccormick, line 293:
```suggestion
    def file_url(self) -> str:
```

openedx_learning/core/publishing/models.py
```diff
@@ -30,9 +30,20 @@ class LearningPackage(models.Model):  # type: ignore[django-manager-missing]
 
     Each PublishableEntity belongs to exactly one LearningPackage.
     """
+    # We do not expect to have more than 2 billion LearningPackages on a given
+    # site, but many, many things have foreign keys to this model and uniqueness
+    # indexes on those foreign keys + their own fields, so going from the
+    # app-default of 8-byte IDs to 4-byte IDs will add up over time.
+    id = models.AutoField(primary_key=True)
```
kdmccormick, line 37: I think this is saying that we're using 4 bytes instead of 8 for this model's ID field, but I wasn't entirely sure, so here's the comment reworded in a way that'd make me sure. If I'm wrong please correct me 😄 
```suggestion
    # Explictly declare a 4-byte ID instead of using the app-default 8-byte ID.
    # We do not expect to have more than 2 billion LearningPackages on a given
    # site. Furthermore, many, many things have foreign keys to this model and
    # uniqueness indexes on those foreign keys + their own fields, so the 4 bytes
    # saved will add up over time.
    id = models.AutoField(primary_key=True)
```

kdmccormick (commented):
Alrighty, all I have is a bunch of docstring nits.

I'm sure my next review will be a ✅ , so feel free to merge if someone beats me to it.

openedx_learning/core/contents/api.py
ormsbee: <sigh> Thank you for catching this. I went back and forth on it a few times before killing it off again.

openedx_learning/core/components/api.py
```diff
@@ -297,43 +300,42 @@ def get_component_version_content(
     component_key: str,
     version_num: int,
     key: Path,
-) -> ComponentVersionRawContent:
+) -> ComponentVersionContent:
```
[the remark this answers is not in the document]
ormsbee: This was something that I made back when I took the a first pass at the `media_server` app and I needed something to look up the static assets based on all the path info that would come in via a URL. The query was annoying enough to write that I kept it around even when it was clear that the `media_server` as implemented was not going to feasible as our long term solution.

I like the `look_up_*` convention. And I don't feel shy about violently refactoring the API while we're going through the process of it's pre-1.0 libraries implementation trial-by-fire. I'll make this change.

openedx_learning/core/components/api.py
```diff
@@ -18,44 +18,48 @@
...
-    return component_type.id
+    return component_type
 
 
 def create_component(
     learning_package_id: int,
     /,
-    namespace: str,
-    type_name: str,
+    component_type: ComponentType,
```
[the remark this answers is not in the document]
ormsbee: I'd rather keep this as is for now. The main reason I made `create_component` take a `ComponentType` instead of an ID or the namespace/name combo was so that `create_component` wouldn't have to query for `ComponentType` information every time. I figured that we often create many Components at once, like we would for an import, and that the caller can take the hit for pulling the ComponentType once per type.

But the `get_component_by_key` and `component_exists_by_key` is fetching one component, usually as a lookup for rendering a request for a specific block (since we have much more efficient ways of iterating through lists of them). So I thought it'd be easier for them to just pass through the arguments and let the API worry about fetching them in one query, and not have the caller think about `ComponentTypes`.

I realize that there are a few different ways we could add caching to the mix to make this better, but they all carry complications that I didn't think were worth it just yet (e.g. request caching adds a major dependency and can have edge cases around celery tasks, `lru` has the rollback invalidation edge case, etc.).

I'd probably just make a function for both if Python had better support for it. (I mean, I know `singledispatch` exists, but I've never seen anyone actually use it.)

ormsbee:
@kdmccormick: Incorporated all your suggestions except [this one](https://github.com/openedx/openedx-learning/pull/149#discussion_r1483856944) on `get_component_by_key/component_exists_by_key`.

openedx_learning/core/components/api.py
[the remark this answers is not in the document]
kdmccormick: Makes sense to me 👍🏻

kdmccormick (approved):
🚀

PR #156, opened 2024-02-07
Add total descendant count to tag list APIs [FC-0036]

bradenmacdonald:
This PR updates the APIs for retrieving the tags in the taxonomy, so that they state not only how many direct children they have, but also how many total descendants.

The direct children count is still included, not only to preserve backwards compatibility, but because it's necessary to compute pagination and display a "load more children" button when necessary.

What about **performance** of these queries? I ran the whole test suite before and after this change, which includes some very large taxonomies, and there was no measurable difference on average.

Ref: https://github.com/openedx/modular-learning/issues/191

Private ref: [FAL-3655](https://tasks.opencraft.com/browse/FAL-3655)

Note: needs a version bump when merged.


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

openedx_tagging/core/tagging/models/base.py
```diff
@@ -170,6 +170,18 @@ def child_count(self) -> int:
...
+    @cached_property
+    def descendant_count(self) -> int:
+        """
+        How many descendant tags this tag has in the taxonomy.
+        """
+        if self.taxonomy and not self.taxonomy.allow_free_text:
+            return self.taxonomy.tag_set.filter(
+                Q(parent__parent=self) |
+                Q(parent__parent__parent=self)
+            ).count() + self.child_count
```
pomegranited, line 182: Can we combine these two calls together?

```suggestion
            return self.taxonomy.tag_set.filter(
                Q(parent=self) |
                Q(parent__parent=self) |
                Q(parent__parent__parent=self)
            ).count()
```

pomegranited (approved):
:+1: with one suggested change.

Once you address that and do a version bump, I think this is good to merge.

- [x] I tested this on my devstack using https://github.com/openedx/frontend-app-course-authoring/pull/826
    * Navigated to the taxonomies list and selected a taxonomy
    * Ensured that the descendant counts are showing as expected next to each tag in the tree.
- [x] I read through the code
- [x] I checked for accessibility issues by using my keyboard to navigate
- [x] Includes documentation
- [x] <strike>User-facing strings are extracted for translation</strike> N/A

openedx_tagging/core/tagging/models/base.py
bradenmacdonald: We could, but if `child_count` is already cached (because it's a `@cached_property`), that does one set of extra database lookups that we don't need. I guess it's not a big deal if you think the consistency here is more important?

pomegranited: Ah good point, and since we have to return the `child_count` too, it's better to leave this as-is. Suggestion retracted :)

PR #160, opened 2024-02-14
Rename "key" fields

ormsbee:
```
refactor: rename fields from "key" to "_key"

We're doing this because "key" is a reserved keyword for MySQL. While
Django will escape this for us automatically, we've historically seen
downstream tooling break when column names use reserved keywords. Also,
it makes things harder for people who are querying the database directly
and might not realize that they need to esacape column names.

In any event, this has caused enough pain in the past to warrant a check
in edx-platform's CI, where it checks column names against a whole list
of reserved keywords for several systems:

https://github.com/openedx/edx-django-release-util/blob/master/release_util/management/commands/default_reserved_keywords.yml

It's likely that we'll want rename at least some of these columns later,
but this lets us introduce them into the system without worrying that
they'll break anything.
```


bradenmacdonald (approved):

ormsbee:
(sigh). Forgot for the hundredth time that django generates migrations with imports in an order that isort doesn't like.

PR #175, opened 2024-03-26
Upgrade DRF

farhan:
Description:
Bump DRF to `v3.15.1`

References:
- https://docs.djangoproject.com/en/4.2/ref/models/fields/#enumeration-types
- https://github.com/encode/django-rest-framework/compare/3.14.0...3.15.1#diff-7768b980611e24de782acc70ad64cd57b9d169d42f80b200cb7c6ef277fd74dbR1440
- https://www.django-rest-framework.org/community/3.15-announcement/#django-rest-framework-315


openedx_tagging/core/tagging/models/import_export.py
```diff
@@ -12,15 +11,15 @@
 from .base import Taxonomy
 
 
-class TagImportTaskState(Enum):
+class TagImportTaskState(models.TextChoices):
```
awais786, line 14: https://docs.djangoproject.com/en/5.0/ref/models/fields/#enumeration-types

openedx_tagging/core/tagging/models/import_export.py
```diff
@@ -42,7 +41,7 @@ class TagImportTask(models.Model):
 
     status = models.CharField(
         max_length=20,
-        choices=[(status, status.value) for status in TagImportTaskState],
+        choices=TagImportTaskState.choices,
```
irtazaakram, line 44: This change should create new migrations.

irtazaakram (approved):

feanil (changes requested):
I don't think we want to switch this from an enum to text choices, especially not translated, any translation should be happening upstream of the database.  @ormsbee can you take a look?

feanil:
I think instead you need to update the serializer here: https://github.com/openedx/openedx-learning/blob/1333cf6421e35b7e621fd5e10fc91c6a6f373634/openedx_tagging/core/tagging/rest_api/v1/serializers.py#L349

Currently it's trying to use the default serializer but that won't work with a tuple so you might need to make a custom serializer for just that field.

ormsbee:
@ChrisChV, @pomegranited, @bradenmacdonald: Could you please take a look at this? I'm not clear on the implications.

bradenmacdonald:
@feanil 

> I don't think we want to switch this from an enum to text choices, especially not translated, any translation should be happening upstream of the database.

I think it's unimportant but OK to translate this - the translations will show up in the Django Admin (only). Though generally I'm not in the habit of translating Django admin stuff for Open edX.

> Currently it's trying to use the default serializer but that won't work with a tuple so you might need to make a custom serializer for just that field.

There's no tuple involved - it's still a `CharField` and before and after this change it should be just a simple string value like `"error"` or `"success"` that's stored in MySQL or returned from the REST API. (Though in the middle, before this change, the Python-land view of this value was a special enum value object, not a plain string.)

I don't think any changes to the serializer are needed?

I reviewed the diff here - looks good. And since our REST API tests like

https://github.com/openedx/openedx-learning/blob/2ba5cf4cb26b13ff395d7010a9d0399fd3091b31/tests/openedx_tagging/core/tagging/test_views.py#L2621

are passing unchanged, this is good with me.

bradenmacdonald (approved):

feanil (commented):
@bradenmacdonald thanks for taking a look, I missed that the initial model was also a Charfield mis-read the choices bit.  Thanks for the clarification.

feanil (approved):

PR #211, opened 2024-08-14
fix: prevent soft-deletes from being re-published every time

ormsbee:
Details are in a comment from this change:

```python
        # Account for soft-deletes:
        # NULL != NULL in SQL, so simply excluding entities where the Draft
        # and Published versions match will not catch the case where a
        # soft-delete has been published (i.e. both the Draft and Published
        # versions are NULL). We need to explicitly check for that case
        # instead, or else we will re-publish the same soft-deletes over
        # and over again.
```

### DIFF
--- a/openedx_learning/apps/authoring/publishing/api.py
+++ b/openedx_learning/apps/authoring/publishing/api.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 from datetime import datetime, timezone
 
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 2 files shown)


ormsbee:
@pomegranited or @navinkarkera: Review if one of you has time, please?

navinkarkera (approved):
@ormsbee Looks good. :+1:
- [x] I tested this: Published library with soft deleted components and verified that they are not included in the publish log records.
- [x] I read through the code
- [x] Includes documentation

PR #212, opened 2024-08-16
Remove Django 3.2 collations compatibility hack

ormsbee:
Setting the collation via Field db_parameters was not properly supported before Django 4.x. We worked around this with a Django 3.2-compatible hack in https://github.com/ormsbee/openedx-learning/commit/084ec21a0292584f42f79930b6d6fe3bcf0ef08b. Now that we require at least Django 4.2, we no longer need the compatibility hack.

### DIFF
--- a/openedx_learning/lib/collations.py
+++ b/openedx_learning/lib/collations.py
@@ -36,48 +36,6 @@ class MultiCollationMixin:
         super().__init__(*args, **kwargs)
         self.db_collations = db_collations or {}
 
-        # This is part of a hack to get this to work for Django < 4.1. Please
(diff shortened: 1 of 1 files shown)


kdmccormick (approved):

ormsbee:
It's stupid, but I'm irrationally fond of the hack that I'm removing here. I mean, the hack is absolutely terrible and we're definitely better off getting rid of it, but it worked surprisingly well. 😛

kdmccormick:
Yeah, OOP and getter/setter magic can drive me crazy but this is a good demonstration of when it's really helpful 😄

PR #215, opened 2024-08-19
refactor: create_next_version -> create_next_component_version

ormsbee:
Since the authoring apps now have their APIs aggregated into the public `openedx_learning.api.authoring` module, `create_next_version` will be ambiguous as soon as we create other versioned things like Units.

I'm including an alias to give more flexibility for switching edx-platform over.


kdmccormick (approved):
LGTM with the micro-nit that I'd call this a non-breaking `feat:` since it changes the public API

ormsbee:
@kdmccormick: Ah, good point. I'll change the commit message.

PR #217, opened 2024-08-22
Basic asset response support

ormsbee:
Low-level Learning Core API calls for returning references to content associated with a Component. Uses `X-Accel-Redirect`.

```
This adds a number of API calls to the Authoring public API in order to
support associating static assets with Components and allowing them to
be downloaded.

Added:

* get_component_by_uuid
* get_component_version_by_uuid
* get_content_info_headers
* get_redirect_response_for_component_asset

Modified:

* create_next_component_version - made title optional
* create_component_version_content - annotation for learner_downloadable

Most of the justification for this approach can be found in the
docstring for get_redirect_response_for_component_asset.

Note that there is currently no backend redis/memcached caching of the
ComponentVersion and Content information. This means that every time a
request is made, we will do a couple of database queries in order to do
this lookup. I had actually done a version with such caching in an
earlier iteration of this PR, but it added a lot of complexity for what
I thought would be minimal gain, since going through the middleware will
cause about 20 database queries anyway.

This implements the first part of the static assets ADR for Learning
Core: https://github.com/openedx/openedx-learning/pull/110

Important notes:

* No view or auth is implemented. That is the responsibility of
  edx-platform. Learning Core only provides the storage and response
  generation.
* The responses generated will require the use of a reverse proxy that
  can handle the X-Accel-Redirect header. See ADR for details.
```

### DIFF
--- a/openedx_learning/apps/authoring/components/api.py
+++ b/openedx_learning/apps/authoring/components/api.py
@@ -13,11 +13,16 @@ are stored in this app.
 from __future__ import annotations
 
 from datetime import datetime
+from enum import StrEnum, auto
+from logging import getLogger
 from pathlib import Path
+from uuid import UUID
 
 from django.db.models import Q, QuerySet
 from django.db.transaction import atomic
+from django.http.response import HttpResponse, HttpResponseNotFound
 
+from ..contents import api as contents_api
 from ..publishing import api as publishing_api
 from .models import Component, ComponentType, ComponentVersion, ComponentVersionContent
 
@@ -34,10 +39,18 @@ __all__ = [
     "create_component_and_version",
     "get_component",
     "get_component_by_key",
+    "get_component_by_uuid",
+    "get_component_version_by_uuid",
     "component_exists_by_key",
     "get_components",
     "create_component_version_content",
+    "look_up_component_version_content",
+    "AssetError",
+    "get_redirect_response_for_component_asset",
 ]
+
+
+logger = getLogger()
 
 
 def get_or_create_component_type(namespace: str, name: str) -> ComponentType:
@@ -112,9 +125,9 @@ def create_component_version(
 def create_next_component_version(
     component_pk: int,
     /,
-    title: str,
     content_to_replace: dict[str, int | None],
     created: datetime,
+    title: str | None = None,
     created_by: int | None = None,
 ) -> ComponentVersion:
     """
@@ -150,8 +163,11 @@ def create_next_component_version(
     last_version = component.versioning.latest
     if last_version is None:
         next_version_num = 1
+        title = title or ""
     else:
         next_version_num = last_version.version_num + 1
+        if title is None:
+            title = last_version.title
 
     with atomic():
         publishable_entity_version = publishing_api.create_publishable_entity_version(
@@ -245,6 +261,14 @@ def get_component_by_key(
                         component_type__name=type_name,
                         local_key=local_key,
                     )
+
+
+def get_component_by_uuid(uuid: UUID) -> Component:
+    return Component.with_publishing_relations.get(publishable_entity__uuid=uuid)
+
+
+def get_component_version_by_uuid(uuid: UUID) -> ComponentVersion:
+    return ComponentVersion.objects.get(publishable_entity_version__uuid=uuid)
 
 
 def component_exists_by_key(
@@ -351,7 +375,7 @@ def create_component_version_content(
     content_id: int,
     /,
     key: str,
-    learner_downloadable=False,
+    learner_downloadable: bool = False,
 ) -> ComponentVersionContent:
     """
     Add a Content to the given ComponentVersion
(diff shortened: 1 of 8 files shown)


ormsbee:
@bradenmacdonald, @kdmccormick: This is an implementation of static asset serving on the Learning Core side of things. I'm still writing tests and I need to do cleanup on the management command side, but the API and model code should be more or less there. Please feel free to take an early pass at it to make sure it looks sane. Otherwise, I'll ping you folks when it's baked enough for real review.

ormsbee:
This PR also has a small fix for an unrelated issue with the `version_num` helper, which came up in pylint errors. Not sure how that got through before...

openedx_learning/apps/authoring/components/api.py
```diff
@@ -363,3 +387,171 @@ def create_component_version_content(
...
+    return {
+        # Component
+        "X-Learning-Core-Component-Key": component.publishable_entity.key,
+        "X-Learning-Core-Component-Uuid": component.uuid,
+        # Component Version
+        "X-Learning-Core-Component-Version-Uuid": component_version.uuid,
+        "X-Learning-Core-Component-Version-Num": component_version.version_num,
+        # Learning Package
+        "X-Learning-Core-Learning-Package-Key": learning_package.key,
+        "X-Learning-Core-Learning-Package-Uuid": learning_package.uuid,
```
bradenmacdonald, line 418: Tiny tiny nit: I know these are mostly for debugging, but it seems weird to me to call these `X-Learning-Core-...` instead of something like `X-OEX-...`. Are these Open edX components that just happen to be implemented in the learning core package, or is Learning Core an end user brand?

Even from a developer perspective, the name "Learning Core" is not helpful for letting people know where the corresponding implementation is located, because the python package and github repo are both called `openedx-learning` which is not obviously the same thing as `Learning-Core`.

openedx_learning/apps/authoring/components/api.py
```diff
@@ -363,3 +387,171 @@ def create_component_version_content(
...
+    an ``HttpResponse`` with a status code of ``200``.
+
+    If the specified asset does not exist for this ``ComponentVersion``, or if
+    the ``ComponentVersion`` itself does not exist, the response code will be
+    ``404``.
+
+    *This function does not do auth checking of any sort*–it will never return
+    a ``401`` or ``403`` response code. That is by design. Figuring out who is
+    making the request and whether they have permission to do so is the
+    responsiblity of whatever is calling this function.
```
bradenmacdonald, line 457: That said, the `learner_downloadable_only` does seem like a type of auth check to me.

openedx_learning/apps/authoring/components/management/commands/add_assets_to_component.py
```diff
@@ -0,0 +1,111 @@
+"""
+Management command to add files to a Component.
+
+This is mostly meant to be a debugging tool to let us to easily load some test
+asset data into the system.
```
bradenmacdonald, line 5: Nice! This will be very helpful for dev/debugging.

bradenmacdonald (approved):
I haven't tested this out but the code makes sense to me! Nice work.

openedx_learning/apps/authoring/components/api.py
ormsbee: Yeah, I feel a bit weird about that flag in general, but it's mostly there to try to keep people from shooting themselves in the foot when they're adding random things like LaTeX files and such–and to avoid issues we've had in the past around accidental downloads we've had in the past (e.g. `python-lib.zip`). So I thought it'd be better if the entire endpoint would make those assets unfindable/servable, and not just leave it up to the authz code.

openedx_learning/apps/authoring/components/api.py
ormsbee: I think something in middleware will force this to come out mixed case, i.e. `X-OEX-` will get turned into `X-Oex-`. Regardless, I'm fine with changing this. Maybe `X-Open-edX-Learning-Package-Key`?

kdmccormick: If there's no concern about length of the header name, then I'd say spell it out-- no room for ambiguity with `X-Open-edX-` :)

openedx_learning/apps/authoring/components/api.py
kdmccormick: I think it's a solid design choice. I don't feel strongly, but maybe just reword the comment to be clear about it:

```
    Other than checking the coarse-grained ``learner_downloadable_only`` flag,
    *this function does not do auth checking*–it will never return a ``401`` or ``403``
    response code. That is by design. Figuring out who is making the request and
    whether they have permission to do so is the responsiblity of whatever is calling
    this function.
```

kdmccormick (commented):
I didn't do an in-depth review, but based on what you've solidly specced out in https://github.com/openedx/openedx-learning/blob/main/docs/decisions/0015-serving-static-assets.rst, plus Braden's review, I say merge away 👍🏻

Issue #230, opened 2024-09-19
New App Checklist


ormsbee:
Need to write a doc of things so we don't forget. Should mention

- [ ] import linter
- [ ] model conventions
- [ ] private API conventions
- [ ] public API conventions
- [ ] preferred way to make tests
- [ ] places where we deviate from OEP-49 (particularly how we pass around querysets)
- [ ] philosophy on small apps that encapsulate discrete pieces of functionality
- [ ] philosophy on extensible models
- [ ] boundaries we want to enforce (e.g. leaving authorization out)

(other folks please feel free to add stuff)

PR #231, opened 2024-09-19
test: add collections app to import linter config

ormsbee:
All new apps should have an entry in the .importlinter config file so that our internal dependency tree stays clean, but we don't actually document that anywhere yet.

Made https://github.com/openedx/openedx-learning/issues/230 to track work around making a "new app" checklist doc to help with this sort of thing in the future.

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -46,6 +46,10 @@ layers=
     # without versioning information. These belong to a single Learning Package.
     openedx_learning.apps.authoring.contents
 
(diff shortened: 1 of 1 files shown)


ormsbee:
@pomegranited, @navinkarkera, @rpenido: Just wanted you folks to be aware that this import-linting exists (I completely forgot to mention it during my review of the collections app).

navinkarkera (approved):
@ormsbee Got it. Looks good :+1:

bradenmacdonald (approved):

PR #234, opened 2024-09-27
Asset handling updates (config, admin)

ormsbee:
I think this should be the last of the Learning Core changes needed to get minimal asset handling into Content Libraries.

The meat of this is:

```
feat!: require OPENEDX_LEARNING setting for Content storage

WARNING: This commit will break any file-backed Content entries you
currently have. Though you will likely only have them if you've used the
very recently created add_assets_to_component command.

This commit switches us away from using the default_storage and requires
the project to set an OPENEDX_LEARNING settings dictionary with a MEDIA
key like:

OPENEDX_LEARNING = {
    'MEDIA': {
        "BACKEND": "django.core.files.storage.FileSystemStorage",
        "OPTIONS": {
            "location": "/openedx/media-private/openedx-learning",
        }
    }
}

If no such setting is present, Content operations that require file
access will raise a django.core.exceptions.ImproperlyConfigured error.

In addition to that, Content files will be stored in a "content"
subdirectory of the specified media location. This is to allow us more
flexibility as we start to use file storage for other things like the
import/export process.

We need to have a separate space for Learning Core media files because
these files should NOT be directly accessible via the browser (as the
MEDIA_ROOT is). This is because of access policies and the fact that the
filenames will not be meaningful by themselves and must be translated by
app logic. For details, please see:

* docs/decisions/0015-serving-static-assets.rst
* openedx_learning/apps/authoring/contents/models.py
* openedx_learning/apps/authoring/components/models.py

Hiding these files in a new location also required changes to the Django
admin, which are included here. This commit also adds a little extra to
the admin to make it easier to map Component assets to actual files on
disk.

This commit also adds the following helpers to the Content model:

* read_file(): return opened Django File object for the Content.
* os_path(): get the full OS path to the stored file, if available.
* path(): get the logical path for this asset relative to our configured
  OPENEDX_LEARNING['MEDIA'] storage root.

This commit also auto-strips file paths starting with '/'. Allowing
absolute paths tripped me up multiple times as I was setting up my test
data, and I don't think there's any advantage to allowing
inconsistencies (e.g. "/static/foo.png" and "static/foo.png"). We now
force the relative form ("static/foo.png").```
```

Also includes the following performance tweak:

```
perf: reduce db calls for ComponentVersion after lookup by UUID

When you fetch a single ComponentVersion, you are often going to want to
pull data related to its Component, and possibly the LearningPackage it
belongs to. This commit adds a select_related call to eliminate the
extra roundtrips for this information.
```

This will not work without the `OPENEDX_LEARNING` setting set, which will be done by default in tutor dev once https://github.com/overhangio/tutor/pull/1124 lands.

Screenshots:

admin-content-preview
admin-component-asset

### DIFF
--- a/openedx_learning/apps/authoring/contents/models.py
+++ b/openedx_learning/apps/authoring/contents/models.py
@@ -7,11 +7,13 @@ from __future__ import annotations
 
 from functools import cache, cached_property
 
-from django.core.exceptions import ValidationError
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.core.files.base import File
-from django.core.files.storage import Storage, default_storage
+from django.core.files.storage import Storage
 from django.core.validators import MaxValueValidator
 from django.db import models
+from django.utils.module_loading import import_string
 
 from ....lib.fields import MultiCollationTextField, case_insensitive_char_field, hash_field, manual_date_time_field
 from ....lib.managers import WithRelationsManager
@@ -28,13 +30,25 @@ def get_storage() -> Storage:
     """
     Return the Storage instance for our Content file persistence.
 
-    For right now, we're still only storing inline text and not static assets in
-    production, so just return the default_storage. We're also going through a
-    transition between Django 3.2 -> 4.2, where storage configuration has moved.
+    This will first search for an OPENEDX_LEARNING config dictionary and return
+    a Storage subclass based on that configuration.
 
-    Make this work properly as part of adding support for static assets.
+    If there is no value for the OPENEDX_LEARNING setting, we return the default
+    MEDIA storage class. TODO: Should we make it just error instead?
     """
-    return default_storage
+    config_dict = getattr(settings, 'OPENEDX_LEARNING', {})
+
+    if 'MEDIA' in config_dict:
+        storage_cls = import_string(config_dict['MEDIA']['BACKEND'])
+        options = config_dict['MEDIA'].get('OPTIONS', {})
+        return storage_cls(**options)
+
+    raise ImproperlyConfigured(
+        "Cannot access file storage: Missing the OPENEDX_LEARNING['MEDIA'] "
+        "setting, which should have a storage BACKEND and OPTIONS values for "
+        "a Storage subclass. These files should be stored in a location that "
+        "is NOT publicly accessible to browsers (so not in the MEDIA_ROOT)."
+    )
 
 
 class MediaType(models.Model):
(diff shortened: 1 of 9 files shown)


ormsbee:
@kdmccormick, @bradenmacdonald: I still need to add a couple of small tests, but this is the basic learning core piece of it.

ormsbee:
Now updated with tests.

openedx_learning/apps/authoring/contents/models.py
```diff
@@ -284,17 +298,34 @@ def mime_type(self) -> str:
...
+        """
+        return f"content/{self.learning_package.uuid}/{self.hash_digest}"
+
+    def storage_path(self):
+        """
+        The full OS path for the underlying file for this Content.
+
+        This will not be supported by all Storage class types.
+        """
+        return get_storage().path(self.file_path())
```
bradenmacdonald, line 314: Just noting that I found these names slightly confusing - I expected "storage_path" would be a path that works on any storage including cloud storage like s3, whereas "file_path" would be the "real os filesystem path" - but it's actually the reverse.

Perhaps `path` or `logical_path` instead of `file_path`, and `real_path` instead of `storage_path` if it's not too late to change. Not a big deal though.

tests/openedx_learning/apps/authoring/contents/test_file_storage.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Test that the file path doesn't change.
+
+        If this test breaks, it means that we've changed the convention for
+        where we're storing the backing files for Content, which means we'll be
+        breaking backwards compatibility for everyone. Please be very careful if
+        you're updating this test.
+        """
+        content = self.html_content
+        assert content.file_path() == f"content/{content.learning_package.uuid}/{content.hash_digest}"
```
bradenmacdonald, line 54: I feel like it would be better to hard-code a UUID and hash digest into the string here, because I could certainly imagine the UUID (and to a lesser extent the hash digest) could be serialized in different ways, breaking backwards compatibility, but not breaking the test.

`545fc5c8-5f07-442f-9f69-69b936d82d51` vs `545fc5c85f07442f9f6969b936d82d51` vs `545FC5C8-5F07-442F-9F69-69B936D82D51` are all the same UUID.

Though I'm not sure if you have a way to create entities with a specific UUID during tests so maybe we can't.

bradenmacdonald (approved):
LGTM! A couple minor comments but nothing blocking.

I tested this as described [here](https://github.com/openedx/edx-platform/pull/35557#issuecomment-2381052162).

openedx_learning/apps/authoring/contents/models.py
ormsbee: Renamed this to `path`/`os_path`, which will hopefully be clearer.

tests/openedx_learning/apps/authoring/contents/test_file_storage.py
ormsbee: We don't right now... we probably should, if we want to accurately re-create from an import. But I don't want to add that to that to the scope of this PR.

Issue #240, opened 2024-10-09
ADR for Units and Containers


ormsbee:
We need to write an ADR (or multiple ADRs) describing Learning Core's approach to modeling Units in particular and Containers in general.

A lot of this is informed by discussions in:

* https://github.com/openedx/openedx-learning/issues/38
* https://github.com/openedx/openedx-learning/issues/219

ormsbee:
## High Level Requirements (non-exhaustive)

### Units

The first concrete use case for a more generalized Container capability is to model Units for Libraries and eventually Courses.

### Dynamic Content

We must support permutations of content in containers that is customized for a group of people or even an individual learner. Examples of this:

* Two groups: A/B testing two different sets of Components to see which performs better.
* Per student: Selecting three problems from a set of twenty, in random order.

### Extensible

Plugin authors must be able to create their own container types, as well as their own dynamic content types.

### Course Hierarchy

Courses currently have a hierarchy of Course → Section → Subsection → Unit, and we must be able to represent this in Learning Core. This representation does not have to be represented as nested Containers as it is today in ModuleStore, but we must have some plan for how these larger groupings will work.

### Fast Queries

A drawback of XBlock-based dynamic children has been poor performance. The system should be created so that queries to get the content for a given user are predictable and fast. While invoking plugin code may be unavoidable for assigning a user to a particular variant of dynamic content, the system should be able to read these assignments without invoking plugin code.

### Independent of XBlock

We must support current XBlocks that have dynamic children (LibraryContentBlock, SplitTestBlock, and RandomizeBlock), but nothing in Learning Core should require or assume XBlocks. Learning Core should hold the primitives that are needed to build these XBlocks on top of, but should be more broadly usable for Components that are not XBlocks.

### History

Containers should be able to differentiate between changes in the metadata of the container itself (e.g. title, ordering of contents, adding or removing content), vs. changes to the contents (e.g. a component in a Unit is updated with new text).

### Efficient Storage

The Split ModuleStore system for storing course content has the unfortunate property that even minor changes require large writes of the entire structure, resulting in a lot of wasted space. Changes to the content of a Container should ideally not require rewriting a new version of the Container.

### Scale

An individual Container must scale to at least 100 items, though ideally it should work for up to a 1000 (TBD how much of a requirement this is). The median size of Containers will be much smaller, say ~10, give or take.

ormsbee:
## Sketch of Models

These are the models that I've sketched out to try to capture:

[...]

Issue #219, opened 2024-08-30
How do we deal with a Unit's unpinned references to components when those components are soft-deleted?


ormsbee:
Is it worth adding more metadata to the `Published` table to cover use cases where something has been soft deleted (e.g.  `last_version_before_deletion`)? This came to mind when I was thinking about what happens if we use the "a null version pointer means take the latest draft or published" convention and then soft-delete+publish a Component that was being referenced. Do we always want to carry over the deletion, or do we sometimes just want to lock to the last valid version?

This information is derivable from `PublishLogRecord` (find where the entry for this `PublishableEntity` gets its `new_version` set to `null`). But that's slower to access for the "get a big list of components" scenarios.

(This came up tangentially in [this comment](https://github.com/openedx/openedx-learning/issues/38#issuecomment-2318925937) w.r.t. how Slots in Units might change over time, e.g. someone deletes a Component from a pool of possibilities to be randomly selected from.)

ormsbee:
FYI @kdmccormick, @bradenmacdonald

bradenmacdonald:
> This came to mind when I was thinking about what happens if we use the "a null version pointer means take the latest draft or published" convention and then soft-delete+publish a Component that was being referenced. Do we always want to carry over the deletion, or do we sometimes just want to lock to the last valid version?

Can I get some clarity on this?

Lib component Alpha version 1 is published
Lib component Alpha version 2 is published
That Alpha component is used in a course, with a null reference that means "take the latest published version", and version 2 is copied into the course.
Lib component Alpha version 3 is published, and then immediately (or perhaps some time later) is soft-deleted
Now when the course author goes to apply updates from the library, are you saying we should be able to pull in the soft deleted (but last published) version 3 as an update to replace the existing version 2 in the course? If so, I don't think that matters.

_If_ I'm understanding correctly, I don't think we need to work too hard to support this case. I think users would understand if they're now stuck on version 2 even though it wasn't the last published version, because the whole thing has been deleted upstream. And if it were really important to upgrade to version 3, the library author could un-delete it.

ormsbee:
@bradenmacdonald: I think I'm going to close this one as unnecessary anyway, but to give more clarity:

The "let the version float" convention we were talking about–where `UnitVersions` only store the `components` that are in them and leave the `component_versions` unpinned–works for controlling version explosion where every tiny change to every `Component` in a course cascades up with fully realized new versions. So the scenarios where new `UnitVersions` are actually created are:

[...]

Issue #38, opened 2023-04-02
Modeling Units and Versioning


ormsbee:
A Unit can have Components that are both fixed to a particular version (e.g. borrowed content from a Library), as well as references that should always point to the latest version of a Component (e.g. a Component in the same course). This pattern repeats itself at different scales (e.g. CCX courses), where sometimes we want to only update our version of borrowed content explicitly vs. always grabbing the latest published version.

We could model this sort of relationship explicitly in a Unit by making foreign key references to both the versioned and unversioned model and having a null value for the versioned field mean that we always grab the latest one via a join on PublishedComponent.

<strike>Or maybe we do always explicitly create a new version of a Unit whenever one of its child Components updates, and we keep a flag as to whether to auto-update or to lock to a specific version on a per-Component basis? We'd then hook into the publish workflow to publish the new version of the Unit along with the Component?</strike>

ormsbee:
> So we could model this sort of relationship explicitly in a Unit by making foreign key references to both the versioned and unversioned model and having a null value for the versioned field mean that we always grab the latest one via a join on PublishedComponent.

I'm leaning more towards this one. More specifically, something like:

```python
class UnitVersionComponentVersion(models.Model):
    unit_version = models.ForeignKey(UnitVersion, on_delete=models.CASCADE)
    component = models.ForeignKey(Component, on_delete=models.RESTRICT)
    component_version = models.ForeignKey(ComponentVersion, on_delete=models.RESTRICT, null=True)
    order_num = models.PositiveInteger(null=False)
```

This would mean that we *don't* usually create new UnitVersions when Components are updated–only when the Unit itself changes. That will reduce the noise a lot when we're talking about containers (Units, Subsections, Sections, etc.) within the same LearningPackage. At the same time, we can fix to specific versions when using external resources. But I think this also lets us do the CCX use case where we want to reference external things that are constantly updating.

In other words, in the past I think we've been talking about who controls updates–the people creating library-style content or the people using that content in their courses. But I think if we have this pattern for composition, we determine that policy independently for every container's contents.

ormsbee:
@kdmccormick, @bradenmacdonald, @feanil

ormsbee:
Wanted to capture some things that were discussed in an in-person whiteboarding session:

### Borrowing a Unit via Cloning

[...]

Issue #154, opened 2024-02-07
Prune unused content


ormsbee:
The `PublishLog` and having LearningPackage-local Content entries makes it easier for us to do pruning in small cycles, like as a post-publish task.

### Proposed Solution

Step 1: As a post-publish async task for any given `PublishableEntity`, delete all `PublishableEntityVersions` that are older than a certain period (1 week?), but preserve the following:
* The current Draft version
* Any `PublishableEntityVersion` that has ever been published (appears in a `PublishLogRecord`)

Rely on cascading deletion behavior to delete `Component/ComponentVersion`.

Step 2: After the deletions in Step 1, find any unreferenced `Content` entries and delete those.

ormsbee:
Some more thoughts on this...

We can delete old `PublishableEntityVersions` as new ones are created–no need to wait for publish. We should be able to do this relatively quickly, particularly since we'd only be deleting one at a time in that case.

The hard part about pruning is determining which `Content` are safe to delete.  The `components` app knows how to prune unused `Content` for `Content` that it has associations with, but other things might use that same content. Esp. if we model large collections of files as something other than `Components`. Also, pruning the files backing `Content` will be a slower operation.

If we're willing to allow `Content` pruning to be slow, we can have a pluggable thing where multiple apps get to contribute querysets to exclude from pruning.

So `Content` pruning would do this:

* Get a queryset for all content in a LearningPackage
* but exclude a pluggable list of querysets
* Components would be the first one to implement it. But FileSystems could as well. In any event, it's a contents-level concern.

So this prune gets called more periodically, say after a publish. It also works in small increments.

**Edge case**: Multiple `Content` entries can point to the same backing file if they're of different media types, so we need to be careful not to delete that file if there is any other `Content` referencing it.

PR #251, opened 2024-10-28
docs: ADRs for modeling containers capability

mariajgrimaldi:
### Description

ADRs with high-level decisions regarding the modeling of containers as a generalized capability of holding content, selectors as a strategy for selecting content that builds on the container idea, and units as a concrete implementation of containers that can contain static and dynamic content through selectors.

These decisions resulted from discussions in #38 and #219, which were consolidated into issue #240. 

In order to validate the proposal, we created a POC implementation #254 of what has been discussed so far. These ADRs and the POC are still a work in progress, but early reviews are welcome to validate the proposal.


openedx-webhooks:
Thanks for the pull request, @mariajgrimaldi!

### What's next?

*Please work through the following steps to get your changes ready for engineering review:*

#### :radio_button: Get product approval

#### :radio_button: Provide context

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation





#### :radio_button: Get a green build

#### :radio_button: Update the status of your PR

Your PR is currently marked as a draft.  After completing the steps above, update its status by clicking "Ready for Review", or removing "WIP" from the title, as appropriate.




#### :radio_button: Let us know that your PR is ready for review:

### Who will review my changes?


This repository is currently maintained by `@axim-engineering`.  Tag them in a comment and let them know that your changes are ready for review.


### Where can I find more information?

### When can I expect my changes to be merged?

- The need for product review

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,69 @@
...
+- Containers support both fixed and version-agnostic references for members, allowing members to be pinned to a specific version or set to reference the latest draft or published state.
+- The latest draft or published states can be referenced by setting the version to ``None``, avoiding the need for new instances on each update.
+- A single member (publishable entity) can be referenced by multiple containers, allowing for reuse of content across different containers.
+
+4. Version Control
+==================================
+
+- A new version is created if and only if the container itself changes (e.g., title, ordering of contents, adding or removing content) and not when its content changes (e.g., a component in a Unit is updated with new text).
+- When a new version is created because a member is added or removed, new references to the members are created to maintain the state of the content for the previous version.
+- Changes to the order of members within a container require creating a new version with the new ordering.
```
mariajgrimaldi, line 48:
```suggestion
- Changes to the order of members within a container require creating a new version of the container with the new ordering.
```

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,69 @@
...
+- Changes in pinned published or draft states require creating a new version to maintain the state of the content for the previous version.
```
mariajgrimaldi, line 49:
```suggestion
- Changes in pinned published or draft states require creating a new version of the container to maintain the state of the content for the previous version.
```

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
+17. Modeling Containers as a Generalized Capability for Holding Content
+========================================================================
+
+Context
+-------
+
+This ADR proposes a model for containers that can hold different types of content, including custom content types, and that can be used to create courses. The model defines the core structure and purpose of containers, the types of containers and content constraints, the members and relationships of containers, version control, publishing, and pruning.
```
mariajgrimaldi, line 7:
```suggestion
This ADR proposes a model for containers that can hold different types of content, including custom content types, and that can be used to model other content types with this behavior like units, subsections, sections or courses. The model defines the core structure and purpose of containers, the types of containers and content constraints, the members and relationships of containers, version control, publishing, and pruning.
```

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+- A single member (publishable entity) can be referenced by multiple containers, allowing for reuse of content across different containers.
+
+4. Version Control
+==================================
+
+- A new version is created if and only if the container itself changes (e.g., title, ordering of contents, adding or removing content) and not when its content changes (e.g., a component in a Unit is updated with new text).
+- Changes to the order of members within a container require creating a new version of the container with the new ordering.
+- Changes in pinned published or draft states require creating a new version of the container to maintain the state of the content for the previous version.
+- Each time a new version is created because of metadata changed, its members are copied from the previous version to preserve the state of the content at that time.
+- When using version-agnostic references, no new version is created when the content changes, and the reference is updated to the latest draft or published state.
```
mariajgrimaldi, line 52:
```suggestion
- When using version-agnostic references to members, no new version is created when members change since the latest draft or published state is always used.
```

mariajgrimaldi, line 52:
```suggestion
- Each time a new version is created because of metadata changed, its members are copied from the previous version to preserve the state of the content at that time.
- Changes in pinned published or draft states require creating a new version of the container to maintain the state of the content for the previous version.
- When using version-agnostic references, no new version is created when the content changes, and the reference is updated to the latest draft or published state.
```

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+- A single member (publishable entity) can be referenced by multiple containers, allowing for reuse of content across different containers.
+
+4. Version Control
+==================================
+
+- A new version is created if and only if the container itself changes (e.g., title, ordering of contents, adding or removing content) and not when its content changes (e.g., a component in a Unit is updated with new text).
+- Changes to the order of members within a container require creating a new version of the container with the new ordering.
+- Changes in pinned published or draft states require creating a new version of the container to maintain the state of the content for the previous version.
+- Each time a new version is created because of metadata changed, its members are copied from the previous version to preserve the state of the content at that time.
+- When using version-agnostic references to members, no new version is created when members change since the latest draft or published state is always used.
```
mariajgrimaldi, line 52:
```suggestion
- Each time a new version is created because of metadata changed, its members are copied from the previous version to preserve the state of the content at that time.
- Changes in pinned published or draft states require creating a new version of the container to maintain the state of the content for the previous version.
- When using version-agnostic references to members, no new version is created when members change since the latest draft or published state is always used.
```

docs/decisions/0018-units-as-containers.rst
```diff
@@ -0,0 +1,37 @@
...
+1. Unit Types and Content Constraints
+======================================
+
+- Units can only hold components as their members but will not enforce this restriction at the model level.
+- Content restrictions for units are implemented at the app layer, allowing units to limit their members to components.
+
+1. Unit Members and Relationships
+==================================
+
+- The members of a unit can only be components.
```
mariajgrimaldi, line 32: Should these decisions be part of this ADR as well?

- Unit members can exist as standalone items outside the unit.
- Members should indicate which unit they belong to.
- Members can be duplicated from units but keeping original references.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+- If a member is soft-deleted, the container will create a new version with the member removed.
+
+5. Publishing
+=============
+
+- Containers can be published, allowing their content to be accessible from where the container is referenced.
+- When a draft container is published, all its draft members are also published.
+- Members of a container can be published independently of the container itself.
+- If a member of a container is published independently, then it'd be published in the context of the container where it is referenced.
```
mariajgrimaldi, line 61: Should these decisions be part of this ADR as well?

- All members within the container should be published before the container can be reused.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+==================================
+
+- A new version is created if and only if the container itself changes (e.g., title, ordering of contents, adding or removing content) and not when its content changes (e.g., a component in a Unit is updated with new text).
+- Changes to the order of members within a container require creating a new version of the container with the new ordering.
+- Each time a new version is created because of metadata changed, its members are copied from the previous version to preserve the state of the content at that time.
+- Changes in pinned published or draft states require creating a new version of the container to maintain the state of the content for the previous version.
+- When using version-agnostic references to members, no new version is created when members change since the latest draft or published state is always used.
+- If a member is soft-deleted, the container will create a new version with the member removed.
+
+5. Publishing
```
mariajgrimaldi, line 55: What I'm still missing here is the behavior of publishing containers when they're being used in a course but they have been modified in some way. I understand we discussed this back in our last meeting, but it's still not clear to me what our approach is.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+
+3. Container Members and Relationships
+=======================================
+
+- The members of a container can be any type of publishable content.
+- Members within a container are maintained in a specific order as an ordered list.
+- Containers represent their content hierarchy through a structure that defines parent-child relationships between the container and its members.
+- The structure defining these relationships is anonymous, so it can only be referenced through the container.
+- Containers can hold both static and dynamically generated content, including user-specific variations.
+- Each container holds different states of its members (user-defined, initial, and frozen final state) to support rollback operations.
```
bradenmacdonald, line 37: I think an expanded description of these three states would be helpful. From this I'm not totally clear on what's the difference between initial and user-defined. And why "initial" is immutable but "frozen" is not (?) even though it sounds like it should be from the name.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+- Members can be added or removed from a container, and the container will maintain the state of the content for the previous version (frozen final state).
+- The initial state of a container is immutable.
+- When a container's structure changes, e.g., when a new member is added, the user-defined state of the container is updated with the new members list.
+- Containers support both fixed and version-agnostic references for members, allowing members to be pinned to a specific version or set to reference the latest draft or published state.
+- The latest draft or published states can be referenced by setting the version to ``None``, avoiding the need for new instances on each update.
```
bradenmacdonald, line 42: Is there a difference between specifying "latest draft" and specifying "latest published" or is there just a way to say "latest" (version=None), and whether that is draft or published depends on _______ ?


EDIT: OK, from looking at the code it appears that each container (via `EntityListRow`) can specify _both_ a `draft_version` and a `published_version` at the same time, separately. And each can be either pinned or unpinned. I guess I'm very unclear on how this works when the `ContainerEntityVersion` itself is versioned and has a draft + published version.

 If I have a draft `ContainerEntityVersion`, what do its `EntityListRow`s' `published_version`s mean? And when I publish it, so it's now a published `ContainerEntityVersion`, what do its `EntityListRow`s' `draft_version`s mean?

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+- A single member (publishable entity) can be referenced by multiple containers, allowing for reuse of content across different containers.
+
+4. Version Control
+==================================
+
+- A new version is created if and only if the container itself changes (e.g., title, ordering of contents, adding or removing content) and not when its content changes (e.g., a component in a Unit is updated with new text).
+- Changes to the order of members within a container require creating a new version of the container with the new ordering.
+- Each time a new version is created because of metadata changed, its members are copied from the previous version to preserve the state of the content at that time.
+- Changes in pinned published or draft states require creating a new version of the container to maintain the state of the content for the previous version.
```
bradenmacdonald, line 51: Can you "pin" to a specific draft version or published version? I would assume you can only pin to a specific version number, which isn't really considered "draft" or "published" - it's just a specific version (that may have been the latest published version and/or latest draft version at some point, but may or may not be now).

e.g.

```
1 2 3 4 5 6 7 8
    ^   ^     ^
    |   |     v8 is draft
    |   v5 is published
    Pinned to v3

In this scenario, v3 is just a specific version, neither published nor draft.
```

Edit: OK, I see now you can pin "draft" and "published" separately for each EntityRow, but I'm confused on the implications of that.

docs/decisions/0018-units-as-containers.rst
```diff
@@ -0,0 +1,37 @@
...
+All decisions from `0017-generalized-containers.rst`_ are still valid so that this decisions will build on top of them.
+
+.. _`0017-generalized-containers.rst`: 0017-generalized-containers.rst
+
+1. Units as Containers
+=======================
+
+- A unit is a concrete type of container that holds components.
+- A unit is a container, making it also a publishable entity.
+- A unit application will offer shared definitions for use by other unit subtypes.
```
bradenmacdonald, line 21: What is a "unit application"?

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+- Containers will follow extensibility principles in `0003-content-extensibility.rst`_ for creating new container types or subtypes.
+
+3. Container Members and Relationships
+=======================================
+
+- The members of a container can be any type of publishable content.
+- Members within a container are maintained in a specific order as an ordered list.
+- Containers represent their content hierarchy through a structure that defines parent-child relationships between the container and its members.
+- The structure defining these relationships is anonymous, so it can only be referenced through the container.
+- Containers can hold both static and dynamically generated content, including user-specific variations.
```
bradenmacdonald, line 36: How will "dynamically generated content" work?

bradenmacdonald (commented):
Thanks, I'm finding this to be a nice clear explanation in many places, but there are some things I'm confused by and I think we need to clarify them in the ADR.

mariajgrimaldi:
Thank you, @bradenmacdonald, for your comments! I'll go through your review now :)

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: For the expanded descriptions, I created a separate section for "Container States," grouping all the state management decisions instead: https://github.com/openedx/openedx-learning/pull/251/commits/af7dce4eb392ffa8967f7c9333ffb80c4f154489. I think now each definition is clear according to what was proposed in the model sketches.

mariajgrimaldi: > why "initial" is immutable but "frozen" is not (?) even though it sounds like it should be from the name

I understand frozen lists are "this is how the container looked (a screenshot?) when a new version was created". So, if there's a unit U0 with components `defined_list = P0latest, P1latest, P2latest` all pointing to their latest versions at the time of the next container creation where `frozen_list` would be `None` and `initial_list = P0, P1, P2` where P0, P1, and P2 are the pinned versions of `P0latest `and so on at moment of creation. When U1 is created, `defined_list` looks like `defined_list = P0latest, P1latest, P2latest` so frozen_list would be `frozen_list = P0', P1'', P2'` (the pinned versions of P0latest and so on). New versions for P0,...,1 are created, we'd need a way of knowing what versions were in U1 in case we want to go back to it, so we use frozen lists.

Frozen lists are useful when there are unpinned versions of members, since when they're all pinned, then `defined_list = initial_list = frozen_list`.

IThese in-line comments Dave left in #240 might explain this better: https://github.com/openedx/openedx-learning/pull/254/files#diff-6f2c589dc4ba5960e91d39f6488eb5e2e2e63ddaff63a75909091c760b877802R112-R154

bradenmacdonald: Thanks for expanding the descriptions, but I'm still confused :/

The code seems to call these "lists" not "states" - it would be good to pick a consistent term.

> The user-defined state of a container is the state that the author has defined for the version of the container

I don't think we should call this "user-defined" when it's actually "author-defined." To me, "user-defined" sounds more like "the version of this container that a specific user will see (after we've accounted for groups, A/B testing, randomized content, exam permissions, etc.). But "author-defined" is clear.

> - The initial state of a container is the state of the container when it was first created.

>    `# inital_list is an EntityList where all the versions are pinned, to show`
>    `# what the exact versions of the children were at the time that the`
>    `# Container was created.`

Is initial list really the list when the **container** was first created? (which I assume would usually be an empty list). I _think_ this should say "the initial list is a copy of the author-defined list that has all versions pinned as they were at the time the **ContainerEntityVersion** was created." Because if this is really something immutable from when the _container_ was created, not something for when each _version_ is created, then it should be on the container model, not redundantly specified for every ContainerEntityVersion.

> The frozen final state of a container is the state of the container at the time when a new version is created.

I think this needs an example: "While this ContainerEntityVersion is the current draft, this will be None which means that unpinned entities in the list should be showing their latest version. However, once this is published or an even newer draft is created, this frozen list should be saved to reflect the history of the list at that exact point in time when the version was finalized."

---------

I also don't really understand the need for an `initial_list` at all. Isn't it the same as "the frozen_list of the previous version"? And whether that's the case or not, when/why do we need to use it (initial_list)? I don't see any explanation in https://github.com/openedx/openedx-learning/issues/240 .

mariajgrimaldi: Oh, now I see that in the ADR itself I called it container states in some places and in others lists. I updated this also considering the author-defined suggestion:  https://github.com/openedx/openedx-learning/pull/251/commits/506d9cdc3a3dc53df3fdd2d33e8acfac95dae550

mariajgrimaldi: Oh, now I see that in the ADR itself I called it container states in some places and in others lists. I updated this also considering the author-defined suggestion:  https://github.com/openedx/openedx-learning/pull/251/commits/506d9cdc3a3dc53df3fdd2d33e8acfac95dae550

mariajgrimaldi: > Is initial list really the list when the container was first created? (which I assume would usually be an empty list). 

No. It's when the next container version is created, not the container entity itself. 

> I think this should say "the initial list is a copy of the author-defined list that has all versions pinned as they were at the time the ContainerEntityVersion was created." 

Yes, that's a better description of what's actually happening here. However, I want to keep this as high-level as possible, so I won't mention container entities or other model-specific definitions.

> Because if this is really something immutable from when the container was created, not something for when each version is created, then it should be on the container model, not redundantly specified for every ContainerEntityVersion.

Exactly, these lists are actually related to each container version, not the container entity. I'll make that clear.

mariajgrimaldi: Oh, now I see that in the ADR itself I called it container states in some places and in others lists. I updated this also considering the author-defined suggestion. Thanks for the suggestions!

---

> Is initial list really the list when the container was first created? (which I assume would usually be an empty list). 

No. It's when the next container version is created, not the container entity itself. 

> I think this should say "the initial list is a copy of the author-defined list that has all versions pinned as they were at the time the ContainerEntityVersion was created." 

Yes, that's a better description of what's actually happening here. However, I want to keep this as high-level as possible, so I won't mention container entities or other model-specific definitions unless you consider them necessary.

> Because if this is really something immutable from when the container was created, not something for when each version is created, then it should be on the container model, not redundantly specified for every ContainerEntityVersion.

Exactly, these lists are actually related to each container version, not the container entity. I'll make that clear.

---

> I think this needs an example: "While this ContainerEntityVersion is the current draft, this will be None which means that unpinned entities in the list should be showing their latest version. However, once this is published or an even newer draft is created, this frozen list should be saved to reflect the history of the list at that exact point in time when the version was finalized."

In "...unpinned entities in the list should be showing their lates version"  do you refer to the author-defined list, right?

---

> I also don't really understand the need for an initial_list at all. Isn't it the same as "the frozen_list of the previous version"? And whether that's the case or not, when/why do we need to use it (initial_list)? I don't see any explanation in https://github.com/openedx/openedx-learning/issues/240 .

I think the idea of an initial state came from this comment: https://github.com/openedx/openedx-learning/issues/38#issuecomment-2395589146 -- still, it doesn't say the end purpose. But as far as I understand, it's included for convenience according to: 

    # inital_list is an EntityList where all the versions are pinned, to show
    # what the exact versions of the children were at the time that the
    # Container was created. We could technically derive this, but it would be
    # awkward to query.
    #
    # If the Container was defined so that all references were pinned, then this
    # can point to the exact same EntityList as defined_list.

bradenmacdonald: Yeah, I saw that, but I still don't understand the use case. I also can't tell if it's any different than "the frozen_list of the previous version", or if those are the same thing.

@ormsbee maybe you can clarify?

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: According to #240, dynamically generated content will be built on top of what the issue called "selectors". I thought we could write a different ADR for it. I'm going to create an empty ADR with some basic info at the moment, so it's clear we will address it in this PR.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Decisions
+---------
+
+1. Core Structure and Purpose of Containers
+===========================================
+
+- A container is designed as a generalized capability to hold different types of content.
+- A container is a publishable content type that holds other content types through a parent-child relationship.
+- A container application will offer shared definitions for use by other container types.
```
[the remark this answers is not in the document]
mariajgrimaldi: I think I explained it better in this comment: https://github.com/openedx/openedx-learning/pull/251#discussion_r1842059553. I was referring to a Django application where the APIs and mixins that other container types would be built on.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,90 @@
...
+
+3. Container Members and Relationships
+=======================================
+
+This section defines container members, their order, and relationships, covering flexible connections and support for draft and published states of their members.
+
+- The members of a container can be any type of publishable content.
+- Members within a container are maintained in a specific order as an ordered list.
+- Containers represent their content hierarchy through a structure that defines parent-child relationships between the container and its members.
+- The structure defining these relationships is anonymous, so it can only be referenced through the container.
```
mariajgrimaldi, line 41: note to self: this seems too implementation-specific to be part of these high level decisions.

docs/decisions/0018-units-as-containers.rst
bradenmacdonald: > * Generalized containers (containers app–lowest level of these)
> * Selectors for dynamically selecting 0-N PublishableEntities, i.e. how we're going to do things like SplitTest and Randomized (selectors app, builds on containers).
> * Units (units app, builds on containers and selectors)–basically empty shells at the moment.

Can you include this whole example? That really makes it more clear.

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: I removed this from the generalized container capability into the unit ADR, since it makes more sense to be there.

docs/decisions/0018-units-as-containers.rst
```diff
@@ -0,0 +1,67 @@
+18. Modeling Units as a Concrete Implementation of the Container Capability
```
mariajgrimaldi, line 1: I did this exercise, in which I instanced the decisions made for containers to be for units and components, and it's way easier to understand!

docs/decisions/0017-generalized-containers.rst
bradenmacdonald: Ah, I see now:

> ```
>     # These entities
>     # could be Selectors, in which case we'd need to do more work to find the right
>     # variant.
> ```

The part that I'm struggling with is that we seem to be using a lot of complexity and work (frozen_list, initial_list) to support both pinned and unpinned versions at this base layer, and I was kind of assuming that by paying the price of that complexity, we would also be able to handle "selectors" and dynamic content. But it seems like that's going to be yet another layer on top of this.

The way that "selectors" will have to deal with pinned/unpinned references and children seems very similar and I am kind of hoping that we'll be able to find some commonality and simplify this. It may be too early to know that though.

> Note that we could implement containers and units first, and add selectors later, but I would like to validate selectors because it's the riskiest part of this design.

I agree with this statement. It's hard to know how good this high level design is until we see how it works with selectors.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,70 @@
...
+=======================================
+
+- The members of a container can be any type of publishable content.
+- Members within a container are maintained in a specific order as an ordered list.
+- Containers represent their content hierarchy through a structure that defines parent-child relationships between the container and its members.
+- The structure defining these relationships is anonymous, so it can only be referenced through the container.
+- Containers can hold both static and dynamically generated content, including user-specific variations.
+- Each container holds different states of its members (user-defined, initial, and frozen final state) to support rollback operations.
+- Members can be added or removed from a container, and the container will maintain the state of the content for the previous version (frozen final state).
+- The initial state of a container is immutable.
```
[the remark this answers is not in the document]
mariajgrimaldi: Now, it mentions the initial list of members instead of "state".

docs/decisions/0018-units-as-containers.rst
mariajgrimaldi: Yes, sure!

docs/decisions/0017-generalized-containers.rst
ormsbee: > If I have a draft `ContainerEntityVersion`, what do its `EntityListRow`s' `published_version`s mean? And when I publish it, so it's now a published `ContainerEntityVersion`, what do its `EntityListRow`s' `draft_version`s mean?

I originally had it with just `version`, but I vaguely remember thinking that I had to add `draft_version` separately from `published_version` to address some edge case... which I have totally forgotten. And maybe was wrong-headed to begin with. 😛 Let's take it back out for now–if said weird edge case was real, I'm sure we'll run into it in the not-too-distant future, and deal with it then.

docs/decisions/0017-generalized-containers.rst
ormsbee: > The way that "selectors" will have to deal with pinned/unpinned references and children seems very similar and I am kind of hoping that we'll be able to find some commonality and simplify this. It may be too early to know that though.

I think that when I first sketched this part of the data model, my thinking was that we'd want to always pin all the entities referenced from the Variant, because we might be dynamically generating a different Variant per user–something that would make it really expensive to amend things if someone deleted an entity that was referenced in an unpinned way. I also suspect that I sketched the Selector/Variant piece before I started making the distinction between defined/initial/frozen lists, since a lot of that was added to deal with deleting members.

But I agree with your point that there's more overlap here than this model is representing now. Author-defined Variants (like A/B tests) are much more like their own kind of ContainerEntityVersions. Dynamically generated Variants (like randomized subset), could be pointers to those containers + a pinned EntityList.

I feel like all the fundamental pieces are there, if we just nudge things around a bit. I'll chew on it.

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: @ormsbee: during our latest discussion, you mentioned that having both, mainly the draft version, covered the CCX use cases where we don't necessarily control the latest draft version, so we'd want to pin it for the author's context. Do you think we should cover that case later down the line?

We also discussed that having both would help us know which published/draft versions were in the frozen list when creating the next version, but I think that the use case could be covered by having a single reference to the current version draft or published. Would you agree?

docs/decisions/0019-selectors.rst
```diff
@@ -0,0 +1,4 @@
+Selectors for Dynamically Selecting Content
```
mariajgrimaldi, line 1: I'm currently working on this section. Thanks!

mariajgrimaldi: This is the first version of selectors based on the information that's available. I still haven't considered [these comments](https://github.com/openedx/openedx-learning/pull/251#discussion_r1845288014) so I'll be working on it.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,101 @@
...
+==========
+
+This section defines the rules for pruning container versions, explaining when a container version can be pruned and the effects of pruning on the container and its members.
+
+- A container version can be pruned if:
+  #. It's not being used by any other container.
+  #. It's not a published version.
+  #. It's not the latest version of the container.
+- In a top-down approach, start the deletion process with the parent container and work your way down to its members. E.g., when pruning Section V2 > Subsection V1 > Unit V3, the deletion process starts in the greater container working its way down to the smaller.
+- Pruning a container version will not affect the container's history or the members of other container versions, so containers will not be deleted if they are shared by other containers.
```
mariajgrimaldi, line 97: This is my first approach to pruning, based on https://github.com/openedx/openedx-learning/issues/240#issuecomment-2403407812 and #154. But I still need some clarification about the implication of removing an entity version that's pinned somewhere. When doing so, all versions referencing that version would have to go, too since they'd be invalidated. Is that a correct assumption?

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,101 @@
...
+- The author-defined list is the list of members that the author has defined for the version of the container.
+- The author-defined list won't change for a specific container version even if its references get soft-deleted.
+- The initial list is a copy of the author-defined list that has all versions pinned as they were at the time the container version was created.
+- The initial list is immutable for a container version.
+- The frozen list refers to the list of members at the time when the next version of the container is created.
+- The author-defined list is used to show the content of a container version as the author specified it, the frozen list can be used for discard operations on a draft version and the initial-list is part of the history of evolution of the container.
+
+Let's say a course author creates a unit with three components, all using floating versions. Each component's latest version is V1. The author-defined list would include these three components, ordered as the author decided. The initial list would have the components pinned to V1, while the frozen list would be empty until we create the next version for the container.
```
mariajgrimaldi, line 62:
```suggestion
Let's say a course author creates a unit version with three components, all using floating versions. Each component's latest version is V1. The author-defined list would include these three components, ordered as the author decided. The initial list would have the components pinned to V1 since that's the components' versions for this moment in time, while the frozen list would be empty until we create the next version for the unit.
```

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Now, when the author creates a new version of the unit, for example, V2, we need to store the latest state of the container in the frozen list. This means pinning the latest versions of the components at that time, let's say V1, V2, and V3, respectively.
+
+Next, imagine the course author creates a new unit but uses pinned references for the components instead of floating versions, as they don't want to use the latest updates. In this case, the author-defined list, initial list, and frozen list would all be the same, as the component versions remain fixed. If we were to use different pinned versions, then a new unit version would be created instead.
```
mariajgrimaldi, line 66:
```suggestion
Next, imagine the course author creates a another unit version but uses pinned references for the components instead of floating versions, as they don't want to use the latest updates. In this case, the author-defined list, initial list, and frozen list would all be the same, as the component versions remain fixed. If we were to use different pinned versions, then a new unit version would be created instead.
```

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Next, imagine the course author creates a another unit version but uses pinned references for the components instead of floating versions, as they don't want to use the latest updates. In this case, the author-defined list, initial list, and frozen list would all be the same, as the component versions remain fixed. If we were to use different pinned versions, then a new unit version would be created instead.
+
+5. Next Container Versions
+==================================
+
+This section defines the rules for version control in containers, explaining when new versions are created based on changes to container structure or metadata.
+
+- A new version is created if and only if the container itself changes (e.g., title, ordering of members, adding or removing members) and not when its members change (e.g., a component in a Unit is updated with new text). For instance, a new version of a unit is created when a component is removed, not when a new version of a component is created.
```
mariajgrimaldi, line 73: Does unpinning a version from a container merit creating a new version for the container? Or is this an entirely different container?

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: > But I agree with your point that there's more overlap here than this model is representing now. Author-defined Variants (like A/B tests) are much more like their own kind of ContainerEntityVersions. Dynamically generated Variants (like randomized subset), could be pointers to those containers + a pinned EntityList.

So, would variants be a container type with special rules (author-defined or per-user) for their author-defined list shown to students?

docs/decisions/0018-units-as-containers.rst
```diff
@@ -0,0 +1,78 @@
...
```
mariajgrimaldi, line 1: I'm not sure if this deserves its own ADR, but it helps illustrate the decisions using a more familiar concept like units.

docs/decisions/0018-units-as-containers.rst
```diff
@@ -0,0 +1,78 @@
...
+- Unit subtypes can be created by following the extensibility principles in `0003-content-extensibility.rst`_.
+
+3. Unit Members and Relationships
+==================================
+
+- The members of a unit can only be components.
+- Components are referenced as an ordered list in a unit.
+- Units can hold both static and dynamic content, such as user-specific variations.
+- Units can reference pinned and unpinned versions of its components.
+- The latest draft or publish version of a component can be set by using `None` in thr parent-child relationship between units-components.
```
mariajgrimaldi, line 38:
```suggestion
- The latest draft or publish version of a component can be set by using `None` in the parent-child relationship between units-components.
```

docs/decisions/0018-units-as-containers.rst
```diff
@@ -0,0 +1,78 @@
...
+- The author-defined list is the list of components defined by the author for a specific unit version.
+- The author-defined list of components won't change for a specific version.
+- The initial list is a copy of the author-defined list that has all components pinned to the versions at the time of the unit version creation.
+- The initial list is immutable for a unit version.
+- The frozen list refers to the list of components at the time when the next version of the unit is created.
+- When creating the author-defined list of a new version with pinned references, then the author-defined list is the same as the initial and frozen list. When creating a new version with unpinned references, then the frozen list starts as `None` and should be updated with the author-defined components pinned when a new version is created.
+- The author-defined list is used to show the content of a unit version as the author specified it, the frozen list can be used for discard operations on a draft version and the initial-list is part of the history of evolution of the unit.
+- These lists allow history tracking of a unit version and revert operations.
```
mariajgrimaldi, line 52: note to self: This needs to be updated according the latest changes in the container ADR.

docs/decisions/0017-generalized-containers.rst
ormsbee: > @ormsbee: during our latest discussion, you mentioned that having both, mainly the draft version, covered the CCX use cases where we don't necessarily control the latest draft version, so we'd want to pin it for the author's context. Do you think we should cover that case later on?

I think I was mistaken there, because if the author wanted to pin to a specific version, they could do so, and then when they decided to undo that pin to get something later, they could do so with a new version of the container–no need to keep the publish versioned separate.

> We also discussed that having both would help us know which published/draft versions were in the frozen list when creating the next version, but I think that the use case could be covered by having a single reference to the current version draft or published. Would you agree?

Yeah, I agree.

I think where my head might have been going is what happens in the following scenario:

1. UnitVersion UV1 has an unpinned reference to Component C1.
2. C1 has ComponentVersions CV1 and CV2. CV1 is the current published version, while CV2 is the draft version.
3. C1 is soft-deleted, forcing UV1 to pin down its references to the last version. But does it pin to CV1 or CV2?

-----

So I'm still kicking it around in my head, but I'm trying to see how things line up with Unit/Container modeling if we don't force new container versions to be created when things get deleted, and instead filter out the deleted stuff at read time. I think it could get rid of a lot of the bookkeeping needs for the model. I'm trying to sketch this out this evening to see if it fits together in a reasonable way...

docs/decisions/0019-selectors.rst
```diff
@@ -0,0 +1,40 @@
...
```
mariajgrimaldi, line 1:
```suggestion
19. Selectors for Dynamically Selecting Content
```

mariajgrimaldi:
I'm opening this PR for review as we continue to discuss the proposal, although it doesn't mean we've reached a consensus yet. Thank you!

mariajgrimaldi:
You can find the latest discussion about unit/container modeling in this slack thread: https://openedx.slack.com/archives/C05NT2YN820/p1732135462177269

I'm hoping we'll write down the outcome of the discussion here soon.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Let's say a course author creates a unit version with three components, all using floating versions. Each component's latest version is V1. The author-defined list would include these three components, ordered as the author decided. The initial list would have the components pinned to V1 since that's the components' versions for this moment in time, while the frozen list would be empty until we create the next version for the unit.
+
+Now, when the author creates a new version of the unit, for example, V2, we need to store the latest state of the container in the frozen list. This means pinning the latest versions of the components at that time, let's say V1, V2, and V3, respectively.
+
+Next, imagine the course author creates a another unit version but uses pinned references for the components instead of floating versions, as they don't want to use the latest updates. In this case, the author-defined list, initial list, and frozen list would all be the same, as the component versions remain fixed. If we were to use different pinned versions, then a new unit version would be created instead.
```
mariajgrimaldi, line 65: According to our latest discussion had [here](https://openedx.slack.com/archives/C05NT2YN820/p1732135462177269) this must be changed to address:

1. References to defined, initial and frozen must be dropped in favor of relaxing the requirement of "container versions should not reference soft-deleted children"
2. Emphasize "only changes of the container itself generate a new version of the container," so NO external change (changes in container children) should create a new version of a container. If a container has invalid references to children, then it should be filtered out as needed.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,97 @@
...
+
+- The members of a container can be any type of publishable content. E.g., sections, subsections, units, components, and any other publishable thing. For more details on publishable content, see `PublishableEntity`_.
+- Members within a container are maintained in a specific order as an ordered list. E.g., components within a unit, or units within a subsection, are presented in a specific order.
+- Containers represent their content hierarchy through a structure, like Course > Section > Subsection > Unit > Component, which defines parent-child relationships at each level.
+- Containers can reference a specific version of their members or be set to point to their latest versions. For instance, component V1 might be used in a unit instead of its latest version. The latest version of a member can be referenced by setting its version to ``None`` which consists of the chosen standard for this representation.
+- A single member (publishable entity) can be shared by multiple containers, allowing for reuse of content across different containers. For instance, a component can be shared by multiple units.
+
+4. Container Version History
```
mariajgrimaldi, line 48: During discussions, container members were usually referenced as "children", so I will change all references to members to children.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,97 @@
...
+
+5. Next Container Versions
+==========================
+
+This section defines the rules for version control in containers, explaining when new versions are created based on changes to container structure or metadata.
+
+- A new version is created if and only if the container itself changes (e.g., title, ordering of members, adding or removing members) and not when its members change (e.g., a component in a Unit is updated with new text). For instance, a new version of a unit is created when a component is removed, not when a new version of a component is created.
+- When a shared member is soft-deleted in a another container, all containers referencing it should create a new version without the member. This new version will be the new draft version of the container. For example, suppose a component is shared between two units, if the component is soft-deleted independently, then we'd need to create a new version for both units sharing the component.
```
mariajgrimaldi, line 73: Please, review: https://github.com/openedx/openedx-learning/pull/251/files#r1871312529

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,79 @@
...
+- The course hierarchy Course > Section > Subsection > Unit will be implemented as relationships between containers, with each level acting as a container that holds other content. The hierarchy will be enforced by the content restrictions of each particular container but allowed to be overridden to support `Approach to Content Flexibility <0002-content-flexibility.rst>`_.
+- Containers will follow extensibility principles in `Content Extensibility Through Model Relations <0003-content-extensibility.rst>`_ for creating new container types or subtypes.
+
+3. Container Children and Relationships
+=======================================
+
+This section defines container children, their order, and relationships, covering flexible connections and support for draft and published states of their children.
+
+- Each container version holds a list of children that the author has defined for that version.
+- The author-defined list is used to show the content of a container version as the author specified it
```
mariajgrimaldi, line 43:
```suggestion
- The author-defined list is used to show the content of a container version as the author specified it.
```

docs/decisions/0019-selectors.rst
```diff
@@ -0,0 +1,40 @@
...
+- The logic for pushing members into variants depends on the selector selection method. For example, A/B split testing two different sets of components or select three problems from a set of twenty.
+- Variants hold the members selected for a container based on what the selection method is. E.g., if the selector is "select 5 components out this pool of 20 components" then the variant would be the 5 components selected for the user.
+- Variants are build on the parent-child relationship used for containers and their members, storing the dynamically selected content as an ordered list as containers do.
+
+2. Selector Types and Selecting Content
+=======================================
+
+This section describes how different types of selectors work and how they handle the selection of dynamic content.
+
+- A selector can be of any type, which means it can implement any method to select members from a pool. Therefore, selectors will follow extensibility principles in `0003-content-extensibility.rst`_ for creating new selector types.
```
mariajgrimaldi, line 30:
```suggestion
- A selector can be of any type, which means it can implement any method to select members from a pool. Therefore, selectors will follow extensibility principles in `0003-content-extensibility.rst <0003-content-extensibility.rst>`_ for creating new selector types.
```

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,79 @@
...
+- Containers can reference a specific version of their children or be set to point to their latest versions. For instance, component V1 might be used in a unit instead of its latest version. The latest version of a child can be referenced by setting its version to ``None`` which consists of the chosen standard for this representation.
+- A single child (publishable entity) can be shared by multiple containers, allowing for reuse of content across different containers. For instance, a component can be shared by multiple units.
+
+4. Next Container Versions
+==========================
+
+This section defines the rules for version control in containers, explaining when new versions are created based on changes to container structure or metadata.
+
+- A new version is created if and only if the container itself changes (e.g., title, ordering of children, adding or removing children) and not when its children change (e.g., a component in a Unit is updated with new text). For instance, a new version of a unit is created when a component is removed, not when a new version of a component is created.
```
bradenmacdonald, line 56: > a new version of a unit is [not created] when a new version of a component is created.

How does this interact with the library workflows?

You have a unit in a library with one component. You publish everything and use it in a course. You edit that component in the library. The unit does not show as changed and can't be published. The component _does_ show as changed, and can be published. When you publish it, you'll see a notification in the course that a newer version of the component is available, and you'll update it directly, not via updating the unit?

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,79 @@
...
+- No external change to a child will trigger the creation of a new version for the container. For example, updating or soft-deleting a component will not trigger the creation of a new version for the unit.
```
bradenmacdonald, line 57: > soft-deleting a component will not trigger the creation of a new version for the unit.

How does a "soft deletion" get published? Is it always applied immediately? Based on the ADR as written, it seems it is impossible to soft delete a child, preview that in the LMS "preview" mode before students see it, then later publish the deletion so that students see it.

ormsbee: If a component is soft-deleted today, that means its `draft_version` is set to None. Its `published_version` remains whatever it was before until that soft-delete is published. So you should still be able to look at the draft version of a Unit and see the soft-delete reflected before the publishing of the component happens.

That being said, I think the normal flow of things when we're working in the Studio view of a Unit is that when we delete a component from a Unit, we actually want to create a new version of the Unit to reflect that deletion. Soft-deletion not automatically _triggering_ cascading creations of new containers was a way to try to deal with things that may up in many, many container-like things (e.g. variants of a randomized pool).

ormsbee: I also have this vague recollection that it helped to simplify some revert scenarios, but i'm just now getting back into this doc and trying to page all this stuff back into my brain.

mariajgrimaldi:
@bradenmacdonald @ormsbee: thank you for reviewing this! I'll revisit it early next week. :)

bradenmacdonald:
@mariajgrimaldi Great! I have also started expanding the test cases in your prototype, which you can see here: https://github.com/openedx/openedx-learning/pull/278

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: I imagine that would happen if we set the unit component to point to the latest version. We wouldn't need to update the unit itself because nothing really changed. Does that make sense?

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: > That being said, I think the normal flow of things when we're working in the Studio view of a Unit is that when we delete a component from a Unit, we actually want to create a new version of the Unit to reflect that deletion.

Agree. I _think_ this item refers to what happens when soft-deleting the component itself, but I think that makes describing the workflow a bit more confusing. If that's the case, it also makes sense that a new version is also created when the deletion is published which is not exactly what I said here.

docs/decisions/0017-generalized-containers.rst
bradenmacdonald: Yeah, I think I understand this now.

> You have a unit in a library with one component. You publish everything and use it in a course. You edit that component in the library. The unit ~~does not show as changed and can't be published~~ **shows that it _contains_ unpublished changes and you can publish those changes** ~~The component does show as changed, and can be published.~~ When you publish it, you'll see a notification in the course that a newer version of the component is available, and you'll update it directly, not via updating the unit ~~?~~.

docs/decisions/0017-generalized-containers.rst
bradenmacdonald: I think there's a lot of ambiguity here, so I'd like to clarify it as three different scenarios:

1. You *remove* a component from a unit, but the component continues to exist. This necessarily creates a new version of the unit.
2. You [soft] delete a component that happens to be used in a unit; you do this from somewhere else in Studio like when viewing that component independently in the library (not in the unit context). Once you publish that deletion, the component will disappear from the learner views of the course. This does not create a new version of the unit (changes don't cascade up).
   -  If you later restore the component "un-delete", it will re-appear in the unit.
   - The Studio UI must indicate somewhere that a deleted component appears as a child, and allow expicitly removing it from the unit (triggering scenario 1), and preventing the un-deletion from affecting the unit in the future. This would be optional and separate. This view that shows a "ghost component" where the component used to be could also provide a button for un-deleting it and/or cloning it and replacing the deleted component with a new duplicate.
3. From the unit page in Studio, you [soft] delete a component from a unit. In this case, the frontend/platform will combine two actions: soft deleting the component and removing it from the list of the unit's children. This will create a new version of the unit. You publish these changes in a single publish action, creating a `PublishLog` that affects both the unit and the deleted component.

Does that make sense?

mariajgrimaldi: Yes, that makes a lot of sense. Should we update the ADR to reflect these specifications? It'd clear up a lot of the ambiguity you mentioned.

Thanks for updating the unit tests to illustrate this better!

bradenmacdonald:
I think the main TODO is that in https://github.com/openedx/openedx-learning/pull/278 I combined `defined_list`, `initial_list`, and `frozen_list` into a single `entity_list`, and I think it's a nice simplification. So we should simplify the ADR by combing those as well. The unit tests in my PR demonstrate that you can still achieve all the project requirements without maintaining those separate lists.

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: @bradenmacdonald: yes! I'm on it :)

mariajgrimaldi:
@bradenmacdonald, I updated the ADR back in December to consider the single list change: https://github.com/openedx/openedx-learning/pull/251/files#diff-48d371d83c194bb10846202d3f78323c577c9ca2e9c7e840c8edec4829c90083R42-R49. In any case, I'm going to review it again to see if there's anything we can improve or that needs changing. Thanks!

mariajgrimaldi:
@kdmccormick @bradenmacdonald: updating this is my current priority, thanks for the patience!

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: Done: https://github.com/openedx/openedx-learning/pull/251/commits/901c560f661be99b13b04ac427dca60ff49f7d4e

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,81 @@
...
+
+4. Container Versioning, Soft-Deletions and Structural Changes
+==============================================================
+
+This section defines the versioning rules for containers, explaining when a new version is created based on structural or metadata changes. It also clarifies how changes in child elements affect (or do not affect) container versions.
+
+- A new version of a container is created only when the container itself changes, such as modifying its title, ordering of children, or adding/removing children. Changes to a child do not cascade up, meaning modifications to a child do not trigger a new version of its parent. This includes the following cases:
+
+  - Soft-deleting a child that is referenced by a container does not create a new version of the container. For example, soft-deleting a component used by a unit will not generate a new version of the unit. Once the deletion is published, the child will be filtered out from the container as needed to ensure consistency.
```
kdmccormick, line 58: What is soft-deleting?

kdmccormick (commented):
(Partial review, overall looking great so far)

docs/decisions/0017-generalized-containers.rst
bradenmacdonald: Basically deleting. But we achieve it by setting the Draft to NULL, and when published this sets the published version to NULL, but all the historic versions are there and it can be un-deleted at any time by pointing the Draft back to any version (non-NULL).

mariajgrimaldi: I added a key concepts section with those terms to (try) reducing ambiguity:  https://github.com/openedx/openedx-learning/pull/251/commits/20688f13ab7adaea0fd73d5d00f5b40c43d01a6c

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,94 @@
...
+- The course hierarchy Course > Section > Subsection > Unit will be implemented as relationships between containers, with each level acting as a container that holds other content. The hierarchy will be enforced by the content restrictions of each particular container but allowed to be overridden to support `Approach to Content Flexibility <0002-content-flexibility.rst>`_.
+- Containers will follow extensibility principles in `Content Extensibility Through Model Relations <0003-content-extensibility.rst>`_ for creating new container types or subtypes.
+
+3. Container Children and Relationships
+=======================================
+
+This section defines container children, their order, and relationships, covering flexible connections and support for draft and published states of their children.
+
+- Each container version holds a list of children that the author has defined for that version, we will call this the entity list.
+- The entity list represents the content of a container version as defined by the author. It can be used to derive the initial versions of the container's children at the time of creation or the latest versions available at that moment.
```
mariajgrimaldi, line 56: @bradenmacdonald: what do you think about this section about the new single entity list?

bradenmacdonald: I think it's more accurate to the prototype to just say "The entity list represents the content of a container version as defined by the author, **which may be a list of specific component versions or just a list of components without version information**."

docs/decisions/0017-generalized-containers.rst
bradenmacdonald: @kdmccormick We use "removal" and "deletion" for completely different things. Removal just means taking it off of some list (e.g. remove a unit from a container), but nothing is [soft] deleted.

> As a user, I remove a component from a unit within a library. This is a soft-deletion, right? Has the component also been removed from the library?

No, it's just removal, not deletion. The component continues to exist in the library and perhaps in other units.

> As a user, I remove a component from the library. 

There is no such concept as to "remove a component from the library". You may [soft] delete it however. Deleting a component is different than removing it from a unit.

kdmccormick: > There is no such concept as to "remove a component from the library"

But, from a user's perspective, it seems like there is such a concept, and it's we're calling it soft-deletion, no?

If not, then what does a user do in order to trigger the soft-deletion of something?

bradenmacdonald: @kdmccormick That's just called "deleting the component" (and in the code we call it "soft deletion" but that phrase is never used in the UI).

Screenshot 2025-03-05 at 10 52 26 AM

I just don't want to call that "removing" the component because it's doing a deletion, and in the unit context removal and deletion are very different things - the latter affects other stuff and the former only affects the unit you're removing from.

kdmccormick: OK, I understand now, ty.

Some terminology nits for the ADR, consider these all optional: the words "removal" and "(soft-)deletion" did not mean anything different to me until you explained the difference. And "soft-deletion" kept making me think that there'd be a such thing as "hard-deletion", which there isn't, unless that's what pruning is. I would just qualify it every time: "removal from the parent container", "deletion from the library". No need to specify "soft-" since IMO it's obvious that if we can undo a deletion prior to pruning, then the deletion does not actually involve immediately wiping the data from the database--any system that supports Undo would have to work this way.

bradenmacdonald: I agree that constantly using the term "soft deletion" is awkward. Those suggestions seem reasonable to me.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Key Concepts
+============
+
+The following definitions provide context for the terms used in this ADR:
+
+- Container: A content structure that holds other content types in a parent-child hierarchy (e.g., sections, subsections, units).
+- PublishableEntity: A versioned content type that supports publishing and can exist independently or within a container (e.g., sections, units, components).
+- Versioning: The process of creating new container versions when its structure or metadata changes (e.g., title, order, adding/removing children).
+- Soft-deletion: Removing a draft or published entity by setting its version to NULL, while keeping historical versions for potential reverts.
```
kdmccormick, line 17:
```suggestion
- Deletion: Removing a draft or published entity from the learning package. This is done by setting its version to NULL, while keeping historical versions for potential reverts.
- Removal from a container: Removing an entity from a parent container, without removing it from the learning package. 
```

kdmccormick (approved):
My only feedback is to drop "soft-" from "soft-deletion"; I find it confusing. Consider that optional feedback.

Otherwise, LGTM--solid architecturally and very well written. As someone coming back to this project after a few weeks working on other things, these did a good job helping me get back up to speed.

docs/decisions/0017-generalized-containers.rst
mariajgrimaldi: Got it! Thank you

kdmccormick:
I removed Braden's review request since he's on vacation for the rest of the month.

@ormsbee , did you want to do another pass , or is Maria good to merge?

mariajgrimaldi:
@ormsbee: thank you! I'll be very attentive if any changes are needed.

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,95 @@
...
+---------
+
+1. Core Structure and Purpose of Containers
+===========================================
+
+This section defines the purpose and structure of containers, explaining how they are designed to hold various types of content through a parent-child setup.
+
+- A container is designed as a generalized capability to hold different types of content.
+- A container is a publishable content type that holds other content types through a parent-child relationship. For example, sections, subsections and units.
+- The generalized container capability will have its own Django application as part of the authoring application allowing other types of containers and content types will build on top of. For instance:
```
ormsbee, line 33: Containers were folded into the `publishing` app because of the current coupling that exists between drafts and publishing between children and parents. The alternative would have been to do everything via signals, but we decided that it wasn't worth the added complexity and potential for confusion.

mariajgrimaldi: Updated! Thank you

kdmccormick (changes requested):
Latest edits LGTM content-wise.

@mariajgrimaldi , I think some of the headings are mixed up--the title heading and the subsection headings are the same size. Can you take look at those, and also update them to match the RST style guide? https://docs.openedx.org/en/latest/documentors/references/quick_reference_rst.html#headings

docs/decisions/0017-generalized-containers.rst
```diff
@@ -0,0 +1,95 @@
...
+
+5. Publishing
+=============
+
+This section explains the publishing process for containers, detailing how containers and their children become accessible, either together or independently, based on their publication state. The publishing process happens on container versions, but throughout this section we'd call them containers for simplicity.
+
+- Containers can be published, allowing their content to be accessible from where the container is being used.
+- When a draft container is published, all its draft children are also published. For instance, after publishing a draft version of subsection which contains a draft unit with an updated title, the latest published version of the unit will be the one with the updated title, reflecting the changes made previously.
+- Children of a container can be published independently of the container itself. E.g., a shared component can be published independently of the unit if it also exists outside the unit.
+- Containers are not affected by the publishing process of its children. This means that publishing a component won't trigger new publishing processes for a container.
```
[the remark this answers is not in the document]
mariajgrimaldi: @kdmccormick: what do you think about this?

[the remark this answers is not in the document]
kdmccormick: I think I'm confused. @ormsbee , why does publishing a child affect its parent's "last published at"? Is that just a UI-level decision we made, or is there something in the data model / python API which is bubbling the publish action up to the parent's metadata?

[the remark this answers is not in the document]
kdmccormick: Thanks Dave. I am curious how we will avoid confusing authors into thinking that their full container's changes are published when only a child's changes have been published. But, we don't need to resolve that for this ADR.

@mariajgrimaldi , may I suggest:
> -  Containers are not affected by the publishing process of its children. This means that publishing a component won't trigger new publishing processes for a container. However, authors will find it relevant that a container's child has changed; therefore, in the UI, publishing a child will most likely affect the date at which its ancestor containers are shown to have been "last published at".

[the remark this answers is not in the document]
mariajgrimaldi: Done! Thank you both for the help :)

https://github.com/openedx/openedx-learning/pull/251/commits/728c1e767e19acad67ca1d5a3030adae15ae0eb5

mariajgrimaldi:
@ormsbee @bradenmacdonald @kdmccormick: thank you all for the help!

PR #254, opened 2024-11-05
[WIP] feat: implement sketch models and api for interactions based on 0017/18

mariajgrimaldi:
### Description
This implementation is based on the high-level decisions made in #251

Unit tests are now the way I'm testing this out.


openedx-webhooks:
Thanks for the pull request, @mariajgrimaldi!

### What's next?

*Please work through the following steps to get your changes ready for engineering review:*

#### :radio_button: Get product approval

#### :radio_button: Provide context

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation





#### :radio_button: Get a green build

#### :radio_button: Update the status of your PR

Your PR is currently marked as a draft.  After completing the steps above, update its status by clicking "Ready for Review", or removing "WIP" from the title, as appropriate.




#### :radio_button: Let us know that your PR is ready for review:

### Who will review my changes?


This repository is currently maintained by `@axim-engineering`.  Tag them in a comment and let them know that your changes are ready for review.


### Where can I find more information?

### When can I expect my changes to be merged?

- The need for product review

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,334 @@
...
+            draft_version_pks=[None] * len(publishable_entities_pk),
+            published_version_pks=[None] * len(publishable_entities_pk),
+        )
+        container_version = ContainerEntityVersion.objects.create(
+            publishable_entity_version=publishable_entity_version,
+            container_id=container_pk,
+            defined_list=new_user_defined_list,
+            initial_list=last_version.initial_list,
+            frozen_list=last_version.defined_list,
+        )
```
mariajgrimaldi, line 236: @ormsbee: this is a naive approach to the implementation for the sake of simplicity. Now, I'll be modifying it to comply with the requirements I've compiled in the cover letter. Although, I still have some questions about the approach:

_The defined_list is the EntityList that the author defines. This should never change, even if the things it references get soft-deleted (because we'll need to maintain it for reverts)._

**Questions** 
1. How do we ensure this doesn't change even if references get soft deleted?
2. Is the discard operation (made changes that created a new draft version) different from revert?

_initial_list is an EntityList where all the versions are pinned, to show what the exact versions of the children were at the time that the Container was created. We could technically derive this, but it would be awkward to query._ 

**Questions**
1. Let's say a version for a container is created referencing the latest versions for all members. Then, we manually pin all members when creating this list. Pinning a version would mean pointing the EntityListRow draft/published version to the latest versions at the time for this particular case, am I right? 

 _frozen_list is mutable if and only if there are unpinned references in defined_list. In that case, frozen_list should start as None, and be updated to pin references when another version of this Container becomes the Draft version. But if this version ever becomes the Draft *again* (e.g. the user hits "discard changes" or some kind of revert happens), then we need to clear this back to None._

**Questions**
1. Even if there's an unpinned reference, should frozen_list start as None?
2. When a new version is created and frozen_list is None, then we should pin all references from the defined_list when creating the new frozen_list?
3. Should we explicitly check when a version becomes draft **again**, so we can clear the frozen list? How could we detect a version becomes a draft again after a revert?

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,334 @@
...
+
+    Returns:
+        The container with the given primary key.
+
+    TODO: should this use with_publishing_relations as in components?
+    """
+    return ContainerEntity.objects.get(pk=pk)
+
+
+def get_defined_list_for_container_version(
```
mariajgrimaldi, line 292: I wonder whether we should provide this kind of API method or whether users of the containers API should be unaware of EntityListRows' existence.

openedx_learning/apps/authoring/units/api.py
```diff
@@ -0,0 +1,208 @@
...
+        unit_pk: The unit ID.
+        title: The title.
+        publishable_entities_pk: The components.
+        entity: The entity.
+        created: The creation date.
+        created_by: The user who created the unit.
+    """
+    with atomic():
+        # TODO: how can we enforce that publishable entities must be components?
+        # This currently allows for any publishable entity to be added to a unit.
```
mariajgrimaldi, line 109: **Question**
Should we enforce the fact that publishable entities added to a unit are components here? Or should this be flexible enough to provide methods to do so but not explicitly enforce it?

openedx_learning/apps/authoring/units/api.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+        unit_version = UnitVersion.objects.create(
+            unit=unit,
+            container_entity_version=container_entity_version,
+            publishable_entity_version=container_entity_version.publishable_entity_version,
+        )
+    return unit_version
+
+
+def create_next_unit_version(
```
mariajgrimaldi, line 90: Regarding this requirement:

_When we pin to different versions of the Container, it indicates a change in how we've defined the Unit, as we explicitly pin a set of versions for the children and may later change to a different set._

This raises a couple of questions. Should we provide methods to pin children, which will then create new versions for the container? Or should we always check whether previous versions were pinned or not, to derive whether versions are now pinned?

openedx_learning/apps/authoring/units/api.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def get_user_defined_list_in_unit_version(unit_version_pk: int) -> QuerySet[EntityListRow]:
+    """Get the list in a unit version.
+
+    Args:
+        unit_version_pk: The unit version ID.
+    """
+    unit_version = UnitVersion.objects.get(pk=unit_version_pk)
+    return container_api.get_defined_list_for_container_version(unit_version.container_entity_version)
```
mariajgrimaldi, line 188: I have the same question here about whether to return entity lists or publishable entities. I'm guessing the latter, but I'd like to confirm.

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,334 @@
...
+            entity.pk,
+            version_num=version_num,
+            title=title,
+            created=created,
+            created_by=created_by,
+        )
+        # This implementation assumes:
+        # 1. We are creating the first version of the container, so the defined list is the same as the initial list.
+        # 2. The frozen list is empty because this is the first version.
+        # 3. Published and draft versions are always the latest for all members.
```
mariajgrimaldi, line 170: This is also a naive approach, this should also be compliant with the requirements I mentioned in the cover letter.

mariajgrimaldi:
Hey @ormsbee! I’ve been making progress on this and left some comments in the code where I could use some guidance. Thanks a ton for the help!

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,380 @@
...
+    return entity_list
+
+
+def create_container_version(
+    container_pk: int,
+    version_num: int,
+    title: str,
+    publishable_entities_pk: list[int],
+    draft_version_pks: list[int | None],
+    published_version_pks: list[int | None],
```
mariajgrimaldi, line 178: Should the draft/published versions be derived somehow from the publishable entities?

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,380 @@
...
+    entity_list = create_entity_list()
+    with atomic():
+        entity_list_rows = EntityListRow.objects.bulk_create(
+            [
+                EntityListRow(
+                    entity_list=entity_list,
+                    entity_id=row.entity.id,
+                    order_num=row.order_num,
+                    draft_version_id=row.entity.draft.version.pk,
+                    published_version_id=row.entity.published.version.pk,
```
mariajgrimaldi, line 163: Is this what pinning a version looks like?

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,380 @@
...
+            rows=get_defined_list_rows_for_container_version(last_version)
+        )
+
+        container_version = ContainerEntityVersion.objects.create(
+            publishable_entity_version=publishable_entity_version,
+            container_id=container_pk,
+            defined_list=new_user_defined_list,
+            initial_list=last_version.initial_list,
+            frozen_list=new_frozen_list,
+        )
```
mariajgrimaldi, line 277: Here, I'm creating a new defined entity list (with new rows) for each new version and also copying things over from the previously defined list into a new frozen list (also, with new rows). This is done to comply with:

1. defined_list should never change for a given ContainerEntityVersion to simplify reasoning.
2. The defined_list is the EntityList that the author defines. This should never change, even if the things it references get soft-deleted (because we'll need to maintain it for reverts).

But we also have this requirement:

3. Ordering should be treated as immutable–if the ordering needs to change, we create a new EntityList and copy things over.

So no new entity list rows for defined lists should be created if the ordering doesn't change or versions are not pinned, which I'm currently not considering. I remember discussing this with Dave a couple of weeks ago too, so implementing it seems like the right approach here.

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,380 @@
...
+        new_frozen_list = copy_rows_to_new_entity_list(
+            rows=get_defined_list_rows_for_container_version(last_version)
+        )
+
+        container_version = ContainerEntityVersion.objects.create(
+            publishable_entity_version=publishable_entity_version,
+            container_id=container_pk,
+            defined_list=new_user_defined_list,
+            initial_list=last_version.initial_list,
```
mariajgrimaldi, line 275:
```suggestion
            initial_list=new_user_defined_list, -> but pinned
```

openedx_learning/apps/authoring/containers/api.py
mariajgrimaldi: I'm testing my changes out before pushing, I'm hoping I can do it today.

openedx_learning/apps/authoring/units/api.py
```diff
@@ -0,0 +1,218 @@
...
+    return unit
+
+
+def create_unit_version(
+    unit: Unit,
+    version_num: int,
+    title: str,
+    publishable_entities_pks: list[int],
+    draft_version_pks: list[int | None],
+    published_version_pks: list[int | None],
```
bradenmacdonald, line 59: Why do we specify three different lists of the components in this new unit version? I would naively expect one list of components [and their versions?].

openedx_learning/apps/authoring/containers/models.py
```diff
@@ -0,0 +1,154 @@
...
+    #
+    # So our approach to this is to use a value of None (null) to represent an
+    # unpinned reference to a PublishableEntity. It's shorthand for "just use
+    # the latest draft or published version of this, as appropriate".
+    draft_version = models.ForeignKey(
+        PublishableEntityVersion,
+        on_delete=models.RESTRICT,
+        null=True,
+        related_name="draft_version",
+    )
```
bradenmacdonald, line 66: What is the use case for pinning the *draft* to a specific version, rather than just always saying that the "draft should use the latest draft version" (None) ?

openedx_learning/apps/authoring/containers/models.py
```diff
@@ -0,0 +1,154 @@
...
+    published_version = models.ForeignKey(
+        PublishableEntityVersion,
+        on_delete=models.RESTRICT,
+        null=True,
+        related_name="published_version",
+    )
```
bradenmacdonald, line 72: If publishable entities can be un-pinned, that means that the contents of a published Unit can change, if one of the components within it is published. Is that desired behavior in Studio and the LMS?

I know we'll need something like that or some kind of partial publish, because we don't want to have to publish the whole course every time a component changes (which would then require a unit to publish, which requires a subsection to publish, which requires a section to publish, which requires the course to publish) - or do we?

openedx_learning/apps/authoring/containers/models.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # This is the EntityList that's created when the next ContainerEntityVersion
+    # is created. All references in this list should be pinned, and it serves as
+    # "the last state the children were in for this version of the Container".
+    # If defined_list has only pinned references, this should point to the same
+    # EntityList as defined_list and initial_list.
+    #
+    # This value is mutable if and only if there are unpinned references in
+    # defined_list. In that case, frozen_list should start as None, and be
```
bradenmacdonald, line 143: > This value is mutable if and only if there are unpinned references in `defined_list`. 

In what circumstances do we ever have _pinned_ references in `defined_list`? From what I've seen of the UX designs so far, and everything I know about libraries at least, what we really want is a much simpler world where the draft versions of things in containers are always "unpinned". (And perhaps where the published versions are always pinned, but see my other question about whether or not we want that.)

mariajgrimaldi (commented):
I'm sorry it took me so long to get here, @bradenmacdonald. Thank you so much for the patience and the review! Should we close this in favor of #278?

openedx_learning/apps/authoring/units/api.py
bradenmacdonald: I have changed this in my PR.

openedx_learning/apps/authoring/containers/models.py
bradenmacdonald: Yeah, I think I understand this now.

bradenmacdonald:
@mariajgrimaldi If you're happy with the changes in #278, I think we can close this.

mariajgrimaldi:
Closed in favor of #278. @bradenmacdonald, thank you!

bradenmacdonald:
Thank _you_, @mariajgrimaldi !

Issue #237, opened 2024-10-02
Remove `learner_downloadable` field in `ComponentVersionContent`


ormsbee:
I originally added the `learner_downloadable` flag to `ComponentVersionContent` with the thought that it would reduce the odds of sensitive assets being made publicly downloadable. But in practice, it's just been a source of unnecessary complexity, and we're more or less ignoring it in libraries code that calls into openedx-learning. Instead, we're using the convention that any path starting with `static/` is meant to be downloadable–a convention that Learning Core can stay completely ignorant of.

This would involve removing the model field and all calls to it. However, public API functions could continue to accept the field, and log a warning message–that might make it easier to support minor fixes to Sumac over the next few weeks without branching.

PR #256, opened 2024-11-06
refactor!: remove learner_downloadable field

Ian2012:
This PR removes the `learner_downloadable` field.

Closes: #237


openedx_learning/apps/authoring/components/api.py
```diff
@@ -422,7 +420,7 @@ def create_component_version_content(
     content_id: int,
     /,
     key: str,
-    learner_downloadable: bool = False,
+    learner_downloadable: bool = None,
```
ormsbee, line 423: Okay, change in plan: Let's just kill the argument entirely, and then bump the version to 0.7.0.

ormsbee (changes requested):
Looks great! I have a minor request because a change in requirements on my part. Thank you!

openedx_learning/apps/authoring/components/api.py
Ian2012: Sure

ormsbee (changes requested):
Code looks good, but please bump the version to 0.17.0

PR #260, opened 2024-11-18
fix: make database function compatible with postgresql and mysql both

qasimgulzar:
## Summary

This pull request includes updates to the `StringAgg` class in `openedx_tagging/core/tagging/models/utils.py` to enhance its functionality and compatibility with different database backends. The most important changes include adding support for PostgreSQL, MySQL, and SQLite, and implementing abstract methods from the `Combinable` class.

Enhancements to `StringAgg` class:

* Added support for PostgreSQL by using `STRING_AGG` with a specified separator and ensuring expressions are cast to `TEXT`.
* Added support for MySQL and SQLite by using `GROUP_CONCAT` with a specified separator.
* Implemented abstract methods from the `Combinable` class to allow logical operations (`AND`, `OR`, `XOR`).

Additional changes:

* Imported `db_connection` from `django.db` to check the database backend.
* Updated the `StringAgg` class to accept a `delimiter` parameter and handle the `distinct` option and output type accordingly.

This PR is subsequent of https://github.com/openedx/edx-platform/pull/35762

### DIFF
--- a/openedx_tagging/core/tagging/models/utils.py
+++ b/openedx_tagging/core/tagging/models/utils.py
@@ -10,7 +10,7 @@ RESERVED_TAG_CHARS = [
            # e.g. lineage="Earth\tNorth America\tMexico\tMexico City"
     ' > ',  # Used in the search index and Instantsearch frontend to separate tag levels
             # e.g. tags_level3="Earth > North America > Mexico > Mexico City"
-    ';',   # Used in CSV exports to separate multiple tags from the same taxonomy…
+    ';',   # Used in CSV exports to separate multiple tags from the same taxonomy
            # e.g. languages-v1: en;es;fr
 ]
 TAGS_CSV_SEPARATOR = RESERVED_TAG_CHARS[2]


mphilbrick211:
Hi @qasimgulzar! Just flagging some failing checks for you.

qasimgulzar:
Thank you I will be looking into it.

qasimgulzar:
@mphilbrick211 please approve the actions

qasimgulzar:
I am going to take care of these broken tests

qasimgulzar:
It is on my radar for this week.

qasimgulzar:
@mphilbrick211 could you please approve the workflows

mphilbrick211:
All set, @qasimgulzar!

mphilbrick211:
@openedx/axim-engineering - this is ready for review. Thanks!

ormsbee (approved):
@pomegranited, @navinkarkera, @bradenmacdonald: Do any of you folks want to take a look at this before I merge?

openedx_tagging/core/tagging/models/utils.py
```diff
@@ -34,21 +35,56 @@ def as_sqlite(self, compiler, connection, **extra_context):
...
+        # Handle the distinct option and output type
+        distinct_str = 'DISTINCT ' if distinct else ''
+
+        extra.update({
+            'distinct': distinct_str,
+            'output_field': CharField(),
+        })
+
+        # Check the database backend (PostgreSQL, MySQL, or SQLite)
+        if 'postgresql' in db_connection.vendor.lower():
```
pomegranited, line 60: Thank you for this PR @qasimgulzar !

But I don't quite understand the approach -- why not subclass [`contrib.postgres.aggregates.StringAgg`](https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/aggregates/#stringagg) and override functionality for MySQL and SQLite, instead of reimplementing postgres functionality here?

I'm asking because I don't have a way to test this with postgres (and neither does our CI), but we can easily test mysql and sqlite.

qasimgulzar: I had thought that way but by adding a subclass it will require me to update api.py and explicitly check the driver to replace with PG or Mysql based of the database engine as I am doing here. Personally I feel this one a batter as but are the relational databases and same class is giving the abstraction.

However in both the cases we don't have way to independently test the change however there is a parent PR https://github.com/openedx/edx-platform/pull/35762 which I have used to test this change and it also have attached prove of testing.

pomegranited: @qasimgulzar Ah, I see why we can't subclass [contrib.postgres.aggregates.StringAgg](https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/aggregates/#stringagg) here -- that triggers a `ModuleNotFoundError: No module named 'psycopg2'`. 🤦‍♀️

openedx_tagging/core/tagging/models/utils.py
```diff
@@ -34,21 +35,56 @@ def as_sqlite(self, compiler, connection, **extra_context):
...
+            self.function = 'STRING_AGG'
+            self.template = '%(function)s(%(distinct)s%(expressions)s, %(delimiter)s)'
+            extra.update({"delimiter": delimiter})
```
pomegranited, line 63: If `delimiter` cannot be customized for MySQL or SQLite, then I don't think it should be exposed as an argument to the constructor here.

Can we just hardcode it to `','` for postgres?
```suggestion
            extra.update({"delimiter": ","})
```

pomegranited: And is it possible to avoid overriding `as_sql()` below by providing the [same `output_field` that postgres StringAgg does](https://docs.djangoproject.com/en/4.2/_modules/django/contrib/postgres/aggregates/general/#StringAgg)?

```suggestion
            extra.update({
                "delimiter": ",",
                "output_field": TextField(),    
            })
```

pomegranited (commented):
I'm not going to block merging this, so please consider my comments as suggestions :)

openedx_tagging/core/tagging/models/utils.py
qasimgulzar: @pomegranited I just updated the PR as per your recommendations, thank you for the great suggestion

pomegranited (approved):
:+1: Oh good, I'm glad my suggestions were helpful @qasimgulzar . Thank you for doing this, and all the other parts of this massive effort!

- [x] I tested this with MySQL under my usual devstack (not with Postgres or SQLite)
- [x] I read through the code
- [x] Includes documentation

openedx_tagging/core/tagging/models/utils.py
```diff
@@ -1,15 +1,16 @@
...
+from django.db.models import Aggregate, CharField, TextField
+from django.db.models.expressions import Combinable, Func
 
 RESERVED_TAG_CHARS = [
     '\t',  # Used in the database to separate tag levels in the "lineage" field
            # e.g. lineage="Earth\tNorth America\tMexico\tMexico City"
     ' > ',  # Used in the search index and Instantsearch frontend to separate tag levels
             # e.g. tags_level3="Earth > North America > Mexico > Mexico City"
-    ';',   # Used in CSV exports to separate multiple tags from the same taxonomy
+    ';',   # Used in CSV exports to separate multiple tags from the same taxonomy…
```
bradenmacdonald, line 13:
```suggestion
    ';',   # Used in CSV exports to separate multiple tags from the same taxonomy
```
Not sure why this ellipsis character was added here.

openedx_tagging/core/tagging/models/utils.py
```diff
@@ -34,21 +35,48 @@ def as_sqlite(self, compiler, connection, **extra_context):
...
-    but this version works with MySQL and SQLite.
+    This version supports PostgreSQL (STRING_AGG), MySQL (GROUP_CONCAT), and SQLite.
     """
+    # Default function is for MySQL (GROUP_CONCAT)
     function = 'GROUP_CONCAT'
     template = '%(function)s(%(distinct)s%(expressions)s)'
 
-    def __init__(self, expression, distinct=False, **extra):
+    def __init__(self, expression, distinct=False, delimiter=',', **extra):
+        self.delimiter = delimiter
```
bradenmacdonald, line 50: Is this used, or is it always `,` ? Need to either update line 55 below or remove this and update the docstring.

mphilbrick211:
@ormsbee are you able to merge this if it's all set?

Lord-KalEl:
🎉

mphilbrick211:
Bumping for when @ormsbee is back from vacation.

qasimgulzar:
@ormsbee can we merge this one, I need to push for edx-platform after this one

Lord-KalEl:
Hi, I'm glad to read this … but 1 question to this process:
How can I see, which release version it is? I use podman compose to deploy applications, but in this case, I dont find any version, release or something else …

I hope, it is not too offtopic …I'm sorry …

qasimgulzar:
> @ormsbee I don't think we should merge a function that accepts a `delimiter` parameter but doesn't use it, and instead always hard-codes `,`. But I could be missing something about how that is used. So I would like to see that changed or a comment added to explain before this merges. Otherwise no blockers.

@bradenmacdonald I can update function to use delimiter parameter instead of hardcoding it, if that resolves your concerns.

bradenmacdonald:
@qasimgulzar Yes that would resolve my concerns. Or remove the parameter, since we didn't have it before - I believe you added it but it's unused as far as I can tell.

bradenmacdonald:
Thanks @qasimgulzar. That resolves my concern. @ormsbee This can move ahead as long as someone else has reviewed/approved it - I don't have any concerns remaining nor plan to review any further.

ormsbee (approved):

Issue #34, opened 2023-02-22
API entity lookups


ormsbee:
When trying to work out an `api.py` for Components, I realized that there are an awkwardly large number of ways you can uniquely look up a Component. Namely:

* ID (primary key): Best for in-database references. Immutable.
* UUID (unique): Best for external references, like data entries in other systems. Also immutable. Possibly useful for import/export use cases across instances.
* LearningPackage + identifier: Most human readable thing, but _mutable_, and not good for storing references to, but could make for pretty URLs and sensible export files.

And that last use case gets weird, because there are multiple ways to look up a LearningPackage. I started to get these api functions that were just really lookup variants. It made me think of just going straight to models, but then I was wondering if we could try to hide some of that behind some typing conventions. So say we had something that looked like this:

```python
from typing import Union, Tuple
from uuid import UUID

LearningPackageIdLookup = int
LearningPackageUUIDLookup = UUID
LearningPackageIdentifierLookup = str
LearningPackageLookup = Union[
    LearningPackageIdLookup,
    LearningPackageUUIDLookup,
    LearningPackageIdentifierLookup
]

ComponentIdLookup = int
ComponentUUIDLookup = UUID
ComponentIdentifierLookup = Tuple[LearningPackageLookup, str]
ComponentLookup = Union[
    ComponentIdLookup,
    ComponentUUIDLookup,
    ComponentIdentifierLookup,
]

ComponentVersionIdLookup = int
ComponentVersionUUIDLookup = UUID
ComponentVersionNumLookup = Tuple[ComponentLookup, int]
ComponentVersionLookup = Union[
    ComponentVersionIdLookup,
    ComponentVersionUUIDLookup,
    ComponentVersionNumLookup,
]


def get_component_version(lookup: ComponentVersionLookup) -> ComponentVersionData:
    # ...

```

The inside of that would be a number of `isinstance` calls to figure out what queries need to be fed to the ORM to get what we want (we can't use `match` until Python 3.10).

So people would call it like:

```python
get_component_version(20)

get_component_version(some_uuid)

get_component_version(ComponentVersionNumLookup(component_lookup, 4))
```

I haven't really seen this pattern before though. Do you think it'd be too confusing?

FYI: @bradenmacdonald, @kdmccormick, @feanil

bradenmacdonald:
In general, I'm fine with any approach as long as we define the types very clearly like you're doing here!

[...]

PR #278, opened 2025-02-14
Understanding containers/units: High level APIs, high level test cases, open questions

bradenmacdonald:
1. Do we actually want methods like `get_defined_list_rows_for_container_version`/`get_initial_list_rows_for_container_version`/`get_frozen_list_rows_for_container_version` in the public API? I would consider those implementation details. For this PR I've implemented higher level functions that make more sense to me based on use cases: get the entities in the draft version of the container, and get the entities in the published version of the container.

2. Check out the new test cases. Is the high-level reasoning/expectations correct so far?

Open questions:

3. UX specs say modifying a child will mark the unit as "having unpublished changes" but ADR says otherwise. Does it make sense to implement a container_has_unpublished_changes() helper function that recursively checks for unpublished changes in a container (and its child containers, and so on...), so we can display this in the UI?  [See slack discussion](https://openedx.slack.com/archives/C05NT2YN820/p1738708639728279)

4. Do I understand correctly then that the version history of units should NOT be thought of as a series of snapshots of the unit (as I had been thinking) but rather a changelog of edits made to the unit itself (but ignoring most changes made its the children)?
   - If so, To get a snapshot of the unit at a point in time, it's necessary to pass in a timestamp. Should we have an API to get an old version of a unit by passing in a timestamp?

5. Is it correct that units allow you to pin the version of a component, which means you always get exactly that component, but sequences/subsections will not really support pinning in the same way, because even if you pin a unit to a specific version, the unpinned components in that unit can still be updated?
   - If so, is there any value in even implementing "pinning" for containers larger than units?

6. Do we have any plans to implement pinned/unpinned components in the UI or is everything unpinned?

7. If we have unpinned published versions, why do we need the various defined_list, initial_list, frozen_list etc? So far in the test cases I've been writing I haven't found a need to even reference `initial_list` nor `frozen_list`.

8. Why do we specify `draft_version_pks` and `published_version_pks` when creating a new draft version of a unit? Seems wrong to say anything at all about the published version when it doesn't yet exist and may never exist.

9. Note to self: if soft deleting a component doesn't mark the unit as changed, we need to update the UI to show soft deleted draft components and allow publishing the deletion. Or maybe this is handled by a `container_has_unpublished_changes` and recursivelying pseudo-publishing the container (no changes to container but force publish the children including soft deletes).

>    # This value is mutable if and only if there are unpinned references in
>    # defined_list. In that case, frozen_list should start as None, and be
>    # updated to pin references when another version of this Container becomes
>    # the Draft version. But if this version ever becomes the Draft *again*
>    # (e.g. the user hits "discard changes" or some kind of revert happens),
>    # then we need to clear this back to None.

10. Based on the above, frozen list doesn't make a lot of sense, because the versions of the container don't capture many of the changes to the contents. e.g. container version 1 had component version 1 through 50, and container version 2 had component versions 51 through 100, what is the point of tracking whether container version 1 "should" show v1 or v50 when they're wildly different?


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

ormsbee:
> 3. UX specs say modifying a child will mark the unit as "having unpublished changes" but ADR says otherwise. Does it make sense to implement a container_has_unpublished_changes() helper function that recursively checks for unpublished changes in a container (and its child containers, and so on...), so we can display this in the UI? [See slack discussion](https://openedx.slack.com/archives/C05NT2YN820/p1738708639728279)

I think we'll want some bookkeeping models in the `containers` app that tracks when a new child version is created and recursively marks its ancestors "dirty". Similarly, I think we'll want to have one that tracks "this publish affected these ancestors, even if they haven't otherwise changed". I haven't worked out any details beyond the vague thoughts that:

* We need to keep it really lightweight.
* It might be useful to have a major/minor version numbering of that for containers, in the sense that major number tracks changes to the container's metadata, ordering, and membership; and the minor number tracks changes to the contents. So 1.0, 1.1, 1.7, 2.0, 2.1, etc.
* I used to think that we had to track all containers that used a particular entity and then auto-generate new draft versions of those containers with the entity missing if the entity was soft-deleted... and then also auto-publish those containers when te entity's deletion was published. But now I'm thinking that we do that where we know it makes sense (like in the originating Unit where we're editing it), but we otherwise are tolerant of when unpinned children disappear like that.

ormsbee:
> 8. Why do we specify draft_version_pks and published_version_pks when creating a new draft version of a unit? Seems wrong to say anything at all about the published version when it doesn't yet exist and may never exist.

Yeah, I feel like we had this discussion at some point and came to the conclusion that it was because my original sketch for `EntityListRow` had `draft_version` and `published_version`, but really it should have just had a single `version` field.

ormsbee:
> 7. If we have unpinned published versions, why do we need the various defined_list, initial_list, frozen_list etc? So far in the test cases I've been writing I haven't found a need to even reference initial_list nor frozen_list.

I had originally thought that these would be useful for history purposes, but I think it's not necessary if we have minor version tracking.

I do wonder if the relationship between containers and versioning is basic enough where `containers` should be folded directly into the `publishing` app (with specific container-types still having their own apps).

ormsbee:
Though I guess one very confusing part about having a major and minor version for containers is that it's possible to publish any particular subset of the children, so it's not like there's really a published "1.12" version–"1.12" just means "twelve changes have happened to some descendent". So it'd be more like we'd be running two minor counters for children–one for draft changes, one for publishes.

bradenmacdonald:
> I think we'll want some bookkeeping models in the containers app that track when a new child version is created and recursively marks its ancestors "dirty".

For now, I implemented a `contains_unpublished_changes()` that recursively scans a container for changes; it uses a fixed 3 queries to determine if a unit has changes, regardless of how many components are in the unit. This should be enough to write the test cases for now and confirm correctness, but will need to be replaced by a tracking-based approach or something like a [closure table](https://www.slideshare.net/billkarwin/sql-antipatterns-strike-back?src=embed#68) for all the containers in each learning package. (Presumably only for drafts or possibly separate closure tables for draft and published trees.) I kind of like the closure table idea more than a tracking-based approach, as long as its performant enough - it would let us optimize so many different things.

openedx_learning/apps/authoring/units/api.py
```diff
@@ -188,31 +194,51 @@ def get_latest_unit_version(unit_pk: int) -> UnitVersion:
...
+    @property
+    def component(self):
+        return self.component_version.component
 
-def get_frozen_list_in_unit_version(unit_version_pk: int) -> list[int]:
-    """Get the frozen list in a unit version.
 
-    Args:
-        unit_version_pk: The unit version ID.
+def get_components_in_draft_unit(
```
bradenmacdonald, line 210: It feels a bit suboptimal to me that we have to have separate APIs for `get_components_in_draft_unit` and `get_components_in_published_unit`. This is because when things are unpinned, we need to know whether to use the latest draft or latest published version.

I don't think this is a big deal, but it does "feel" to me like it goes against the spirit of the publishing app, where you have a series of versions and draft/published are just pointers to a particular version. With components (and presumably most PublishableEntities in general other than our new containers), if you ask for "version 37", you get a specific thing regardless of whether it has/was been a draft or published. Whereas in this new case, you could try to request a particular version of a unit, but unless we know whether you want it as a draft or you want it as published, we usually can't fully describe that particular version's contents.

-------

That is also why my proposed `get_components_in_published_unit` API doesn't accept a version number; it can't really get anything other than the current published version. (The unit version is insufficient to specify a snapshot of the unit when things are unpinned.)

That said, there likely will still be a way to get snapshots of any historic _published_ version you want, analogous to requesting a specific component version, but it will require asking for the state of the published unit when the entire LearningPackage was at a specific version, [which is given by the PublishLog ID](https://github.com/openedx/openedx-learning/blob/9b75ba4ed9383d58a1a7af59f118664fa51afac2/openedx_learning/apps/authoring/publishing/models.py#L399-L406).

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -99,41 +108,35 @@ def create_next_defined_list(
...
         new_rows = []
-        for order_num, entity_pk, draft_version_pk, published_version_pk in zip(
-            order_nums, entity_pks, draft_version_pks, published_version_pks
+        for order_num, entity_pk, entity_version_pk in zip(
+            order_nums, entity_pks, entity_version_pks
         ):
-            row = publishable_entities_in_rows.get(entity_pk)
-            if row and row.order_num == order_num:
-                new_entity_list.entitylistrow_set.add(row)
-                continue
```
bradenmacdonald, line 111: This had a bug - the code was _moving_ `EntityListRow` objects into the new `EntityList` rather than copying them into it.

bradenmacdonald:
@ormsbee @kdmccormick CC @mariajgrimaldi @pomegranited Could I please get time from you next week to review this PR? The earlier in the week the better. Or let me know if there's some other way you'd like to land this prototype - it doesn't _have_ to be through this big PR.

bradenmacdonald:
I'd also like to squash this whole thing down to one or two commits by the end of this week, before y'all start reviewing. But I could also open a new PR if you want me to keep the entire history.

openedx_learning/apps/authoring/containers/models.py
```diff
@@ -0,0 +1,111 @@
...
+        related_name="+",  # Do we need the reverse relation?
+    )
+
+
+class ContainerEntity(PublishableEntityMixin):
+    """
+    NOTE: We're going to want to eventually have some association between the
+    PublishLog and Containers that were affected in a publish because their
+    child elements were published.
+    """
```
kdmccormick, line 75: Can you add some words to this docstring about what a ContainerEntity is, like the other models have?

kdmccormick (changes requested):

openedx_learning/apps/authoring/units/models.py
```diff
@@ -0,0 +1,31 @@
...
+__all__ = [
+    "Unit",
+    "UnitVersion",
+]
+
+
+class Unit(ContainerEntityMixin):
+    """
+    A Unit is Container, which is a PublishableEntity.
+    """
```
bradenmacdonald, line 17: One concern that I have and definitely want to flag is the inheritance pattern of `Unit`:

```
class ContainerEntity(PublishableEntityMixin):
    ...

class ContainerEntityMixin(PublishableEntityMixin):
    container_entity = models.OneToOneField(ContainerEntity)
    ...

class Unit(ContainerEntityMixin):
    ...
```

What this means (I think) is that `PublishableEntityMixin` is used separately twice, so in theory, you have two different `PublishableEntity` records associated with each unit: `unit.publishable_entity` and `unit.container_entity.publishable_entity`. The only reason this isn't causing bugs in the prototype is that [the `create_unit` API method explicitly sets them to be the same](https://github.com/open-craft/openedx-learning/blob/e2fe520aa0e36275ef41c841b3ba41ccdf2c6c2e/openedx_learning/apps/authoring/units/api.py#L49-L51).

But I'm wondering if it would be simpler to just remove `PublishableEntityMixin` from either `ContainerEntity` or `ContainerEntityMixin`.

I think the `ContainerEntityVersion` family of models has the same issue.

openedx_learning/apps/authoring/containers/models.py
```diff
@@ -0,0 +1,111 @@
...
+    # the latest draft or published version of this, as appropriate".
+    entity_version = models.ForeignKey(
+        PublishableEntityVersion,
+        on_delete=models.RESTRICT,
+        null=True,
+        related_name="+",  # Do we need the reverse relation?
+    )
+
+
+class ContainerEntity(PublishableEntityMixin):
```
bradenmacdonald, line 70: Are we set on the name `ContainerEntity` and `ContainerEntityVersion`? I think I would prefer the simpler `Container` and `ContainerVersion`, which are also more analogous to `Component` and `ComponentVersion`. We don't call them `ComponentEntity`.

kdmccormick: +1

openedx_learning/apps/authoring/containers/models.py
kdmccormick: This still applies to the Container class ^

openedx_learning/apps/authoring/units/models.py
kdmccormick: I'm glad you realized this.

> But I'm wondering if it would be simpler to just remove PublishableEntityMixin from either ContainerEntity or ContainerEntityMixin.

That is my sense too. We would end up with some extra verbosity, like:

```
unit_version_num = unit.container.versioning.draft.version_num
```
instead of:
```
unit_version = unit.versioning.draft.version_num
```

but I think that's better than risking inconsistency.

bradenmacdonald: I tried this out here: https://github.com/open-craft/openedx-learning/pull/12/files

I thought it might make things cleaner but I actually found it a bit harder to work with. It does make dealing with the versioning in the test cases a little more verbose, and I think it's also kind of confusing that there's a `UnitVersion` that doesn't have `.versioning` info but only does via `unitversion.containerversion.versioning`. It does work though, and the diff is fairly small either way.

openedx_learning/apps/authoring/containers/models.py
bradenmacdonald: Added

bradenmacdonald:
In a separate PR I tried moving the `containers` models into `publishing`. Take a look if you're interested: https://github.com/open-craft/openedx-learning/pull/13

openedx_learning/apps/authoring/units/api.py
pomegranited: What's the use case for a Container that holds both pinned and unpinned Entries? I can see why we'd want them all pinned or all unpinned, but not the mixed case.

But regardless, it's possible to combine both `get_entities_in_draft_container` and `get_entities_in_published_container` into a single method with a `published` parameter, which looks a little awkward in the code, but would make the components and unit APIs simpler.

```python
def get_entities_in_container(
    container: Container | ContainerMixin,
    published: bool = False,
) -> list[ContainerEntityListEntry] | None:
    """
    [ 🛑 UNSTABLE ]
    Get the list of entities and their versions in the draft or published
    version of the given container.
    """
    if isinstance(container, ContainerMixin):
        container = container.container
    if isinstance(container, Container):
        cv = container.versioning.published if published else container.versioning.draft
    else:
        raise TypeError(f"Expected Container or ContainerMixin; got {type(container)}")

    if published and cv is None:
        return None  # There is no published version of this container. Should this be an exception?

    assert isinstance(cv, ContainerVersion)
    entity_list = []
    for row in cv.entity_list.entitylistrow_set.order_by("order_num"):
        entity_version = row.entity_version or (
            row.entity.published.version if published else row.entity.draft.version
        )
        if entity_version is not None:  # As long as this hasn't been soft-deleted:
            entity_list.append(ContainerEntityListEntry(
                entity_version=entity_version,
                pinned=row.entity_version is not None,
        ))
        # else should we indicate somehow a deleted item was here?

    return entity_list
```

tests/openedx_learning/apps/authoring/units/test_api.py
```diff
@@ -0,0 +1,868 @@
...
+
+        assert authoring_api.contains_unpublished_changes(unit)
+        assert self.component_1.versioning.published is None
+        assert self.component_2.versioning.published is None
+
+        # Publish ONLY the unit. This should however also auto-publish components 1 & 2 since they're children
+        authoring_api.publish_from_drafts(
+            self.learning_package.pk,
+            draft_qset=authoring_api.get_all_drafts(self.learning_package.pk).filter(entity=unit.publishable_entity),
+        )
```
pomegranited, line 255: Do we need an API method to encapsulate this `publish_container_and_children` logic?

pomegranited (approved):
:+1: This looks great, and your test cases are spectacular.

I left a bunch of nits, and one comment about combining the `get_entities_in_*_container` methods into a single method, but that's up for debate.

- [x] <strike>I tested this</strike> -- I didn't run this code, but I checked the test cases and coverage carefully.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues by using my keyboard to navigate</strike> N/A
- [x] Includes documentation
- [x] User-facing strings are extracted for translation

tests/openedx_learning/apps/authoring/units/test_api.py
bradenmacdonald: TBD. But we don't really have a use case for publishing a unit without publishing its children, so I was thinking it would just be the same publish API for everything.

openedx_learning/apps/authoring/units/api.py
```diff
@@ -0,0 +1,300 @@
...
+            publishable_entity_version=container_version.publishable_entity_version,
+        )
+    return unit_version
+
+
+def create_next_unit_version(
+    unit: Unit,
+    *,
+    title: str | None = None,
+    components: list[Component | ComponentVersion] | None = None,
```
bradenmacdonald, line 100: Something else that I wanted to call out is that many Learning Core APIs take integer IDs as their parameters but in several of these new APIs I'm taking `Unit` and `Component | ComponentVersion` objects. I think this makes the API much nicer (we can use types to distinguish between `Component` and `ComponentVersion` to determine if things should be pinned or not) and more efficient, but arguably it could be more consistent to only take integer IDs.

tests/openedx_learning/apps/authoring/units/test_api.py
bradenmacdonald: I have a proof of concept here that shows what it could look like if we modify the normal publish API to auto-publish children, rather than making it a separate API: https://github.com/open-craft/openedx-learning/pull/13

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,451 @@
...
+    assert isinstance(cv, ContainerVersion)
+    entity_list = []
+    for row in cv.entity_list.entitylistrow_set.order_by("order_num"):
+        entity_version = row.entity_version or row.entity.published.version
+        if entity_version is not None:  # As long as this hasn't been soft-deleted:
+            entity_list.append(ContainerEntityListEntry(
+                entity_version=entity_version,
+                pinned=row.entity_version is not None,
+            ))
+        # else should we indicate somehow a deleted item was here?
```
mariajgrimaldi, line 361: Maybe `ContainerEntityListEntry` could handle when the entry was previously soft-deleted?

openedx_learning/apps/authoring/units/api.py
bradenmacdonald: > What's the use case for a Container that holds both pinned and unpinned Entries? I can see why we'd want them all pinned or all unpinned, but not the mixed case.

We don't actually have a use case for pinned entities at all yet, so I'm a bit worried that we won't need it at all in the future, and this is needlessly complicated. But there is a strong possibility that we'll need pinned entities to support CCX-style use cases and/or dynamic containers (problem banks, split test, randomization, adaptive learning). We haven't really got a good handle yet on how either will work, which is why all these APIs are being marked "unstable".

bradenmacdonald: Also, I'm open to combining these into one function, but I think we should always require an explicit `published=True` or `published=False` in that case, because Learning Core will be used "equally"? for CMS and LMS use cases and it's not correct to assume that either draft or published is the default.

So I'd like to hear what others think, and also about whether `container: Container | ContainerMixin,` is a nice way to handle the parameter or if that needs to be changed to `container_id: int` for consistency with other Learning Core APIs. The other `authoring` APIs only accept integer IDs or QuerySets as their parameters.

openedx_learning/apps/authoring/containers/api.py
bradenmacdonald: Yes that's what I was thinking, but we don't have any use case for it yet.

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,451 @@
...
+    """
+    if ignore_pinned:
+        qs = Container.objects.filter(
+            publishable_entity__draft__version__containerversion__entity_list__entitylistrow__entity_id=publishable_entity_pk,  # pylint: disable=line-too-long # noqa: E501
+            publishable_entity__draft__version__containerversion__entity_list__entitylistrow__entity_version_id=None,  # pylint: disable=line-too-long # noqa: E501
+        ).order_by("pk")  # Ordering is mostly for consistent test cases.
+    else:
+        qs = Container.objects.filter(
+            publishable_entity__draft__version__containerversion__entity_list__entitylistrow__entity_id=publishable_entity_pk,  # pylint: disable=line-too-long # noqa: E501
+        ).order_by("pk")  # Ordering is mostly for consistent test cases.
```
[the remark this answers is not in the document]
bradenmacdonald: Yes, it really is 😅

> Might be clearer if you build it like this:

I can't do that, because `filter(some__related__thing=A, some__related_other=B)` behaves differently than `filter(some__related__thing=A).filter(some__related_other=B)` - specifically that in the first case A & B must be on the same related object, and in the latter they can be on different related objects.

I thought maybe I was confusing things or mis-remembering, but I realized that I should actually improve the test case to verify what I'm saying, and I've done that in 84f658de5581c358b28226fe266531f6fa080fc3. Now the test case is much better, and I found that (unfortunately) if I try the refactor you're suggesting, the updated test fails, showing that it's not an equivalent query.

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,451 @@
...
+
+def contains_unpublished_changes(
+    container: Container | ContainerMixin,
+) -> bool:
+    """
+    [ 🛑 UNSTABLE ]
+    Check recursively if a container has any unpublished changes.
+
+    Note: container.versioning.has_unpublished_changes only checks if the container
+    itself has unpublished changes, not if its contents do.
```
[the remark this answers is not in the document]
bradenmacdonald: Updated in 62d6d1b4b1643dd830d3ac8fddf78bbe22c2d899 - please tell me if that makes sense now or not.

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,451 @@
...
+            child_container = row.entity.container
+            # This is itself a container - check recursively:
+            if child_container.versioning.has_unpublished_changes or contains_unpublished_changes(child_container):
+                return True
+        else:
+            # This is not a container:
+            draft_pk = row.entity.draft.version_id if row.entity.draft else None
+            published_pk = row.entity.published.version_id if hasattr(row.entity, "published") else None
+            if draft_pk != published_pk:
+                return True
```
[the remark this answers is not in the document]
bradenmacdonald: Good question, but it's because `row.entity` is a `PublishableEntity`, and not a class with `PublishableEntityMixin`. So, rather inconveniently, we cannot access the `versioning` helper in general.

If we knew we were only dealing with Components, say, and not any other type of entity, we could do `row.entity.component.versioning.has_unpublished_changes` but there's no such convenience property/method available on generic `PublishableEntity` instances. The whole entity/mixin design could be refactored to improve this, but that's obviously way out of scope here.

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,451 @@
...
+    itself has unpublished changes, not if its contents do.
+    """
+    if isinstance(container, ContainerMixin):
+        # The query below pre-loads the data we need but is otherwise the same thing as:
+        #     container = container.container
+        container = Container.objects.select_related(
+            "publishable_entity",
+            "publishable_entity__draft",
+            "publishable_entity__draft__version",
+            "publishable_entity__draft__version__containerversion__entity_list",
```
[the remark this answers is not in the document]
bradenmacdonald: I think it's already being preloaded, because when I add `publishable_entity__published__version` here, it has no effect on the query count in the test cases. It's probably pre-loaded via `Container.objects` which is already configured to pre-select a bunch of stuff.

I did realize I can simplify this here, because the first three lines like `publishable_entity ` are redundant to specify when we're also specifying `publishable_entity__draft__version__containerversion__entity_list`

openedx_learning/apps/authoring/containers/api.py
```diff
@@ -0,0 +1,451 @@
...
+    [ 🛑 UNSTABLE ]
+    Get a container by its primary key.
+
+    Args:
+        pk: The primary key of the container.
+
+    Returns:
+        The container with the given primary key.
+    """
+    # TODO: should this use with_publishing_relations as in components?
```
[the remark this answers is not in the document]
bradenmacdonald: Updated and added a test case for it too.

openedx_learning/apps/authoring/units/api.py
bradenmacdonald: @ormsbee could you weigh in please?

ormsbee: I think it was mostly that there were places where I could get by with just an ID, like `create_component_version`, and then I kept things the same for consistency.

We discussed this a bit back in https://github.com/openedx/openedx-learning/issues/34. My last comment at the time was:

> For now, the API mostly uses primary keys internally, except some functions that are explicitly labeled as using keys. I'm closing this for now, with the understanding that we might revisit it if when/if the current conventions become a problem.

So yeah, I'm not wedded to it.

openedx_learning/apps/authoring/containers/api.py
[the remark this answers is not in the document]
pomegranited: > I can't do that, because filter(some__related__thing=A, some__related_other=B) behaves differently than filter(some__related__thing=A).filter(some__related_other=B) - specifically that in the first case A & B must be on the same related object, and in the latter they can be on different related objects.

Ahh... I did not know that! Thank you for the explanation.

> Now the test case is much better, and I found that (unfortunately) if I try the refactor you're suggesting, the updated test fails, showing that it's not an equivalent query.

The succeeding test was my justification for making this comment, so thank you for closing that loophole.

openedx_learning/apps/authoring/units/models.py
bradenmacdonald: @ormsbee This is something else I'd really like to get your opinion on ^

bradenmacdonald:
@ormsbee @kdmccormick CC @mariajgrimaldi @pomegranited 

This should be ready for final-ish review right now. I'm pretty happy with it.

I have just pushed some changes:
1. I moved the `containers` models into the `publishing` app as discussed. I split the models and mixins into separate python files to avoid some circular imports and keep the file size small. There are no changes to any of the existing models like `LearningPackage` or `PublishableEntity` but they'll be showing up in the diff since they're moved. (If you want me to open a PR to make that refactor to the existing `publishing` app first, it I can do that - it would make this diff smaller.) 
2. I updated `publish_from_drafts` so it will auto-publish children. I'm still not 100% sure this is what we want, but I think it makes sense, and it lets me un-skip the relevant test that I had written before, which now passes.

Unresolved discussions from the review that I'd like to point out to make sure they're well considered:
1. Check the function signatures of all the new API methods. I think they're nice, but there have been some discussions around whether they should take integer PKs instead of model instances, and whether `get_components_in_draft_unit` and `get_components_in_published_unit` should be combined into `get_components_in_unit(x, published=T/F)` for example.
2. Auto-publishing children requires us to do one additional query every time something is published (to check if it has children), even if we're not publishing a container. I think that's fine, but just wanted to note it.
3. The publishing app has been re-organized and is more prone to circular imports now, though I've managed to avoid introducing any.
4. The new `get_components_in_draft_unit` method and its related `_container` version return list of a `@dataclass(frozen=True) class UnitListEntry` objects. I like this `dataclass` pattern but it's new to Learning Core.

tests/openedx_learning/apps/authoring/units/test_api.py
kdmccormick: To avoid recreating the crushingly-complex flexibility of ModuleStore, I think we should feel good about coding  some opinionated decisions into the API, based on what product is telling us about user expectations for Content Libraries. "Publishes always cascade" seems like one of those opinionated decisions we can make. It would save API users the trouble of having to think about whether or not they should use the cascading or non-cascading publish, and I imagine it'd be a nice simplifying assumption for ourselves as we continue to expand the API.

If it does become necessary in the future to do non-cascading publishes, we could stick an optional boolean into the API function:
```python
def publish_from_drafts(
    learning_package_id: int,  # LearningPackage.id
    /,
    draft_qset: QuerySet,
    message: str = "",
    published_at: datetime | None = None,
    published_by: int | None = None,  # User.id
    cascade: bool = True,  # <--- new flag, added if & when we need it
) -> PublishLog:
```

ormsbee: Agreed. Not cascading the publish can also lead to some weird and unintuitive situations. For instance, if you create a new child and only publish the container without cascading, then the newly published container would be missing the element you just added. It's a foot-gun that should work hard to get consideration if the use case ever arrives.

bradenmacdonald: Good points! I agree too. I have already included auto-publishing in this PR as of yesterday.

openedx_learning/apps/authoring/units/models.py
ormsbee: The following is a half-baked idea... but given how empty our Unit and UnitVersion models are, does it maybe make sense to make them proxy models on top of Container and ContainerVersion? So we'd add a type field in Container/ContainerVersion and use that for switching logic:

```python
class Unit(Container):
    objects = UnitManager()  # auto-filters on the type, could make this part of a mixin

    # the above could also look something like:
    objects = ContainerManager(ContainerType.UNIT)  # whatever we use to differentiate

    class Meta:
        proxy = True
```

Then putting Unit-specific metadata at some future point comes as a 1:1 model that hangs off of this (which the understanding that it might never happen).

This would probably remove a number of concrete models for various containers and also make it so that we can still do `unit_version.versioning...`

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -310,6 +332,30 @@ def publish_from_drafts(
...
+        container_version_ids = (
+            Container.objects.filter(publishable_entity__draft__in=draft_qset)
+            .values_list("publishable_entity__draft__version__containerversion__pk", flat=True)
+        )
+        if container_version_ids:
+            # We are publishing at least one container. Check if it has any child components that aren't already slated
+            # to be published.
+            unpublished_draft_children = EntityListRow.objects.filter(
+                entity_list__container_versions__pk__in=container_version_ids,
+                entity_version=None,  # Unpinned entities only
```
ormsbee, line 348: I have this weird recollection that it was necessary to do something like `entity_version__isnull=True` rather than `entity_version=None` here, but that doesn't seem to be the case based on some tinkering I just did on the console. I don't know if I'm just completely misremembering, or if Django behavior has changed, or else if there was some weird edge case. No change needed here, but I'm writing this in case it triggers any bells for you.

bradenmacdonald: I have a similar vague recollection but I think maybe we're just thinking of how you have to write the queries when using raw SQL, where `= NULL` is always false but `IS NULL` gives the correct result. According to [this](https://stackoverflow.com/questions/29858701/django-query-isnull-true-or-none) they're mostly the same in Django except for some one particular case involving PostgreSQL JSON field comparisons.

openedx_learning/apps/authoring/units/models.py
bradenmacdonald: I also like those ideas.

> Then putting Unit-specific metadata at some future point comes as a 1:1 model that hangs off of this (which the understanding that it might never happen).

This pattern still has an unsolved problem, which is that the base containers code wouldn't be aware of the new `UnitMetadata` 1:1 model. We'd need some mechanism to "register" the `UnitMetadata` model so that it gets duplicated when a new version is created, and included in imports/exports. Personally I lean toward just storing most metadata in a `JSONField` on `ContainerVersion` - nice and simple, but doesn't support foreign keys. Alternatively, we could also create a registry of these 1:1 hang-on models, which would mostly work well with some limitations (e.g. if you imported a course without the corresponding plugin installed and then later installed the plugin, the data would be missing, because there'd have been no way to import the data without the plugin installed). It's not something we necessarily need to solve now though; just writing it down for future consideration. We also [discussed](https://openedx.slack.com/archives/C05NT2YN820/p1740677547208409?thread_ts=1740621421.055679&cid=C05NT2YN820) this on Slack.

openedx_learning/apps/authoring/publishing/api.py
ormsbee: Oh! Now I remember. This is what I was thinking about:

https://github.com/openedx/openedx-learning/blob/4316b5cb45faa974a058716d2d5edb0658b7dd24/openedx_learning/apps/authoring/publishing/api.py#L237-L247

(So not a problem with this piece of code.)

openedx_learning/apps/authoring/units/models.py
kdmccormick: generic foreign keys?

```python
class Container(Model):
    ...
    metadata_type = models.ForeignKey(ContentType, on_delete=models.RESTRICT)
    metadata_id = models.PositiveIntegerField()
    metadata = GenericForeignKey('metadata_type', 'metadata_id')
    ...

units = Container.objects.filter(metadata_type=UnitMetadata)
sequence = Container.objects.filter(metadata_type=SequenceMetadata)

my_unit_metadata_baz = units.get(foo=bar).metadata.baz  # <- note: mypy might complain about `.baz`
```

additionally, we could still use the proxy pattern in order to make a nicer API:
```python
class Unit(Container):
    objects: Manager[Self] = ... # filter `objects` down to `metadata_type=UnitMetadata`
    metadata: UnitMetadata
    
    class Meta:
        proxy = True

my_unit_metadata_baz = Unit.objects.get(foo=bar).metadata.baz   # <-- mypy is happy
```

kdmccormick: Note: That ^ assumes that UnitMetadata is version-agnostic. I can imagine a similar pattern working for  UnitVersionMetadata, though.

In either case, we wouldn't actually need to duplicate UnitMetadata and UnitVersionMetadata for every version--we'd create a new metadata row only when the metadata changes. Similar to EntityList.

kdmccormick:
> The publishing app has been re-organized and is more prone to circular imports now, though I've managed to avoid introducing any.
> The new get_components_in_draft_unit method and its related _container version return list of a @dataclass(frozen=True) class UnitListEntry objects. I like this dataclass pattern but it's new to Learning Core.

@bradenmacdonald These both look good to me. Frozen dataclasses are fantastic, happy to see them in LC.

openedx_learning/apps/authoring/units/models.py
bradenmacdonald: > In either case, we wouldn't actually need to duplicate UnitMetadata and UnitVersionMetadata for every version--we'd create a new metadata row only when the metadata changes. Similar to EntityList.

Good point, I didn't really think about that.

Your generic foreign key approach has a couple limitations - first it assumes there's only one metadata model, but if we support plugins we may have several metadata models. Second, keys are not enforced by the database, so I don't see what it gains compared to the JSON field approach.

bradenmacdonald: > The following is a half-baked idea... but given how empty our Unit and UnitVersion models are, does it maybe make sense to make them proxy models on top of Container and ContainerVersion? So we'd add a type field in Container/ContainerVersion and use that for switching logic:

I tried this out: https://github.com/open-craft/openedx-learning/pull/14

I think I like it better, but proxy models do have more limitations and need to be used more carefully. Anyhow check it out and let me know what you guys think.

openedx_learning/apps/authoring/units/api.py
pomegranited: > I think we should always require an explicit published=True or published=False in that case, because Learning Core will be used "equally"? for CMS and LMS use cases and it's not correct to assume that either draft or published is the default.

I think it's worth combining these methods into one. 👍 on explicit `published` param.

> whether container: Container | ContainerMixin, is a nice way to handle the parameter or if that needs to be changed to container_id: int for consistency with other Learning Core APIs. 

I think it's fine to leave these parameters as objects. Thinking the perspective of the platform REST API:  we'll get a UsageKey for a Unit from the user, but since oel doesn't know about usage keys, we'll then need to fetch the Unit (aka ContainerMixin) for that UsageKey anyway. So we might as well pass that Unit and prevent re-fetching it here.

bradenmacdonald: > I think it's worth combining these methods into one. 👍 on explicit published param.

OK, I pushed a change to consolidate these into one function in both the `containers` and `units` APIs.

openedx_learning/apps/authoring/publishing/model_mixins/publishable_entity.py
```diff
@@ -5,12 +5,14 @@
```
ormsbee, line 5: Should this just get folded back as `model_mixins.py`, since there we don't have any container mixins any longer?

openedx_learning/apps/authoring/publishing/model_mixins/publishable_entity.py
```diff
@@ -138,11 +135,24 @@ def __init__(self, content_obj):
...
+            if field_to_pev.model != self.content_version_model_cls:
+                # In the case of multi-table inheritance and mixins, this can get tricky.
+                # Example:
+                #   content_version_model_cls is UnitVersion, which is a subclass of ContainerVersion
+                # This versioning helper can be accessed via unit_version.versioning (should return UnitVersion) or
+                # via container_version.versioning (should return ContainerVersion)
+                intermediate_model = field_to_pev.model  # example: ContainerVersion
+                # This is the field on the subclass (e.g. UnitVersion) that gets
+                # the intermediate (e.g. ContainerVersion). Example: "UnitVersion.container_version" (1:1 foreign key)
+                field_to_intermediate = self.content_version_model_cls._meta.get_ancestor_link(intermediate_model)
```
ormsbee, line 151: I really appreciate these comments, btw. Thank you.

openedx_learning/apps/authoring/publishing/models/entity_list.py
```diff
@@ -0,0 +1,60 @@
...
+    #
+    # So our approach to this is to use a value of None (null) to represent an
+    # unpinned reference to a PublishableEntity. It's shorthand for "just use
+    # the latest draft or published version of this, as appropriate".
+    entity_version = models.ForeignKey(
+        PublishableEntityVersion,
+        on_delete=models.RESTRICT,
+        null=True,
+        related_name="+",  # Do we need the reverse relation?
+    )
```
ormsbee, line 60: I think this is missing a unique constraint on `(entity_list, order_num)`, so we don't get duplicates in race conditions.

openedx_learning/apps/authoring/publishing/models/entity_list.py
```diff
@@ -0,0 +1,60 @@
...
+
+    There is a row in this table for each member of an EntityList. The order_num
+    field is used to determine the order of the members in the list.
+    """
+
+    entity_list = models.ForeignKey(EntityList, on_delete=models.CASCADE)
+
+    # This ordering should be treated as immutable–if the ordering needs to
+    # change, we create a new EntityList and copy things over.
+    order_num = models.PositiveIntegerField()
```
ormsbee, line 34: I think this is missing a unique constraint on (entity_list, order_num), so we don't get duplicates in race conditions.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -513,3 +556,419 @@ def filter_publishable_entities(
...
+    @property
+    def entity(self):
+        return self.entity_version.entity
+
+
+def get_entities_in_container(
+    container: Container,
+    *,
+    published: bool,
+) -> list[ContainerEntityListEntry] | None:
```
ormsbee, line 853: (optional, not required for merge, but a future thought): Would it be more natural for folks to query the child entities via attributes on the models instead of having an API function for it?

ormsbee: So someone could write something like:

```python
published_unit = unit.versioning.published
if published_unit:
    component_versions = published_unit.children()
```

openedx_learning/apps/authoring/publishing/model_mixins/publishable_entity.py
bradenmacdonald: Yeah, might as well keep it where it was. Done.

openedx_learning/apps/authoring/publishing/models/entity_list.py
bradenmacdonald: I was thinking it doesn't really matter if `order_num` is not unique, but you're right - if it's not unique, it likely indicates a race condition and we should raise an exception rather than ignoring.

Added in e6334fb0a79598887e9945d85cd1d6a17c954da5

openedx_learning/apps/authoring/publishing/api.py
bradenmacdonald: That makes sense to me, and I think it would be a nice API. My only hesitation is that in general, it's unclear what you're "allowed" to do vs. not "allowed" to do when Django models are involved in Learning Core's API. We don't want people modifying fields on the models and calling `.save()` for example, but nothing distinguishes that from a method like `.children()`. So I guess I prefer to have API consumers treat the models as opaque things as much as possible, and stick to a functional API in `.api` wherever possible.

bradenmacdonald:
```python
published_unit = unit.versioning.published
if published_unit:
    for component_version in published_unit.children():
        # do something...
```
Oh, also this example reminds me: in this case, I think that `published_unit` is just a `UnitVersion` and it doesn't "know" if it's the published or draft or some other version of the unit. And even if it did know that, it doesn't know how to properly retrieve unpinned children. Say that `UnitVersion` 12 is both the latest published and draft version, but some of its draft children have unpublished changes. when you call `unit.versioning.published.children()`, it doesn't "know" whether to return the latest draft unpinned children or the latest published ones. That's why any API to retrieve children needs an explicit `published=True/False` parameter, and also another reason I think it doesn't make sense to have a `.children()` method, unless it just returns the entity list without specifying what version the children are at.

ormsbee: > That makes sense to me, and I think it would be a nice API. My only hesitation is that in general, it's unclear what you're "allowed" to do vs. not "allowed" to do when Django models are involved in Learning Core's API. We don't want people modifying fields on the models and calling `.save()` for example, but nothing distinguishes that from a method like `.children()`. So I guess I prefer to have API consumers treat the models as opaque things, and stick to a functional API in `.api` wherever possible.

The guideline I had in mind was to allow folks to query relationships but strictly draw the line at any sort of mutation/write operations.

> Oh, also this example reminds me: in this case, I think that `published_unit` is just a `UnitVersion` and it doesn't "know" if it's the published or draft or some other version of the unit. And even if it did know that, it doesn't know how to properly retrieve unpinned children.

Oof. Right. So no `children` call, but maybe `draft_children` and `published_children` calls later on.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -6,23 +6,34 @@
...
+    PublishableEntityMixin,
     PublishableEntityVersion,
+    PublishableEntityVersionMixin,
     Published,
     PublishLog,
     PublishLogRecord,
 )
 
+ContainerModel = TypeVar('ContainerModel', bound=Container)
+ContainerVersionModel = TypeVar('ContainerVersionModel', bound=ContainerVersion)
```
ormsbee, line 35: Maybe a comment why this is needed?

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -513,3 +558,419 @@ def filter_publishable_entities(
...
+    return record.new_version if record else None
+
+
+def create_container(
+    learning_package_id: int,
+    key: str,
+    created: datetime,
+    created_by: int | None,
+    # The types on the following line are correct, but mypy will complain - https://github.com/python/mypy/issues/3737
+    container_model: type[ContainerModel] = Container,  # type: ignore[assignment]
```
ormsbee, line 584: Nit (optional, bikeshedding): `container_cls` or `container_type` might be less ambiguous.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -513,3 +558,419 @@ def filter_publishable_entities(
...
+
+    return container_version
+
+
+def create_next_container_version(
+    container_pk: int,
+    *,
+    title: str | None,
+    publishable_entities_pks: list[int] | None,
+    entity_version_pks: list[int | None] | None,
```
ormsbee, line 764: Nit: I think it's slightly more natural to send tuples than have these separate, but again, non-blocking commentary.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -513,3 +558,419 @@ def filter_publishable_entities(
...
+    Args:
+        container: The Container, e.g. returned by `get_container()`
+        published: `True` if we want the published version of the container, or
+            `False` for the draft version.
+    """
+    assert isinstance(container, Container)
+    if published:
+        container_version = container.versioning.published
+        if container_version is None:
+            return None  # There is no published version of this container (yet). Should this be an exception?
```
ormsbee, line 873: Yeah, I'd raise an error here for consistency for now, but if we eventually switch over to an API where they use the relation on the model, this shouldn't be something that we have to worry about.

ormsbee (approved):
Minor nits, but nothing blocking.

openedx_learning/apps/authoring/publishing/api.py
bradenmacdonald: I would like to do that, and I'd like to also introduce some sort of verification that the entity version objects actually match the corresponding entity objects, but I think I'll leave that to another PR or get someone from my team to follow up.

openedx_learning/apps/authoring/publishing/api.py
bradenmacdonald: Sure, updated.

openedx_learning/apps/authoring/publishing/api.py
bradenmacdonald: OK, I went with `container_cls` instead.

openedx_learning/apps/authoring/publishing/api.py
bradenmacdonald: Added.

bradenmacdonald:
@kdmccormick did you want to do another pass on this? GitHub doesn't want me to merge until you clear your "request changes" review.

kdmccormick (approved):

PR #285, opened 2025-03-05
Minor typing improvements [FC-0083]

bradenmacdonald:
Added type annotations so that mypy checks more code. If a function doesn't have any type annotations at all, mypy will ignore it completely.

Django-rules now has types, so we don't need `# type: ignore[import]` on it.

Private ref: part of FAL-4051

### DIFF
--- a/tests/openedx_learning/apps/authoring/collections/test_api.py
+++ b/tests/openedx_learning/apps/authoring/collections/test_api.py
@@ -4,6 +4,7 @@ Basic tests of the Collections API.
 from datetime import datetime, timezone
 
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import User as UserType  # pylint: disable=imported-auth-user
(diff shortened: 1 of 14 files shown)


tests/openedx_learning/apps/authoring/collections/test_api.py
```diff
@@ -4,6 +4,7 @@
 from datetime import datetime, timezone
 
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import User as UserType  # pylint: disable=imported-auth-user
```
pomegranited, line 7: Boo.. too bad we have to trade a mypy ignore for a pylint ignore.

pomegranited (approved):
:+1:

- [x] <strike>I tested this</strike> -- typing changes only, so CI is sufficient to test this.
- [x] I read through the code
- [x] <strike>Includes documentation</strike> N/A
- [x] <strike>User-facing strings are extracted for translation</strike> N/A

PR #290, opened 2025-03-19
Support for grouped changesets of draft modifications

ormsbee:
## New Draft change-tracking models

This introduces `DraftChangeLog` and `DraftChangeLogRecord`, which are mostly draft equivalents of the `PublishLog` and `PublishLogRecord`. A `DraftChangeLog` entry is created for every group of changes (e.g. an import or a reset), and `DraftChangeLogRecord` has a record for every individual publishable entity that was changed.

The motivation for these models:

1. Batch changes into logical groupings, e.g. "discard changes in library" or "import a course's content into this library".
2. Accurate history reconstruction: We don't currently track reset-to-published operations anywhere, so we can't completely faithfully reconstruct historical draft information based purely off of the timestamps of when `PublishableEntityVersions` are created.

## Side-effects

It also introduces a new model `DraftSideEffect`, which should also have an equivalent `PublishSideEffect`. This is to capture the idea that sometimes a change in one publishable entity will affect another one, even we don't explicitly create a new version of the affected entity.

For instance, we define containers to have unpinned references to their children. When a Unit is defined this way, the Unit's version is only updated when the Unit's own metadata (e.g. its title) changes, or when it adds, removes, or reorders some of its children. The Unit's version does not increment when a child Component is updated with new edits. However the Unit is still affected, and is still logically part of the change or publish.

So every time a child of a container is modified, its container will be represented in the corresponding `DraftChangeLog` or `PublishLog`. In the case where only the child has been edited, the container's `DraftChangeLogRecord` or `PublishLogRecord` will show the same version for its `old_version` and `new_version` fields. This brings our backend more in line with user expectations, e.g. the Unit will be "published" whenever one of its Components is "published", even if that publish doesn't change the metadata we store for the definition of the Unit itself.

For now, the only planned side-effects are that changes in child elements affect their parent containers. However, the `DraftSideEffect` model could be used more broadly. For instance, if multiple `LTIBlocks` relied on some shared common configuration, we could make it so that changing that configuration caused side-effects to be written out that affect the appropriate LTI blocks.

That being said, we're going to want to be extremely thoughtful about when and where else we might apply this. For instance, we could use this to model inheritance, but (a) that would lead to an explosion in writes; and (b) that would not correlate to what users intuitively expect (e.g. you don't expect setting a due date on a subsection to count as an "update" of all the problems inside, because due dates are not something you define at the problem level--the fact that it's read by problems through inheritance is an implementation quirk).


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
ormsbee, line 1: I'm cherry-picking this over from https://github.com/openedx/openedx-learning/pull/293

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -496,24 +670,53 @@ def reset_drafts_to_published(learning_package_id: int, /) -> None:
...
+                      .exclude(entity__published__version_id=F("version_id")) \
+                      .exclude(
+                          # NULL != NULL in SQL, so we want to exclude entries
+                          # where both the published version and draft version
+                          # are None. This edge case happens when we create
+                          # something and then delete it without publishing, and
+                          # then reset Drafts to their published state.
+                          Q(entity__published__version__isnull=True) &
+                          Q(version__isnull=True)
+                      )
```
ormsbee, line 689: Note to self: was this removed on main since I branched? Why?

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -449,13 +465,199 @@ def set_draft_version(
...
+    publishable_entity_id: int,
+    publishable_entity_version_pk: int | None,
+    /,
+    set_at: datetime | None = None,
+    set_by: int | None = None,  # User.id
+    create_transaction: bool = True,
+) -> None:
+    """
+    Alias for set_draft_version taking PublishableEntity.id instead of a Draft.
+    """
```
ormsbee, line 541: Just wanted to call out that this is the first time I'm introducing this pattern into openedx-learning (or really anywhere in Open edX code that I'm aware of), where we use `functools.singledispatch` to give multiple versions of the same function and switch based on the first parameter type. Wanted to get people's thoughts.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -449,13 +465,199 @@ def set_draft_version(
...
+    containers that may have been affected by this draft update, UNLESS it is
+    called from within a bulk_draft_changes_for block. If it is called from
+    inside a bulk_draft_changes_for block, it will not add side-effects for
+    containers, as bulk_draft_changes_for will automatically do that when the
+    block exits.
+    """
+    if set_at is None:
+        set_at = datetime.now(tz=timezone.utc)
+    
+    tx_context = atomic() if create_transaction else nullcontext()
```
ormsbee, line 482: New convention: optionally creating a transaction, based on passed in parameters. This is to avoid unnecessarily opening transactions when we're being invoked from things that already open transactions.

openedx_learning/apps/authoring/publishing/models/draft_published.py
```diff
@@ -1,95 +0,0 @@
-"""
```
ormsbee, line 1: I split the two models in this module into `draft_log.py` and `publish_log.py`

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -496,23 +703,53 @@ def reset_drafts_to_published(learning_package_id: int, /) -> None:
...
+    active_change_log = DraftChangeLogContext.get_active_draft_change_log(learning_package_id)
+
+    # If there's an active DraftChangeLog, we're already in a transaction, so
+    # there's no need to open a new one.
+    if active_change_log:
+        tx_context = nullcontext()
+    else:
+        tx_context = bulk_draft_changes_for(
+            learning_package_id, changed_at=reset_at, changed_by=reset_by
+        )
```
ormsbee, line 738: A more extreme version of optional transaction creation.

ormsbee:
I have a couple of tests that I still need to write around less common cases (nesting `bulk_draft_changes_for` calls and side-effect calculation when there are multiple layers of containers). But I'm not likely to get to that until tonight, and I'd like to get eyes on other parts of this PR sooner if possible.

There would be at least two follow-up PRs:
* a small edx-platform one to send the user information to some calls, e.g. ("who did this soft-delete")
* one in openedx-learning that introduces a PublishSideEffect analog to DraftSideEffect.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -457,13 +473,199 @@ def set_draft_version(
     edited some content). Setting a Draft's version to None is like deleting it
     from Studio's editing point of view (see ``soft_delete_draft`` for more
     details).
+
+    Calling this function attaches a new DraftChange and attaches it to a
```
ormsbee, line 477: Self note: search-replace old references to `DraftChange` (it's `DraftChangeLogRecord` now, for consistency with the publishing side).

openedx_learning/apps/authoring/publishing/models/draft_log.py
```diff
@@ -0,0 +1,239 @@
...
+        blank=True,
+    )
+
+
+class DraftChangeLog(models.Model):
+    """
+    There is one row in this table for every time Drafts are created/modified.
+
+    Most of the time we'll only be changing one Draft at a time, and this will
+    be 1:1 with DraftChange. But there are some operations that affect many
```
kdmccormick, line 63:
```suggestion
    be 1:1 with DraftChangeLogRecord. But there are some operations that affect many
```

openedx_learning/apps/authoring/publishing/models/draft_log.py
```diff
@@ -0,0 +1,239 @@
...
+
+    @property
+    def changes(self):
+        """
+        Alias for "records".
+
+        We use "records" for consistency with PublishLogRecord, but somteimes
+        it's more natural to call these "changes".
+        """
+        return self.records
```
kdmccormick, line 85: I'm not big on this. In trying to have both consistency (records) and clarity (changes), we'll end up with neither.

openedx_learning/apps/authoring/publishing/models/draft_log.py
kdmccormick, line 63: Would such "some operations" include cases where parent containers are changed via a DraftSideEffect? If so, won't that actually be most operations, since most PEs have a parent container?

kdmccormick (commented):
Looking really solid so far. Will continue reviewing tonight.

openedx_learning/apps/authoring/publishing/models/draft_log.py
ormsbee: Fair. I'll get rid of this alias and use `records` everywhere.

openedx_learning/apps/authoring/publishing/models/draft_log.py
ormsbee: I guess I was thinking more of present-state libraries when writing this. I'll change it.

openedx_learning/apps/authoring/publishing/models/draft_log.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+class DraftChangeLogRecord(models.Model):
+    """
+    A single change in the Draft version of a Publishable Entity.
+
+    We have one unusual convention here, which is that if we have a
+    DraftChangeLogRecord where the old_version == new_version, it means that a
+    Draft's defined version hasn't changed, but the data associated with the
+    Draft has changed because some other entity has changed.
```
kdmccormick, line 95: 1. The current wording almost reads as if it's an accident or an edge case that old_version will sometimes equal new_version. I think that's unfair; it's really a solid model of what's happening and intuitively makes sense once you grok the whole system. Here's a suggested rewording:

```suggestion
    Changes often take form of a direct change to the content of the entity, which
    result in a bump of that entity's version, and thus new_version > old_version.
    However, this will not always be the case: if the data associated with the Draft
    has changed purely as a side effect of some other entity changing, then this will
    be represented here as a change log record where new_version == old_version.
```

2. Clarifying question: There will be instances where new_version > old_version, *and also* there's a side-effect record, right? As an example, any time an import includes parent-child relationships, I expect that a child and its parent can both have content changes. In other words: `new_version==old_version implies side-effect`, but not the other way around.

openedx_learning/apps/authoring/publishing/models/draft_log.py
```diff
@@ -0,0 +1,237 @@
...
+    old_version = models.ForeignKey(
+        PublishableEntityVersion,
+        on_delete=models.RESTRICT,
+        null=True,
+        blank=True,
+        related_name="+",
+    )
+    new_version = models.ForeignKey(
+        PublishableEntityVersion, on_delete=models.RESTRICT, null=True, blank=True
+    )
```
kdmccormick, line 127: If I'm reading the `on_deletes` correctly: deleting the changelog row will delete all its records, but deleting any PEs or PEVs which are actively referenced by a changelog record is disallowed. As an implication, it seems that pruning any given PEV becomes contingent upon first pruning the changelog records associated with it. Is that all as intended?

kdmccormick, line 127: Confirming my understanding--let me know if these are wrong. These might be useful as comments on this model.

1. It's valid for multiple DraftChangeLogRecords to exist with the same (entity, old_version, new_version), as long as draft_change_log is distinct for each one. For example, if a user is repeatedly editing a container within a unit U @ v1, we will have a series of DraftChangeLogRecords `(U.v1 -> U.v1), (U.v1 -> U.v1), ... , (U.v1 -> U.v1)`.
2. We cannot assume new_version >= old_version, because discarding changes will be modelled as setting the Draft pointer to an older version--specifically, the last-published version.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -457,13 +473,214 @@ def set_draft_version(
...
+            change = DraftChangeLogRecord.objects.create(
+                draft_change_log=change_log,
+                entity_id=draft.entity_id,
+                old_version_id=old_version_id,
+                new_version_id=publishable_entity_version_pk,
+            )
+            _create_container_side_effects_for_draft_change(change)
+
+
+@set_draft_version.register(int)
```
kdmccormick, line 538:
```suggestion
@set_draft_version.register
```

[Looks like you can omit `(int)` this since the first argument is type-annotated.](https://docs.python.org/3/library/functools.html#functools.singledispatch). Same for t

openedx_learning/apps/authoring/publishing/api.py
kdmccormick: Very very cool. I like this a lot more than taking a big `Union` of types and then having to handle them with a bunch of `if` statements at the top of the function. It's also nice because it encourages (a) static typing and (b) having one canonical version of function, and then one or more "aliases" which end up lightly wrapping the canonical verison.

openedx_learning/apps/authoring/publishing/models/draft_log.py
ormsbee: Both of those statements are correct. I'll add comments for them.

openedx_learning/apps/authoring/publishing/api.py
ormsbee: I don't understand why, but it doesn't infer/dispatch properly if I don't include `(int)`. A bunch of the tests break with error messages that show that it called the base version expecting a `Draft` with an `int` argument, e.g.:

```
        tx_context = atomic() if create_transaction else nullcontext()
    
        with tx_context:
>           old_version_id = draft.version_id
E           AttributeError: 'int' object has no attribute 'version_id'
```

ormsbee:
Self note: It's a super-edge case, but if someone uses one `bulk_draft_changes_for` to add and then delete an entity, we'd get a (old_version, new_version) entry of (None, None), which doesn't really mean anything because the Draft state was the same before and after and it was not a side-effect. Another weird edge case is if someone edits an entry to make it go from v1 -> v2 and then sets the draft version back to v1. In which case, we'd get (v1, v1), but again, with no side-effect.

Possible remedy: Make it so that our callback to generate side-effects first prunes out the entries that look like they would end up as side-effects, but which can't be because we haven't generated them yet.

openedx_learning/apps/authoring/publishing/api.py
kdmccormick: Weird, gotcha. I might poke at this if I have time, but no need to block on it.

openedx_learning/apps/authoring/publishing/models/draft_log.py
ormsbee: I rewrote a lot of the docstring for `DraftChangeLogRecord` in order to illustrate these and other possible scenarios.

ormsbee: @kdmccormick and I talked about this a bit offline ([notes here](https://openedx.atlassian.net/wiki/spaces/AC/pages/4612685830/Migrating+Courses+to+Learning+Core#2025-04-02)), but the upshot is that I'm going to keep this for now and re-evaluate how important pruning is going to be down the line. We had a couple of thoughts around hybrid pruning where the publishing models stay but other things are removed, as well as deeper history pruning that could remove old (and less interesting) things from the DraftChangeLog.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -457,13 +490,214 @@ def set_draft_version(
...
+    /,
+    set_at: datetime | None = None,
+    set_by: int | None = None,  # User.id
+    create_transaction: bool = True,
+) -> None:
+    """
+    Alias for set_draft_version taking PublishableEntity.id instead of a Draft.
+    """
+    tx_context = atomic() if create_transaction else nullcontext()
+    with tx_context:
```
bradenmacdonald, line 568:
```suggestion
    with atomic(savepoint=False):
```
Doesn't this have a similar effect?

openedx_learning/apps/authoring/publishing/api.py
bradenmacdonald: I like that it encourages static typing, but it seems to make the code longer and _slightly_ harder to follow compared to just having different names for the different versions of each function.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -457,13 +490,214 @@ def set_draft_version(
...
+        change.save()
+    except DraftChangeLogRecord.DoesNotExist:
+        # If we're here, this is the first DraftChangeLogRecord we're making for
+        # this PublishableEntity in the active DraftChangeLog.
+        change = DraftChangeLogRecord.objects.create(
+            draft_change_log=active_change_log,
+            entity_id=entity_id,
+            old_version_id=old_version_id,
+            new_version_id=new_version_id,
+        )
```
bradenmacdonald, line 621: Nit: could use `update_or_create` here to simplify this a bit.

openedx_learning/apps/authoring/publishing/models/draft_log.py
```diff
@@ -0,0 +1,307 @@
...
+                name="oel_dlr_idx_entity_rdcl",
+            ),
+        ]
+        verbose_name = _("Draft Change Log Record")
+        verbose_name_plural = _("Draft Change Log Records")
+
+
+class DraftSideEffect(models.Model):
+    """
+    Model to track when a change in one Draft affects other Drafts.
```
bradenmacdonald, line 228: This phrasing is confusing to me (is it accurate? because neither `cause` nor `effect` have any relationship to the `Draft` model that I can see). Could we say something like "Model to track when a draft change to an entity implicitly affects other entities such as parent containers" ?

openedx_learning/apps/authoring/publishing/api.py
ormsbee: I don't think it quite works because `create_defaults` isn't supported as an argument in Django 4.2. We want to set both `old_version` and `new_version` if we're creating, but only touch `new_version` if we're updating.

openedx_learning/apps/authoring/publishing/models/draft_log.py
```diff
@@ -0,0 +1,307 @@
...
+    Scenario 3: In a bulk_draft_changes_for context, we edit C1 so that the
+    draft version of C1 is now C1.v2. In the same context, we edit U1's list of
+    children so that C1 is no longer a child of U1.v2. Result:
+    - a DraftChangeLogRecord is created for C1.v1 -> C1.v2
+    - a DraftChangeLogRecord is created for U1.v1 -> U1.v2
+    - no SideEffect is created, since changing C1 does not have an impact on the
+      current draft of U1 (U1.v2). A DraftChangeLog is considered a single
+      atomic operation, so there was never a point at which C1.v1 -> C1.v2
+      affected the draft state of U1.
+    """
```
ormsbee, line 285: Note to self: Add a scenario where a change is effectively "undone", so no side-effect needs to be created, e.g. (C1.v1 -> C1.v2) followed by (C1.v1 -> C1.v1) in the same context means that nothing has changed from before the context to after the context. As such, it had no actual side effect on U1, and we should just prune things like this from the change log without creating a SideEffect entry.

ormsbee:
> A lot of our API methods, like `create_unit_version()` as one random example, accept an optional parameter like `created_by: int | None = None,`. But because it's optional, I've noticed that we haven't been very thorough in populating those fields when using Learning Core APIs within edx-platform. e.g. many of [these libraries APIs](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/content_libraries/api/libraries.py) don't even accept a user parameter so don't pass anything in to Learning Core.

Yeah, I have an edx-platform branch where I added `created_by` in a number of places.

> I'm wondering if we can auto-set the `created_by` and similar fields from the current `DraftChangeLogContext` `changed_by` value. We'd still have a problem of making sure edx-platform explicitly creates a `DraftChangeLogContext` with the current user set, but there'd be a lot less manual passing of user IDs around through each level. Maybe the same for the `changed_at` values too.

Yeah, that sounds useful. I'll play around with it.

openedx_learning/apps/authoring/publishing/api.py
bradenmacdonald: Ah, OK. Didn't check the older docs.

openedx_learning/apps/authoring/publishing/api.py
ormsbee: Okay, a major knock against it is that it seems to deeply confuse Pylance, making it so that the auto-complete is useless. The more I read about singledispatch and its handling in various tools, the more it seems to fall in the category of "this is a weird thing that has to be special cased everywhere". I'm leaning towards actually making it "a big Union of types and then having to handle them with a bunch of if statements at the top of the function" at this point because of this. It's "ugly" but it's simple and tooling-friendly.

ormsbee:
The edx-platform counterpart to this PR is: https://github.com/openedx/edx-platform/pull/36513

ormsbee:
Rebasing this now (need to account for the new migration added to the publishing app).

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -1087,6 +1355,7 @@ def get_containers_with_entity(
         ignore_pinned: if true, ignore any pinned references to the entity.
     """
     if ignore_pinned:
+        # TODO: Do we need to run distinct() on this?
```
kdmccormick, line 1358: @ormsbee @bradenmacdonald I presume that this TODO is not blocking for the PR. Let me know if either you feel that it is.

FWIW, I am confused by this segment of the query:

`...__entity_list__entitylistrow__...`

An entity list has multiple rows... what will this do, select the first one?

openedx_learning/apps/authoring/publishing/contextmanagers.py
```diff
@@ -0,0 +1,157 @@
...
+class DraftChangeLogContext(Atomic):
+    """
+    Context manager for batching draft changes into DraftChangeChangeLogs.
+
+    🛑 This is a PRIVATE implementation. Outside of the publishing app, clients
+    should use the bulk_draft_changes_for() API call instead of using this
+    manager directly, since this is a bit experimental and the implementation
+    may shift around a bit.
+
+    The main idea is that we want to create a context manager that will keep
```
kdmccormick, line 27: @ormsbee @bradenmacdonald I think that much of this docstring--everything except the internal implementation details--should be moved to the public `bulk_draft_changes_for` docstring. Let me know if you disagree, otherwise I'll make that change.

openedx_learning/apps/authoring/publishing/api.py
bradenmacdonald: It's not necessarily related to this PR / blocking, but yes it seems we do need to add `distinct()` here.

I suggest adding this additional code to [`test_units_containing`](https://github.com/openedx/openedx-learning/blob/e5e0bdbfbf5329417cb79281ef480e12755ea22d/tests/openedx_learning/apps/authoring/units/test_api.py#L984) in [this spot](https://github.com/openedx/openedx-learning/blob/e5e0bdbfbf5329417cb79281ef480e12755ea22d/tests/openedx_learning/apps/authoring/units/test_api.py#L984):

```
        with self.assertNumQueries(1):
            result = [
                c.unit for c in
                authoring_api.get_containers_with_entity(self.component_2.pk).select_related("unit")
            ]
        assert result == [
            unit1_1pinned,
            unit2_1pinned_v2,
            _unit3_no,  # Prob. need to rename these variables
            unit4_unpinned,
            _unit5_no,
        ]
```

If we update the test like so, it will fail until `distinct()` is added to `get_containers_with_entity`. If you'd like me to open a separate PR with these changes, I totally can.

> An entity list has multiple rows... what will this do, use the first one?

It's a filter so the intention is that it will search _all_ related rows until it finds that at least one that matches the condition. In this case, it is doing that, and you can verify that with the updated test case that I've suggested above, which now includes filtering against the second/third entities in the list.

openedx_learning/apps/authoring/publishing/contextmanagers.py
bradenmacdonald: @kdmccormick I don't know if I would move quite as much as you're suggesting, but certainly the `bulk_draft_changes_for` docstring is incomplete at the moment and needs more of this information + examples.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -446,25 +455,248 @@ def get_published_version(publishable_entity_id: int, /) -> PublishableEntityVer
...
+
+    The `processed_entity_ids` set holds the entity IDs that we've already
+    calculated side-effects for. This is to save us from recalculating side-
+    effects for the same ancestor relationships over and over again. So if we're
+    calling this function in a loop for all the Components in a Unit, we won't
+    be recalculating the Unit's side-effect on its Subsection, and its
+    Subsection's side-effect on its Section.
+
+    TODO: This could get very expensive with the get_containers_with_entity
+    calls. We should measure the impact of this.
```
bradenmacdonald, line 649: For what it's worth, here's how SQLite executes the query (if I added `distinct()`):

```
  + 16 0 47 SEARCH oel_publishing_entitylistrow USING INDEX oel_publishing_entitylistrow_entity_id_a6aff38f (entity_id=?)
  + 23 0 30 SEARCH oel_publishing_entitylist USING INTEGER PRIMARY KEY (rowid=?)
  + 26 0 17 SEARCH oel_publishing_containerversion USING INDEX oel_publishing_containerversion_entity_list_id_3b6d852b (entity_list_id=?)
  + 34 0 30 SEARCH oel_publishing_publishableentityversion USING INTEGER PRIMARY KEY (rowid=?)
  + 37 0 2 SEARCH oel_publishing_draft USING INDEX sqlite_autoindex_oel_publishing_draft_2 (version_id=?)
  + 45 0 30 SEARCH oel_publishing_publishableentity USING INTEGER PRIMARY KEY (rowid=?)
  + 48 0 30 SEARCH oel_publishing_container USING COVERING INDEX sqlite_autoindex_oel_publishing_container_1 (publishable_entity_id=?)
  + 55 0 47 SEARCH oel_publishing_published USING INDEX sqlite_autoindex_oel_publishing_published_1 (entity_id=?) LEFT-JOIN
  + 62 0 45 SEARCH oel_units_unit USING COVERING INDEX sqlite_autoindex_oel_units_unit_1 (container_id=?) LEFT-JOIN
  + 98 0 0 USE TEMP B-TREE FOR DISTINCT
  + 99 0 0 USE TEMP B-TREE FOR ORDER BY
```
It seems reasonable - first find all the entitylists that have the entity, then find all the containerversions referencing those entity lists, then find any active drafts that reference those containerversions. I think the main issue is that this could produce a lot of rows in the intermediate step of the query if there are a _lot_ of inactive/historic containerversions referencing the given entities.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -505,23 +742,54 @@ def reset_drafts_to_published(learning_package_id: int, /) -> None:
     Also, there is no current immutable record for when a reset happens. It's
     not like a publish that leaves an entry in the ``PublishLog``.
```
bradenmacdonald, line 743: This is no longer true, right? :)

openedx_learning/apps/authoring/publishing/api.py
ormsbee: Ack, sorry. I was working on the side-effect code when I looked at how this function was implemented, thought it looked a little funny, and then put a TODO to remind myself to look into it later... and them promptly forgot all about it. I'd opt to put a fix in a different PR.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -81,6 +86,7 @@
     "contains_unpublished_changes",
```
bradenmacdonald, line 86: Do you think we can change the implementation of `contains_unpublished_changes` to use this new data, as an optimization? Or no? (not a blocker at all)

bradenmacdonald (approved):
Looks nice!

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -446,25 +455,248 @@ def get_published_version(publishable_entity_id: int, /) -> PublishableEntityVer
...
+        # bulk_draft_changes_for() API call), or if we have to make our own.
+        learning_package_id = draft.entity.learning_package_id
+        active_change_log = DraftChangeLogContext.get_active_draft_change_log(
+            learning_package_id
+        )
+        change_log = active_change_log or DraftChangeLog.objects.create(
+            learning_package_id=learning_package_id,
+            changed_at=set_at,
+            changed_by_id=set_by,
+        )
```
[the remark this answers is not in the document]
ormsbee: Agreed. I think this was just me going through a couple of refactoring rounds and not properly cleaning up.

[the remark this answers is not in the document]
kdmccormick: Fixed

openedx_learning/apps/authoring/publishing/api.py
kdmccormick: Removed

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -1145,3 +1410,75 @@ def get_container_children_count(
...
+    If you make a change to an entity *without* using this context manager, then
+    the individual change (and its side effects) will be automatically wrapped
+    in a one-off change context. For example, this::
+
+        update_one_component(component.learning_package, component)
+
+    is identical to this::
+
+        with bulk_draft_changes_for(component.learning_package.id):
+            update_one_component(component.learning_package.id, component)
```
kdmccormick, line 1447: @ormsbee @bradenmacdonald About to merge with this new docstring. Lmk if you have any suggested edits, or we can edit it later if I merge before you get a chance to review.

bradenmacdonald: LGTM!

ormsbee:
@kdmccormick: Thank you for pushing this PR over the line!

PR #293, opened 2025-03-24
test: configure pytest for VS Code

ormsbee:
Add example VS Code configuration to automatically detect, run, and debug tests using VS Code's Testing bar. Developers will still need to copy these example files into their own settings in order to make use of them. This was done to avoid potential issues where developer-specific configuration could cause problems for other developers.

Some example screenshots:

## Test Browsing/Running

Screenshot 2025-03-25 130542
Screenshot 2025-03-25 130609

## Debugger

Screenshot 2025-03-25 131637

## Code Coverage

Screenshot 2025-03-25 135751
Screenshot 2025-03-25 131011

### DIFF
--- a/.vscode/launch.json.example
+++ b/.vscode/launch.json.example
@@ -15,7 +15,10 @@
 
             // This is needed because pytest-cov is incompatible with using the
             // debugger. See https://code.visualstudio.com/docs/python/testing#_pytest-configuration-settings
-            "env": {"PYTEST_ADDOPTS": "--no-cov"}
+            "env": {"PYTEST_ADDOPTS": "--no-cov"},
+
+            // Adds some magic for Django templates.
+            "django": true
           }
     ]
 }
--- a/.vscode/launch.json.example
+++ b/.vscode/launch.json.example
@@ -1,0 +1,21 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python: Debug Tests",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${file}",
+            "purpose": ["debug-test"],
+            "console": "integratedTerminal",
+            "justMyCode": false,
+
+            // This is needed because pytest-cov is incompatible with using the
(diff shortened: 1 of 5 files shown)


ormsbee:
Note: I need to tweak this a bit to make test coverage work at the same time as debugging.

ormsbee:
After discussing with some Axim folks, I've changed this PR to make example config files instead. It's ready for review.

bradenmacdonald:
I have a virtualenv just for `openedx-learning` on my host and always open that repo directly in VS Code, not using Tutor / containerization at all. In that case, almost everything is already working out of the box (and fast). The only thing that's necessary is putting the following into `.vscode/settings.json`:

```jsonc
{
    "python.testing.unittestEnabled": false,
    "python.testing.pytestEnabled": true,
}
```

Screenshot 2025-04-03 at 12 54 09 PM

Both the "run test" and "debug" test buttons are working well. I can set breakpoints in the debugger. I note that the coverage code is being run but doesn't seem to interfere with anything.

.vscode/launch.json.example
```diff
@@ -0,0 +1,24 @@
...
+            "console": "integratedTerminal",
+            "justMyCode": false,
+
+            // This is needed because pytest-cov is incompatible with using the
+            // debugger. See https://code.visualstudio.com/docs/python/testing#_pytest-configuration-settings
+            "env": {"PYTEST_ADDOPTS": "--no-cov"},
+
+            // Adds some magic for Django templates.
+            "django": true
+          }
```
bradenmacdonald, line 22: This launch config isn't working for me for some reason - when I run it via the "Run and Debug" panel with a test file open, it gives `No module named 'openedx_learning'`. But I'm wondering why we even need this launch configuration? Because the "Debug Test" buttons on the "Testing" panel are working just fine for me already.

ormsbee: Interesting. I've always been running it from the "Testing" panel where it works. But I see the same issues as you in the Run/Debug panel. I'll look into that, thank you.

bradenmacdonald: Ah, I didn't think it was affecting the "Testing" panel but now I can see that when I have this code in place it runs the debug mode tests differently. For me the "debug test" button works fine either with or without this `launch.json` config. The only difference I can see is that with the config in place, it produces more output in the terminal.

I don't think it's an issue that it doesn't also work from "Run and Debug", as long as other people aren't confused like I was.

PR #311, opened 2025-04-29
feat: filter out soft-deleted containers by default on get_containers [FC-00083]

rpenido:
## Description

This PR adds the `include_deleted` parameter to `get_containers` api and filter out soft-deleted containers by default

## Additional Information
- Related to https://github.com/openedx/frontend-app-authoring/issues/1870

## Testing Instructions
WIP

___
Private ref: [FAL-4144](https://tasks.opencraft.com/browse/FAL-4144)


openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -1211,12 +1212,17 @@ def get_containers(
     Args:
         learning_package_id: The primary key of the learning package
         container_cls: The subclass of Container to use, if applicable
+        include_deleted: If True, include deleted containers in the result.
```
bradenmacdonald, line 1215:
```suggestion
        include_deleted: If True, include deleted containers (with no draft version) in the result.
```
I think we should just clarify that we mean "draft deleted", since deleting the draft and deleting the published are separate actions.

bradenmacdonald (approved):
Looks great; thanks for fixing this. Just one minor suggestion before I merge.

PR #330, opened 2025-06-25
feat: Browsable Django Admin interface for Containers

kdmccormick:


kdmccormick:
@ormsbee @bradenmacdonald Low priority, but I pulled the admin pages in from the prototype branch. Could be nice to have merged before people who see our talk start poking around in Learning Core.

openedx_learning/apps/authoring/sections/admin.py
```diff
@@ -0,0 +1,36 @@
...
+
+class SectionVersionInline(admin.TabularInline):
+    """
+    Minimal table for subsecdtion versions in a subsection
+    """
+    model = SectionVersion
+
+
+@admin.register(Section)
+class SectionAdmin(ReadOnlyModelAdmin):
```
bradenmacdonald, line 20: Nit: I think the default title used in the admin interface isn't great, and it looks nicer if we override `__str__` on the model to pull the title from the latest draft.

|   |  with `__str__` |
|---|---|
| Screenshot 2025-07-16 at 3 01 06 PM | Screenshot 2025-07-16 at 3 00 35 PM |

openedx_learning/apps/authoring/sections/admin.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Very minimal interface... just direct the admin user's attention towards the related Container model admin.
+    """
+    list_display = ["section_id", "key"]
+    fields = ["see"]
```
bradenmacdonald, line 25: Nit: I found "See" to be a bit ambiguous at first.

Screenshot 2025-07-16 at 3 13 40 PM

I think this would be more clear:

Screenshot 2025-07-16 at 3 12 31 PM

Which can be accomplished by overriding `get_form` on the Section/Subsection/Unit admin class.

```python
    def get_form(self, request, obj=None, change=False, **kwargs):
        help_texts = {'container': 'To see details of this section, click above to see its container view.'}
        kwargs.update({'help_texts': help_texts})
        return super().get_form(request, obj, **kwargs)
```

openedx_learning/apps/authoring/publishing/admin.py
```diff
@@ -246,3 +250,261 @@ def get_queryset(self, request):
...
+            "publishable_entity__draft__version",
+        )
+
+    def draft(self, obj: Container) -> SafeText:
+        """
+        Link to this Container's draft ContainerVersion
+        """
+        if draft := obj.versioning.draft:
+            return format_html(
+                "Version {} ({})", draft.version_num, _entity_list_detail_link(draft.entity_list)
```
bradenmacdonald, line 331: Nit/optional: I think the `uuid` is not very useful, and the title is, so I suggest taking out the former and adding the latter.

```suggestion
                'Version {} "{}" ({})', draft.version_num, draft.title, _entity_list_detail_link(draft.entity_list)
```

Screenshot:

Screenshot 2025-07-16 at 3 19 36 PM

vs.

Screenshot 2025-07-16 at 3 20 07 PM

openedx_learning/apps/authoring/publishing/admin.py
```diff
@@ -246,3 +250,261 @@ def get_queryset(self, request):
...
+            )
+        return SafeText("-")
+
+    def see_also(self, obj: Container):
+        return one_to_one_related_model_html(obj)
+
+    def most_recent_parent_entity_list(self, obj: Container) -> SafeText:
+        if latest_row := EntityListRow.objects.filter(entity_id=obj.publishable_entity_id).order_by("-pk").first():
+            return _entity_list_detail_link(latest_row.entity_list)
+        return SafeText("-")
```
bradenmacdonald, line 351: Is there any reason to wrap these hyphens in `SafeText` ?

bradenmacdonald (approved):
Thanks, this is great! I had some feedback but it's super minor - consider it all optional.

openedx_learning/apps/authoring/publishing/admin.py
kdmccormick: Agreed. Changed for both Draft and Published.

I also removed all other occurrences of UUID from the various container-related admin lists, detail views, and inlines. We're not using the UUIDs for anything right now, so I'd rather it not clutter up the useful information with something irrelevant. If an operator really needs to know something's UUID, it is available on the linked PublishableEntity detail view.

openedx_learning/apps/authoring/publishing/admin.py
kdmccormick: It was for type consistency. But, it looks like mypy understands that SafeText is a subtype of str, so I can just change the return type to str and mypy is happy with me just returning `"-"`.

openedx_learning/apps/authoring/sections/admin.py
kdmccormick: Counterproposal... given that PEs already stringify to their key:
```python
class PublishableEntity(models.Model):
    ...
    def __str__(self):
        return f"{self.key}"
```
how about I make PEVs stringify like this:

```python
class PublishableEntityVersion(models.Model):
    ...
    def __str__(self):
        return f"{self.entity.key} @ v{self.version_num}"
```

and then make it so their mixin models default to the same behavior?
```python
class PublishableEntityMixin(models.Model):
    ...
    def __str__(self) -> str:
        return str(self.publishable_entity) 

class PublishableEntityVersionMixin(models.Model):
    ...
    def __str__(self) -> str:
        return str(self.publishable_entity_version)
```

I know that they titles nicer sometimes, but I feel that the keys and version nums better represent the models.

ormsbee: Maybe have the title at the end?

```python
    # <bikeshed>
    def __str__(self):
        return f"{self.entity.key} @ v{self.version_num} - {self.title}"
    # </bikeshed>
```

openedx_learning/apps/authoring/sections/admin.py
kdmccormick: Screenshot 2025-07-21 at 1 25 12 PM

PR #367, opened 2025-08-21
feat: add slugified filenames with short hash suffix for entity keys

dwong2708:


openedx_learning/lib/dump_naming.py
```diff
@@ -0,0 +1,39 @@
...
+    Make a filesystem-safe slug:
+    - Lowercase
+    - Replace only invalid filename characters with separator
+    - Collapse multiple separators
+    - Strip leading/trailing separators
+    """
+    value = value.lower()
+    value = re.sub(f"[{re.escape(INVALID_CHARS)}]", separator, value)
+    value = re.sub(f"{separator}+", separator, value)  # collapse duplicates
+    return value.strip(separator)
```
ormsbee, line 29: Can you use Django's [`slugify`](https://docs.djangoproject.com/en/5.2/ref/utils/#django.utils.text.slugify) for this instead?

openedx_learning/lib/dump_naming.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def hash_slugify_filename(identifier: str, ext: Optional[str] = "") -> str:
```
ormsbee, line 32: If we're not calling `ext` yet, please just get rid of the param for now.

openedx_learning/lib/dump_naming.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Hybrid filename = slugified identifier + short hash.
+    Keeps most special characters intact except filesystem-invalid ones.
+    """
+    slug = safe_slugify(identifier)
+    short_hash = hashlib.sha1(identifier.encode("utf-8")).hexdigest()[:5]
```
ormsbee, line 38: In general, prefer BLAKE2 to SHA. It's faster, more secure, and you can ask for any byte length for your digests:

```python
short_hash = hashlib.blake2b(digest_size=3).hexdigest()
```

openedx_learning/lib/dump_naming.py
```diff
@@ -0,0 +1,39 @@
+"""
+This module provides utilities for generating safe and consistent filenames for
+dumping learning package content.
+
+Examples
+print(hash_slugify_filename("My:Component"))     # → my_component_a12f4c.toml
+print(hash_slugify_filename("Math/Algebra+V2"))  # → math_algebra+v2_8d31a.toml
+print(hash_slugify_filename("User@Library#1"))   # → user@library#1_c9a0f.toml
+"""
```
ormsbee, line 9: I don't think this needs to be its own module. You can include it in `zipper.py` for now, since nothing else is likely to use it in the near future.

ormsbee (changes requested):
Thanks for the quick turnaround on this! I have a few low-level requests. At a higher level, please also include a test showing what happens in potential identifier name collision, e.g. two identifiers that differ only by case.

openedx_learning/lib/dump_naming.py
```diff
@@ -0,0 +1,39 @@
...
+    value = value.lower()
+    value = re.sub(f"[{re.escape(INVALID_CHARS)}]", separator, value)
+    value = re.sub(f"{separator}+", separator, value)  # collapse duplicates
+    return value.strip(separator)
+
+
+def hash_slugify_filename(identifier: str, ext: Optional[str] = "") -> str:
+    """
+    Hybrid filename = slugified identifier + short hash.
+    Keeps most special characters intact except filesystem-invalid ones.
```
ormsbee, line 35: In addition to saying what this function does, please add comments about why we need to do this at all.

openedx_learning/lib/dump_naming.py
dwong2708: Changes applied. thanks

openedx_learning/lib/dump_naming.py
dwong2708: changed

openedx_learning/lib/dump_naming.py
dwong2708: Done

tests/openedx_learning/apps/authoring/backup_restore/test_slug_hash.py
```diff
@@ -0,0 +1,37 @@
...
+        self.assertEqual(slugify_hashed_filename(long_string), f"{long_string}_4e84b3")
+
+    def test_slugify_hashed_filename_case_insensitivity(self):
+        # Test the slugify_hashed_filename function for case insensitivity
+        self.assertEqual(slugify_hashed_filename("My_Example"), "my_example_4f859c")
+        self.assertEqual(slugify_hashed_filename("MY_EXAMPLE"), "my_example_49be65")
+        self.assertEqual(slugify_hashed_filename("my_example"), "my_example_880989")
+        self.assertEqual(slugify_hashed_filename("mY_eXamPle"), "my_example_d28c02")
+        self.assertEqual(slugify_hashed_filename("My_ExAmPlE"), "my_example_79232e")
+        self.assertEqual(slugify_hashed_filename("mY_EXAMPLE"), "my_example_b91dc0")
```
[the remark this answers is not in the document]
dwong2708: Right, I was just curious about different letter case scenarios. However, I changed it to only one check.

tests/openedx_learning/apps/authoring/backup_restore/test_slug_hash.py
```diff
@@ -0,0 +1,37 @@
...
+    Test the slugify_hashed_filename function.
+    """
+
+    def test_slugify_hashed_filename(self):
+        # Test the slugify_hashed_filename function
+        self.assertEqual(slugify_hashed_filename("my_example"), "my_example_880989")
+
+    def test_slugify_hashed_filename_special_chars(self):
+        # Test the slugify_hashed_filename function with special characters
+        self.assertEqual(slugify_hashed_filename("my@ex#ample!"), "myexample_3366b5")
```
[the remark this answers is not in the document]
dwong2708: Applied. Thanks

PR #328, opened 2025-06-18
Publish side effects: state summary edition

ormsbee:
Attempt to reimplement https://github.com/openedx/openedx-learning/pull/307 with the ideas from https://github.com/openedx/openedx-learning/issues/317

Issue #317, opened 2025-05-18
Explicitly modeling publishing dependencies


ormsbee:

ormsbee:
@kdmccormick, @bradenmacdonald: I pulled the thread a bit on [my last PR comment](https://github.com/openedx/openedx-learning/pull/307#issuecomment-2882525558), and this is the result. Please let me know what you think when you have time.

ormsbee:
Edge case thoughts:

1. Should the state calculation for a container's dependencies include deleted dependencies? Or to put another way: Is the state hash for "three children where one is deleted" exactly the same as "two children, omitting the deleted one"? I'm leaning towards the two states being the same, i.e. skip deleted entries.
2. Since the state hash is based purely off of the dependencies, we can be in a place where the hashes are identical for two different containers, or two different versions of the same container (re-ordering). This is okay.

PR #369, opened 2025-08-25
Publishing dependency tracking and side-effect calculation + dependency state summary for published and draft

ormsbee:
Status: Data models and fundamental logic are implemented. I need to clean up a lot and add tests and some admin debug screens. I also need to do the data migration for existing containers data.

This PR introduces a publishing app concept of dependency tracking, as well as adding a `PublishingSideEffect` analog of the existing `DraftSideEffect`. The high-level goal is to be able to efficiently do operations like, "publish this entity and all its children" or answer queries like "when was the last time the published state of this entity changed (including its children)?".

The `publish_from_drafts` function has been modified to publish all dependencies.

Follow-on to https://github.com/openedx/openedx-learning/pull/328, broadly implements what was discussed in https://github.com/openedx/openedx-learning/issues/317, though I found it simpler to model all dependencies at the PublishableEntity level, rather than splitting them out into separate Draft and Published dependencies. That being said, the `dependencies_hash_digest` must be computed separately, since the published and draft versions of any given dependency can be different at any given time.

### DIFF
--- a/openedx_learning/apps/authoring/publishing/api.py
+++ b/openedx_learning/apps/authoring/publishing/api.py
@@ -1063,45 +1063,58 @@ def hash_for_log_record(
 
     Note that this code is a little convoluted because we're working hard to
     minimize the number of database requests. All the data we really need could
-    be derived from just the record itself, but at a far higher cost.
+    be derived from querying various relations off the record that's passed in
+    as the first parameter, but at a far higher cost.
 
-    When we say "dependency", we mean when one or more PublishableEntities are
-    dependencies of a PublishableEntityVersion. A change in the Draft or
-    Published state of the PublishableEntity causes an implicit change to the
-    PublishableEntityVersion. The common case we have at the moment is when a
-    container type like a Unit has Components as dependencies, i.e. the
-    UnitVersion specifies Component Entities as dependencies.
+    The hash calculated here will be used for the dependencies_hash_digest
+    attribute of DraftChangeLogRecord and PublishLogRecord. The hash is intended
+    to calculate the currently "live" (current draft or published) state of all
+    dependencies (and transitive dependencies) of the PublishableEntityVersion
+    pointed to by DraftChangeLogRecord.new_version/PublishLogRecord.new_version.
 
-    This means that the dependencies_hash_digest that summarizes the total state
-    of a PublishableEntity depends on:
+    The common case we have at the moment is when a container type like a Unit
+    has unpinned child Components as dependencies. In the data model, those
+    dependency relationships are represented by the "dependencies" M:M relation
+    on PublishableEntityVersion. Since the Unit version's references to its
+    child Components are unpinned, the draft Unit is always pointing to the
+    latest draft versions of those Components and the published Unit is always
+    pointing to the latest published versions of those Components.
+
+    This means that the total draft or published state of any PublishableEntity
+    depends on the combination of:
 
     1. The definition of the current draft/published version of that entity.
-    2. The current draft/published versions of all dependencies.
+       Example: Version 1 of a Unit would define that it had children [C1, C2].
+       Version 2 of the same Unit might have children [C1, C2, C3].
+    2. The current draft/published versions of all dependencies. Example: What
+       are the current draft and published versions of C1, C2, and C3.
 
-    Which is why it makes sense to capture in a log record, since
-    PublishLogRecords or DraftChagneLogRecords are created whenever one of the
+    This is why it makes sense to capture in a log record, since
+    PublishLogRecords or DraftChangeLogRecords are created whenever one of the
     above two things changes.
 
     Here are the possible scenarios, including edge cases:
 
-    Soft-deletions
-      If the record.new_version is None, that means we've just soft-deleted
-      something (or published the soft-delete of something). We adopt the
-      convention that if something is soft-deleted, its dependencies_hash_digest
-      is reset to the default value of ''.
-
     EntityVersions with no dependencies
       If record.new_version has no dependencies, dependencies_hash_digest is
-      likewise set to the default value of ''. This will be the most common
-      case.
+      set to the default value of ''. This will be the most common case.
 
-    EntityVersions with dependencies
     EntityVersions with dependencies
       If an EntityVersion has dependencies, then its draft/published state
       hash is based on the concatenation of, for each non-deleted dependency:
         (i)  the dependency's draft/published EntityVersion primary key, and
         (ii) the dependency's own draft/published state hash, recursively re-
              calculated if necessary.
+
+    Soft-deletions
+      If the record.new_version is None, that means we've just soft-deleted
+      something (or published the soft-delete of something). We adopt the
+      convention that if something is soft-deleted, its dependencies_hash_digest
+      is reset to the default value of ''. This is not strictly necessary for
+      the recursive hash calculation, but deleted entities will not have their
+      hash updated even as their non-deleted dependencies are updated underneath
+      them, so we set to '' to avoid falsely implying that the deleted entity's
+      dep hash is up to date.
 
     EntityVersions with soft-deleted dependencies
       A soft-deleted dependency isn't counted (it's as if the dependency were


ormsbee:
@bradenmacdonald, @kdmccormick: Based on your feedback today, I reworked the admin page entries to look like this:

image

It's not quite obvious, but the version information for the draft is italicized if it's different from the published version + dependency state.

openedx_learning/apps/authoring/publishing/models/publishable_entity.py
```diff
@@ -265,11 +277,47 @@ class Meta:
...
+    U1 (so U2.v2).
+
+    This restriction is important because our ability to calculate and cache the
+    state of "this version of this publishable entity and all its dependencies
+    (children)" relies on this being true.
+
+    .. no_pii
+    """
+    version = models.ForeignKey(PublishableEntityVersion, on_delete=models.RESTRICT)
+    entity = models.ForeignKey(PublishableEntity, on_delete=models.RESTRICT)
```
kdmccormick: I think at least one of these field names should change so that it's clear what the direction of the dependency is. One idea:

```suggestion
    version = models.ForeignKey(PublishableEntityVersion, on_delete=models.RESTRICT)
    dependent_entity = models.ForeignKey(PublishableEntity, on_delete=models.RESTRICT)
```

kdmccormick (commented):
still reviewing, just a couple quick thoughts so far.

ormsbee:
I've finished the refactoring to put the hash digests into the log records. Most of that work is in publishing/api.py, particularly `_update_branch_dependencies_hash_digests()` and `hash_for_log_record()`.

I still need to consolidate the migrations and then do some work on my data migration. I had a hacky one working for the old way of doing things, now trying to bring that over to the new refactored hash tracking.

ormsbee:
Okay, I think I fixed the issue. Back to data migration and cleanup.

openedx_learning/apps/authoring/publishing/models/publishable_entity.py
ormsbee: Yeah, I like that one. I'll do that.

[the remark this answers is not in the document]
ormsbee: Oh wait, I did this because if it's CASCADE, then entity deletion would retroactively change all the versions that referenced it, leaving no history of it.

ormsbee:
The mypy issue is fixed now (in a hacky way, but I'm really not interested in debugging mypy on the night before a release cut. :stuck_out_tongue:)

ormsbee:
Addressed all comments so far.

tox.ini
```diff
@@ -75,8 +75,8 @@ deps =
     setuptools
     -r{toxinidir}/requirements/quality.txt
 commands =
-    pylint openedx_learning openedx_tagging tests test_utils manage.py setup.py
-    mypy
+    # pylint openedx_learning openedx_tagging tests test_utils manage.py setup.py
```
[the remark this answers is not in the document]
ormsbee: lol, no, that's just me trying to induce the mypy failures faster, and forgetting to clean up after myself. Thank you for catching that.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -644,87 +782,407 @@ def _add_to_existing_draft_change_log(
...
+    for original_change in change_log.records.all():
+        affected_by_original_change = branch_cls.objects.filter(
+            version__dependencies=original_change.entity
         )
+        changes_and_affected = [
+            (original_change, current) for current in affected_by_original_change
+        ]
 
+        # These are the Published or Draft objects where we need to repoint the
+        # log_record (publish_log_record or draft_change_log) to point to the
```
kdmccormick:
```suggestion
        # log_record (publish_log_record or draft_change_log_record) to point to the
```

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -644,87 +782,407 @@ def _add_to_existing_draft_change_log(
...
+        # side-effect changes, e.g. the DraftChangeLogRecord that says, "This
+        # Unit's version stayed the same, but its dependency hash changed
+        # because a child Component's draft version was changed." We gather them
+        # all up in a list so we can do a bulk_update on them.
+        branch_objs_to_update_with_side_effects = []
+
+        while changes_and_affected:
+            original_change, affected = changes_and_affected.pop()
```
kdmccormick: I would rename this inner variable to something like `change` or `cause`, because:

* If we take "original change" to mean "a change coming directly from the change_log" (as is implied by `affected_by_original_change`), then I think anything at the 2nd layer of affect or lower is no longer "original".
* It's shadowing the outer loop's `original_change`, which is confusing.  I don't think it makes the code any more complicated for the inner loop (the one that descends the layers) to have its own variable.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -644,87 +782,407 @@ def _add_to_existing_draft_change_log(
...
+
+            # If the Draft or Published that is affected by this change is not
+            # already in the change_log, then we have to add it.
+            affected_version_pk = affected.version_id
+
```
kdmccormick:
```suggestion
```
I think this comment is already covered (better) by your comment right above `defaults={` and the extra variable here adds cognitive overhead-- I would just inline `affected.version_id`.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -644,87 +782,407 @@ def _add_to_existing_draft_change_log(
...
+                **change_log_param,
+                entity_id=affected.entity_id,
+                defaults={
+                    # If a change record already exists because the affected
+                    # entity was separately modified, then we don't touch the
+                    # old/new version entries. But if we're creating this change
+                    # record as a pure side-effect, then we use the (old_version
+                    # == new_version) convention to indicate that.
+                    'old_version_id': affected_version_pk,
+                    'new_version_id': affected_version_pk
```
kdmccormick:
```suggestion
                    'old_version_id': affected.version_id,
                    'new_version_id': affected.version_id,
```
(this goes with my previous suggestion)

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -1012,6 +1469,11 @@ def _create_container_version(
             title=title,
             created=created,
             created_by=created_by,
+            dependencies=[
+                entity_row.entity_id
+                for entity_row in entity_list.rows
+                if entity_row.is_unpinned()
+            ]
```
kdmccormick: Checking my understanding.. hypothetically:
* we have a unit U (published and draft @ version Uv1)
* we have component C (never published, draft @ version Cv2)
* Uv1 pins Cv1
* When you render the published version of Uv1, it includes Cv1, even though C is unpublished
* Now, Cv2 is published.
* There is no dependency relationship between Uv1 and C, so this does not affect Uv1
* and thus, we do not consider Uv1 to "contain unpublished changes"
* and, as before, when you render the published version of Uv1, it includes Cv1

right?

kdmccormick (changes requested):
I've looked at most of the code. As always, thanks for the careful docs, they've helped a lot. I'd like to give it at least one more pass early next week.

kdmccormick (changes requested):

openedx_learning/apps/authoring/publishing/api.py
ormsbee: Yup.

ormsbee:
@kdmccormick: Addressed all comments except the `EntityList.entitylistrow` refactoring, which I made a separate ticket for.

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -644,87 +782,402 @@ def _add_to_existing_draft_change_log(
...
+
+    EntityVersions with no dependencies
+      If record.new_version has no dependencies, dependencies_hash_digest is
+      likewise set to the default value of ''. This will be the most common
+      case.
+
+    EntityVersions with dependencies
+      If an EntityVersion has dependencies, we calculate the dependency hash
+      digest by taking a hash of all the UUIDs of the active version. [ fill in
+      more later]
```
kdmccormick: here, let me try to fill it in:
```suggestion
    EntityVersions with dependencies
      If an EntityVersion has dependencies, then its draft/published state
      hash is based on the concatenation of, for each non-deleted dependency:
        (i)  the dependency's draft/published EntityVersion primary key, and
        (ii) the dependency's own draft/published state hash, recursively re-
             calculated if necessary.
```

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -644,87 +782,402 @@ def _add_to_existing_draft_change_log(
...
+    of a PublishableEntity depends on:
+
+    1. The definition of the current draft/published version of that entity.
+    2. The current draft/published versions of all dependencies.
+
+    Which is why it makes sense to capture in a log record, since
+    PublishLogRecords or DraftChagneLogRecords are created whenever one of the
+    above two things changes.
+
+    Here are the possible scenarios, including edge cases:
```
kdmccormick: [optional] i would reorder this to start with the regular cases and end with the edge cases

openedx_learning/apps/authoring/publishing/api.py
```diff
@@ -644,87 +782,402 @@ def _add_to_existing_draft_change_log(
...
+
+    Soft-deletions
+      If the record.new_version is None, that means we've just soft-deleted
+      something (or published the soft-delete of something). We adopt the
+      convention that if something is soft-deleted, its dependencies_hash_digest
+      is reset to the default value of ''.
```
kdmccormick: reminder to add a note here, that we do this not bc it's strictly necessary for the recursive hash calculation, but
because deleted entities will not have their hash updated even as their non-deleted dependencies are updated underneath them, so we set to `''` to avoid falsely implying that the deleted entity's dep hash is up to date.

kdmccormick (approved):
LGTM!

I tested the backfill, tested that publish side affects percolate upwards, and tested the reverse migration.

Notes for future work, none of it critical for Ulmo AFAICT:

* We should get rip out any code which checked for a container's unpublished changes the old way (walking down the tree), if there is any of that.
* I think it could be good to add more cross-links in the admin interface, in support of devs and operators who are browsing around and trying to understand how the log changes and entities are related to one another. I didn't have the time to enumerate these places.
* As we discussed earlier, I think a lot of this code could become more readable if common abstract mixins were factored out from DraftChange{Log,LogRecord,SideEffect} and Publish{Log,LogRecord,SideEffect}.

PR #416, opened 2025-10-17
feat: Add missing metadata backup fields to the result endpoint data

dwong2708:
## Description

This PR adds the missing backup metadata fields, specifically:

- `created_by_email = "lp_user@example.com"`
- `origin_server = "cms.test"`

These fields are now included in the `package.toml` file as part of the backup metadata and will be used by the restore endpoint.

## Motivation

The restore process depends on these metadata fields to properly identify the backup source and author information.

### DIFF
--- a/openedx_learning/apps/authoring/backup_restore/serializers.py
+++ b/openedx_learning/apps/authoring/backup_restore/serializers.py
@@ -33,7 +33,6 @@ class LearningPackageMetadataSerializer(serializers.Serializer):  # pylint: disa
     format_version = serializers.IntegerField(required=True)
     created_by = serializers.CharField(required=False, allow_null=True)
     created_by_email = serializers.EmailField(required=False, allow_null=True)
-    created_by_full_name = serializers.CharField(required=False, allow_null=True)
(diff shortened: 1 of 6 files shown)


openedx_learning/apps/authoring/backup_restore/management/commands/lp_dump.py
```diff
@@ -4,6 +4,7 @@
 import logging
 import time
 
+from django.contrib.auth.models import User as UserType  # pylint: disable=imported-auth-user
```
rodmgwgu: Why do we need to disable a pylint check here?

wgu-taylor-payne: Should we change to use [get_user_model](https://docs.djangoproject.com/en/5.2/topics/auth/customizing/#django.contrib.auth.get_user_model)?

ormsbee (changes requested):
Sorry, this one is totally my fault: Please remove the full name in the output, since we store that in a non-standard way that only exists in edx-platform and not openedx-learning.

ormsbee (approved):

PR #474, opened 2026-02-04
docs: add ADR 0022 assessment criteria model

mgwozdz-unicon:


bradenmacdonald (commented):
Would you mind providing an end-to-end example of the data in each of these models?

I'm trying to understand the need for both `ObjectTag` and `oel_assessment_criteria`.

e.g. If the competency is "Writing Poetry" and "Unit 7: Submit a Poem" has the ObjectTag [Unit7->Writing Poetry], then we need `oel_assessment_criteria` to attach additional data to that specific object in that specific course?

e.g. The criteria _group_ would be "To achieve the 'Writing Poetry' Competency in course X, you have to: (Criteria 1: achieve a grade of 75% or better on [Unit7->Writing Poetry]) AND (Criteria 2: achieve a grade of 85% or better on [Unit9->Remixing Poetry])"

Do authors really want to specify the rules at such a granular level (separately and independently for each activity of each competency of each course)? I would have assumed that they just want a few high-level rules like "Earning a grade of 75% or better for all content in the course tagged with _both_ a competency and the 'CountsForCompetency' tag will earn that competency".

mgwozdz-unicon:
@bradenmacdonald 
> Would you mind providing an end-to-end example of the data in each of these models?

I go through an example in my Technical Design Document here: https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5604573185/Competency+Assessment+Criteria+Technical+Design+Document#Criteria-Aggregation-and-Logical-Grouping
Aside from that, the examples you gave are also valid, and I'm adding a similar one in my next update to the ADR.

> Do authors really want to specify the rules at such a granular level (separately and independently for each activity of each competency of each course)? I would have assumed that they just want a few high-level rules like "Earning a grade of 75% or better for all content in the course tagged with _both_ a competency and the 'CountsForCompetency' tag will earn that competency".

Yes and no. Authors will need to be able to indicate assessment criteria (which we are renaming to "competency criteria") for each individual subsection assignment and in the future also for each individual rubric criteria. However, you are correct that for most folks, they will want to define a default rule at their organization, taxonomy, or course level that applies for all their competency criteria associations. To address this, I've now added a new `oel_competency_rule_profile` table that can house the default for each of these scopes. However, there are some rare occasions where folks will want to override the default, so I've left this capability in the `oel_competency_criteria` table.

I also considered the possibility of combining the `oel_competency_criteria` table and the `oel_tagging_objecttag` table and I've added it to the list of Alternatives in the ADR. However, I think the biggest concern with this is that there may be some cases where the same `ObjectTag` needs to be associated with multiple competency criteria groups, which would necessitate duplicating rows in the `ObjectTag` table. Combining the tables would probably also result in a lot of null values in all of its CBE columns for each time tags are use for non-CBE reasons, and it could also conflate the usages and make it difficult to associate CBE specific constraints.

docs/decisions/0022-assessment-criteria-model.rst
```diff
@@ -0,0 +1,96 @@
...
+
+In order to support these use cases, we need to be able to model these rules (assessment criteria) and their association to the tag/competency to be demonstrated and the object (course, subsection, unit, etc) or objects that are used as the means to assess competency mastery. We also need to leave flexibility for a variety of different types as well as groupings to be able to develop a variety of pathways of different combinations of objects that can be used by learners to demonstrate mastery of a competency.
+
+Additionally, we need to be able to track each learner's progress towards competency demonstration as they begin receiving results for their work on objects associated with the competency via assessment criteria.
+
+Decision
+--------
+
+1. Update `oel_tagging_taxonomy` to have a new column for `taxonomy_type` where the value could be “Competency” or “Tag”.
```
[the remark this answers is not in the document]
bradenmacdonald: OK, that's very helpful. 

I agree it's a different thing: `allow_free_text` affects the content/structure of the taxonomy itself, but this "Competency" type still sounds exactly the same as a "regular" taxonomy in terms of its content/structure; it's just the _use case_ that's different - same sort of taxonomy, but using it in a more structured way (only on specific objects) and enabling new UI views for its usage.

In that case, I'm wondering if we should call it `use_for_competency=True` or `usage_type=Competency`, to clarify that the actual structure/behavior of the taxonomy on the taxonomy editor side is the same.

Can a "Competency" taxonomy still be viewed in the existing taxonomy view, if desired?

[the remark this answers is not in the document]
kdmccormick: What about making a simple CompetencyTaxonomy table, with a primary 1-1 key to the Taxonomy table? Essentially, rows in CompetencyTaxonomy would "mark" the Taxonomy as suitable for use as a competency. The `openedx_tagging` app wouldn't need to know anything about CBE. Furthermore, other CBE-related tables could ForeignKey to CompetencyTaxonomy instead of Taxonomy, ensuring on a DB level that they are referencing a suitable taxonomy.

This is similar to how we indicate that a Container is a Sections, Subsections, or Unit, without making the `containers` applet aware of the `sections`, `subsections`, or `units` applets.

docs/decisions/0022-competency-criteria-model.rst
```diff
@@ -0,0 +1,203 @@
...
+   This new database table will have the following columns:
+   1. `id`: unique primary key
+   2. `oel_tagging_taxonomy_id`: Foreign key to `oel_tagging_taxonomy.id` to identify which taxonomy is being designated as a competency taxonomy.
+
+2. Add new database table for `oel_competency_criteria_group` with these columns:
+
+   1. `id`: unique primary key
+   2. `parent_id`: The `oel_competency_criteria_group.id` of the group that is the parent to this one.
+   3. `oel_tagging_tag_id`: The `oel_tagging_tag.id` of the tag that represents the competency that is mastered when the competency achievement criteria in this group are demonstrated.
+   4. `course_id`: The nullable `course_id` to which all of the child competency achievement criteria's associated objects belong.
```
ormsbee: FWIW, this can now be a foreign key to [`CourseRun`](https://github.com/openedx/openedx-core/blob/62c937e96888b81eb7125554c036ba69e7ae7d2b/src/openedx_catalog/models/course_run.py#L29)

docs/decisions/0022-competency-criteria-model.rst
```diff
@@ -0,0 +1,203 @@
...
+   5. `name`: string
+   6. `ordering`: Indicates evaluation sequence number for this criteria group. This defines the evaluation sequence for siblings and enables short-circuit evaluation.
+   7. `logic_operator`: Either “AND” or “OR” or null. This determines how children are combined at a group node ("AND" or "OR").
+
+   Example: A root group uses "OR" with two child groups.
+
+   - Child group A (`ordering=1`) requires "AND" across Assignment 1 and Assignment 2.
+   - Child group B (`ordering=2`) requires "AND" across Final Exam and viewing prerequisites.
+   - If group A evaluates to true, group B is not evaluated.
```
ormsbee: Not a request for this ADR, but just as a side note: If it is simple to display the rules in a more readable form in the Django admin, it would be very helpful. It may be cumbersome for folks who are not familiar with the models to try to grok what's going on just from looking at the database tables.

docs/decisions/0022-competency-criteria-model.rst
```diff
@@ -0,0 +1,203 @@
...
+4. Add new database table for `oel_competency_criteria` with these columns:
+
+   1. `id`: unique primary key
+   2. `competency_criteria_group_id`: foreign key to competency achievement criteria Group id
+   3. `oel_tagging_objecttag_id`: Tag/Object Association id
+   4. `competency_rule_profile_id`: The nullable `competency_rule_profile_id` that points to the competency rule profile that should be applied when evaluating this competency achievement criteria. If null, then the system should look for a competency rule profile scoped to the competency's taxonomy, then to the course, then to the organization, and if none is found, use a default rule.
+   5. `rule_type_override`: Nullable enumerated rule type: “View”, “Grade”, “MasteryLevel” (Only “Grade” will be supported for now). When set, this overrides the `rule_type` specified in the associated `competency_rule_profile` for this specific competency achievement criteria.
+   6. `rule_payload_override`: Nullable JSON payload keyed by `rule_type` to avoid freeform strings. When set, this overrides the `rule_payload` specified in the associated `competency_rule_profile` for this specific competency achievement criteria. Examples:
```
ormsbee: You don't have to describe it at the specific field level, but it would be great if you could go into why we need this ability to override at these various levels from a business logic point of view.

docs/decisions/0022-competency-criteria-model.rst
```diff
@@ -0,0 +1,203 @@
...
+      2. `competency_criteria_id`: Foreign key pointing to competency achievement criteria id
+      3. `user_id`: Foreign key pointing to user_id (presumably the learner's id, although it appears that it is possible for staff to get grades as well) in `auth_user` table
+      4. `status`: “Demonstrated”, “AttemptedNotDemonstrated”, “PartiallyAttempted”
+      5. `timestamp`: The timestamp at which the student's competency achievement criteria status was set.
+   2. Add a new database table for `student_assessmentcriteriagroupstatus` with these columns:
+
+      1. `id`: unique primary key
+      2. `competency_criteria_group_id`: Foreign key pointing to competency achievement criteria group id
+      3. `user_id`: Foreign key pointing to user_id (presumably the learner's id, although it appears that it is possible for staff to get grades as well) in `auth_user` table
+      4. `status`: “Demonstrated”, “AttemptedNotDemonstrated”, “PartiallyAttempted”
```
ormsbee: If these are going to essentially be the same fields as `student_assessmentcriteriastatus`, it makes me wonder if there should be just one table that student status joins against. So in other words, if it's possible to make it so that there's just one recursive table that represents the structure of the evaluation tree, and a criteria could add its supplemental information in a 1:1 relationship to nodes in that tree, instead of that evaluation tree being conceptually a mix of two different models.

docs/decisions/0022-competency-criteria-model.rst
```diff
@@ -0,0 +1,203 @@
...
+
+   This new database table will have the following columns:
+   1. `id`: unique primary key
+   2. `oel_tagging_taxonomy_id`: Foreign key to `oel_tagging_taxonomy.id` to identify which taxonomy is being designated as a competency taxonomy.
+
+2. Add new database table for `oel_competency_criteria_group` with these columns:
+
+   1. `id`: unique primary key
+   2. `parent_id`: The `oel_competency_criteria_group.id` of the group that is the parent to this one.
+   3. `oel_tagging_tag_id`: The `oel_tagging_tag.id` of the tag that represents the competency that is mastered when the competency achievement criteria in this group are demonstrated.
```
ormsbee: If someone tries to delete this tag, do we allow it to cascade down and delete the associated competency groups, or do we prevent the deletion with an error?

ormsbee (changes requested):
Lots of good stuff here—these are mostly just questions on my part. Thank you!

docs/decisions/0022-competency-criteria-model.rst
```diff
@@ -0,0 +1,314 @@
...
+
+6. Learner progress status concepts (`student_competency*status`)
+
+   When a completion event (graded, completed, mastered, etc.) occurs for an object, determine and track the learner's status in earning the competency. To reduce recalculation frequency, store results at each level.
+
+   Relationship to other concepts:
+
+   - `student_competencycriteriastatus` tracks status at `CompetencyCriterion` leaf level.
+   - `student_competencycriteriagroupstatus` tracks status at `CompetencyCriteriaGroup` node level.
+   - `student_competencystatus` tracks top-level competency demonstration state.
```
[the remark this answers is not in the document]
mgwozdz-unicon: I switched all of these out from database table names to generic concept names. Let me know if I misunderstood the ask.

PR #475, opened 2026-02-04
Cbe assessment criteria location adr

mgwozdz-unicon:


bradenmacdonald:
What would be the downsides of putting both the CBE authoring backend and the learner status tracking features into a single new repo like `openedx_cbe`?

As I understand, Learning Core and even `openedx-platform` are both meant to hold "core" features, and other use cases can be built around them in separate repos.

My impression (which could be wrong) is that most Open edX instances will not be using CBE, so I'm not sure it belongs in the core, though we definitely want the core to provide all the features required (tags, events, versioning, users, learner activity, etc.) CBE seems like a perfect test case for this: do we provide enough hooks in the core so that a feature like CBE can be implemented as an installable app?

Issue #468, opened 2026-01-28
Refactor so that `openedx_content` is a top-level package


ormsbee:
After #457 is complete, refactor so that the `openedx_content` app is a top level package in openedx-learning, rather than something inside of `openedx_learning.apps`. This work will also require that all openedx-platform references to these paths be updated.

It's also worth noting that the `openedx_learning.lib` package has modules that are used by openedx-platform for things like `case_insensitive_char_field`, in order to normalize behavior between SQLite and MySQL (e.g. for test consistency). I'm not clear on where these should go, since they're not really tied to `openedx_content`.

I think it's likely we'll want to use the `openedx_learning` name for another top level app that will have some Learning Pathways related code in it.

bradenmacdonald:
> I think the main argument for it is that it allows us to easily skip all the test related stuff during packaging, including whatever test data files we have for backup/restore tests. FWIW, I lean towards having the tests separate, but it's not something I feel strongly about.

You can specify whatever sort of `exclude` glob you want in `pyproject.toml`, so it's possible to exclude tests regardless of where you place them. I guess having a separate `tests` folder is more common in python projects though, and it's not something I feel strongly about. I just get a little annoyed when I change some code in a file and it takes me some time to browse through the`tests` tree to find the corresponding tests, especially with deeply nested things. (If I change the `components` model for example, I'd need to run `cd ../../../../../tests/openedx_learning/apps/authoring/applets/components/` to get to the test files). This won't be as bad after we move `openedx_content` to the top-level and if we put `src` and `tests` as subdirs of that.

> We might also end up following `uv` conventions and putting things in a top level `src` dir as well.

Ah yeah, seems that https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/ has some solid reasons for recommending the `src` layout.

PR #471, opened 2026-01-30
build!: Rename the PyPI project to `openedx-core`

kdmccormick:
## Description

To represent the increasing scope of this repository, we are renaming it from openedx-learning to openedx-core. See the ADR for more details: https://github.com/openedx/openedx-core/blob/kdmccormick/openedx-core/docs/decisions/0021-openedx-core.rst

Practically, this PR:
* Updates GitHub hyperlinks in this repo from openedx/openedx-learning -> openedx/openedx-core
* Updates "Learning Core" to "Open edX Core" in various docs and comments
* Updates various docs and comments to represent the project's new scope.
* Publishes a new PyPI project called openedx-core, version starting at 0.34
* Publishes a final PyPI release of openedx-learning, versioned 0.33, which is empty and just installs openedx-core as a dependency

BREAKING CHANGE notice for operators: openedx-learning==0.33.0 will be the final release of openedx-learning. It will transitively install openedx-core==0.34.0 (fully code-compatible) and not receive updates going forward.  In order to receive updates, operators should install openedx-core instead.

## Notes for reviewers

 * I _think_ that I've updated pypi-publish.yml and tmp-openedx-learning/setup.py correctly in order to publish the openedx-learning final shell release. But, there's no easy way to test it without just running it. If it fails, I'll either do a quick followup PR, or I'll just log into PyPI and fix things up directly.
 * Since the repo name is now openedx-core, I'm proposing that we rebrand it from "Learning Core" to "Open edX Core", with "OexCore" being an alternative spelling for use in code names. When we need to talk about the particular APIs, we just say "Content API" or "Core Content API", "Tagging API" or "Core Tagging API", etc. [Here's an example in openedx-platform](https://github.com/openedx/openedx-platform/pull/38011/changes#diff-7bef8a690a77bf101d829b89215e4f774b86a651d8a810bac9f87700a5036a82R49-R51). Let me know what you think.

## Supporting info / merge considerations / testing

Before merging this, we'll:
- [x] Complete https://github.com/openedx/openedx-learning/pull/454
- [x] Complete https://github.com/openedx/openedx-learning/issues/468
- [x] Rename the repo in GH settings so that the new links to  openedx/openedx-core work

After merging this, I'll:
- [ ] Tag _both_ v0.33.0 and v0.34.0
- [ ] Confirm that packages were pushed to PyPI correctly. If not, log in with our PyPI admin credentials and fix it up.
- [ ] Finish and merge https://github.com/openedx/openedx-platform/pull/38011
[...]

PR #478, opened 2026-02-09
feat!: Make `openedx_content` and `openedx_tagging` flat, top-level APIs

kdmccormick:
## Context

Closes: https://github.com/openedx/openedx-learning/issues/468

This PR aligns on the new project structure which @ormsbee , @bradenmacdonald and I discussed. It follows on the work which Dave did here:

* https://github.com/openedx/openedx-learning/pull/454

This involves several breaking changes, which we've deemed OK since the package is still `v0.*`. The accompanying openedx-platform PR which addresses all those breaking changes is here:

* https://github.com/openedx/openedx-platform/pull/38004

After we merge those PRs, the final step will be to rename the repo and PyPI package to openedx-core.

* https://github.com/openedx/openedx-learning/pull/471

## The new structure

```
src/  # all packages released under openedx-learning (soon to be "openedx-core")
  openedx_tagging/        # The Tagging djangoapp
  openedx_content/        # The Content djangoapp (formerly "authoring")
  openedx_django_lib/     # Generic Django utils (formely "openedx_learning.lib")
  openedx_core/           # Stub package, just so we have a place to declare `__version__`
```

New high-level applications will be added as peers of these packages. For example, we expect to have `src/openedx_catalog` and `src/openedx_learning`.

New capabilities which fit well into the existing application can be added as ["applets" (ADR link).](https://github.com/openedx/openedx-learning/blob/main/docs/decisions/0020-merge-authoring-apps-into-openedx-content.rst#2-logical-separation-via-applets) The goal is to have manageable number of well-separated django apps at the top level, internally organized into applets. The top-level apps are hard to refactor (due to migrations) but the internal applets are very easy to refactor.

importlinter is used to enforce boundaries between the djangoapps, as well as between their internal applets -- see `.importliner` config file.

## BREAKING CHANGES

For developers installing this package (i.e., openedx-platform developers):

* `openedx_tagging.core.tagging.*` --> `openedx_tagging.*` 
* `openedx_learning.api.authoring` --> `openedx_content.api`
* `openedx_learning.api.authoring_models` --> `openedx_content.models_api`
* `openedx_learning.api.django.openedx_learning_apps_to_install` --> `openedx_content.settings_api.openedx_content_backcompat_apps_to_install`
  *  Additionally, the function's semantics have changed. Operators will need to explicitly add `"openedx_content"` to `INSTALLED_APPS` in addition to splatting the results of this function in.
* `openedx_learning.lib.*` -> `openedx_django_lib.*`
* `openedx_learning.lib.cache` --> removed. LRU caching didn't play well with rollbacks.
* `openedx_learning.lib.test_utils` --> removed. It was just there to help us clear the LRU cache during tests.
* `openedx_learning.contrib.media_server` --> removed
  * from Dave: "The original usage for media_server had to do with the fact that we're intentionally dumping things into a private media storage space that isn't publicly accessible. But in libraries, all the work is being done by [get_component_version_asset](https://github.com/openedx/openedx-platform/blob/d20b87b18062bf74f6bd05a604e30c21116bb4d9/openedx/core/djangoapps/content_libraries/rest_api/blocks.py#L381) in the content_libraries REST API. In the Django admin, I switched it to using base64 encoding, since performance isn't really a concern there" 

For developers hacking on this package (i.e., Learning Core developers):

* The package will need to be uninstalled and re-installed in order to behave correctly. This is because the egg-info file has moved from `./openedx-learning.egg-info` over to `src/openedx-learning.egg.info` and the root package names have changed.
  * For development just on this repo, in a venv: `pip uninstall openedx-learning && make clean && pip install -e .` 
  * For development via tutor, with this repo bind-mounted:
    * Try `make clean` in this repo, then `tutor images build openedx openedx-dev`
    * If that fails, then try rebuilding without the cache (sorry): `tutor images build --no-cache openedx openedx-dev`

## Testing

I manually tested:
 * Taxonomies
   * Import
   * Download
   * Download template
   * Clicked around Django admin, now called "Open edX Core > Tagging"
* Libraries v2
  * Create problem, video, text, unit
  * Publishing 
  * Backup
  * Restore
  * Tagging
  * Clicked around LC's Django admin, now called "Open edX Core > Content"
* Courses
  * Lib unit reference 
  * Modifying a lib-referenced item
  * Tagging
  * Export

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -10,7 +10,7 @@ root_packages =
     openedx_django_lib
     openedx_core
 
-# This is layering within our top-level src folders (mostly Django apps).
+# This is layering between our top-level src folders (mostly Django apps).
 # This should be updated as new Django apps are added.
 # It's possible this might need to grow into multiple rules.
 [importlinter:contract:src_layering]
@@ -27,34 +27,34 @@ layers =
     # Django utilities. Should not dependend on any of the real apps (above).
     openedx_django_lib
 
-    # This just a empty packages to expose the __version__ number.
+    # This just an empty shell package, to expose the __version__ number.
     # Should not depend on anything.
     openedx_core
 
-# This is layering within our Content applets. Every new applet should be added to
-# this list when it it created.
+# This is "applet" layering, within our Content djangoapp.
+# Every new applet should be added to  this list when it it created.
 [importlinter:contract:content_applet_layering]
 name = "openedx_content's internal applets are layered correctly"
 type = layers
 layers=
-    # The public API is at the top. None of the apps should call to it.
+    # The public API is at the top. None of the internal applets should call to it.
     openedx_content.api
 
-    # The "backup_restore" app handle the new export and import mechanism.
+    # The "backup_restore" applet handles the new export and import mechanism.
     openedx_content.applets.backup_restore
 
-    # The "components" app is responsible for storing versioned Components,
-    # which is Open edX Studio terminology maps to things like individual
+    # The "components" applet is responsible for storing versioned Components,
+    # which in Open edX Studio terminology maps to things like individual
     # Problems, Videos, and blocks of HTML text. This is also the type we would
     # associate with a single "leaf" XBlock–one that is not a container type and
     # has no child elements.
     openedx_content.applets.components
 
-    # The "contents" app stores the simplest pieces of binary and text data,
+    # The "contents" applet stores the simplest pieces of binary and text data,
     # without versioning information. These belong to a single Learning Package.
     openedx_content.applets.contents
 
-    # The "collections" app stores arbitrary groupings of PublishableEntities.
+    # The "collections" applet stores arbitrary groupings of PublishableEntities.
     # Its only dependency should be the publishing app.
     openedx_content.applets.collections
 


projects/dev.py
```diff
@@ -35,7 +35,8 @@
     "django.contrib.admindocs",
 
     # Learning Core Apps
-    *openedx_learning_apps_to_install(),
+    "openedx_content",
```
kdmccormick:
```suggestion
    "openedx_tagging",
    "opeendx_content",
```

src/openedx_content/applets/media_server/apps.py
```diff
@@ -11,7 +11,7 @@ class MediaServerConfig(AppConfig):
     Configuration for the Media Server application.
     """
 
-    name = "openedx_learning.contrib.media_server"
+    name = "openedx_content.applets.media_server"
```
ormsbee: File should be deleted entirely?

projects/urls.py
```diff
@@ -7,7 +7,7 @@
 urlpatterns = [
     path("admin/doc/", include("django.contrib.admindocs.urls")),
     path("admin/", admin.site.urls),
-    path("media_server/", include("openedx_learning.contrib.media_server.urls")),
-    path("tagging/rest_api/", include("openedx_tagging.core.tagging.urls")),
+    path("media_server/", include("openedx_content.applets.media_server.urls")),
```
ormsbee:
```suggestion
```

src/openedx_content/settings_api.py
```diff
@@ -0,0 +1,41 @@
...
+
+def openedx_content_backcompat_apps_to_install():
+    """
+    Return list of Django apps to add to INSTALLED_APPS for backwards compatibility.
+
+    Version 0.31.0 of openedx-core (nee openedx-learning) reorganized its installation profile
+    from many Django apps within `openedx_learning.apps.authoring` into a single `openedx_content`
+    Django app. For backwards compatibility, anything that installed this package prior to 0.31.0
+    should install the apps returned by this function, in addition to installing `openedx_content`
+    itself.
```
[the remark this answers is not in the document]
ormsbee: Okay, maybe not always. We can probably make a future squashed migration in `openedx_content` that skips the state transfer magic and just makes tables the normal way, and make more squashed migrations on the openedx-platform side to bridge into it. But as it stands right now, if you have a plugin, you'd need to include the backcompat apps in process when testing so those migrations would run, or openedx_content is going to complain that its dependency migrations aren't found.

.importlinter
```diff
@@ -1,59 +1,73 @@
...
+    openedx_tagging
+    openedx_content
+    openedx_core
+
+[importlinter:contract:tagging_and_content_independent]
+name = "openedx_tagging and openedx_content are independent"
+type = independence
+modules=
+    openedx_tagging
+    openedx_content
```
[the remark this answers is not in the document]
ormsbee: Non-blocking, FWIW. Definitely don't have to hash this out now.

ormsbee (commented):
Did my first pass. Minor request around comment wording for `openedx_content_backcompat_apps_to_install`, and `media_server` removal comments. The tagging stuff looks fine to me, but I don't know if there are any weird search-related issues around moving the app around—it seems unlikely, but I mention it in case it rings any bells for @bradenmacdonald.

openedx_learning/lib/cache.py
```diff
@@ -1,34 +0,0 @@
-"""
```
kdmccormick: Per a discussion with @ormsbee , lru_caching is removed, as it is unused.

kdmccormick (commented):
@bradenmacdonald -- this is ready for review. 

I haven't manually tested yet, but I'll do that in tandem with finishing the openedx-platform PR, which I just got a green build on.

Some miscellaneous notes ⬇️

bradenmacdonald:
> The tagging stuff looks fine to me, but I don't know if there are any weird search-related issues around moving the app around—it seems unlikely, but I mention it in case it rings any bells for @bradenmacdonald.

Yeah, I don't think it should cause any problems.

bradenmacdonald:
The code looks great, @kdmccormick. I'm going to rebuild my devstack and see if it works.

Makefile
```diff
@@ -83,18 +83,14 @@ selfcheck: ## check that the Makefile is well-formed
 
 extract_translations: ## extract strings to be translated, outputting .mo files
 	rm -rf docs/_build
-	cd openedx_learning && ../manage.py makemessages -l en -v1 -d django
-	cd openedx_learning && ../manage.py makemessages -l en -v1 -d djangojs
-	cd openedx_tagging && ../manage.py makemessages -l en -v1 -d django
-	cd openedx_tagging && ../manage.py makemessages -l en -v1 -d djangojs
+	./manage.py makemessages -l en -v1 -d django
+	./manage.py makemessages -l en -v1 -d djangojs
```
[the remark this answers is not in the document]
kdmccormick: This was actually broken because manage.py wasn't marked executable.

I pushed a commit to fix that, but also pushed a commit to generally enable translations on this repo, which they currently aren't. https://github.com/openedx/openedx-learning/issues/483

kdmccormick:
Thanks @bradenmacdonald 

I smoke tested with the sandbox on https://github.com/openedx/openedx-platform/pull/38004, and the only issue I hit is that when I try to back up the library, instead of downloading a ZIP, my browser gets redirected to `/media/user_tasks/2026/02/13/lib-kyle-1-2026-02-13-152329.zip` (for `lib:kyle:1`) which 404s. Not sure if that's a bug in this PR, or an issue with the sandboxes. I'll dig in later today, but let me know whether you observe that too Braden.

bradenmacdonald:
@kdmccormick When running `tutor images build openedx-dev`, I did get both an unrelated warning about the license spec in setup.py and an error about the `openedx_learning.egg-info` directory. It might be something specific to how my mount points are configured, but I had to delete that directory and then try again.

```
 > [development 15/24] RUN uv pip install -e "/mnt/openedx-learning":
0.243 Using Python 3.12.12 environment at: /openedx/venv
0.649 error: The build backend returned an error
0.649   Caused by: Call to `setuptools.build_meta:__legacy__.build_editable` failed (exit status: 1)
...
0.649 [stderr]
0.649 SetuptoolsDeprecationWarning: License classifiers are deprecated.
...
0.649         Please consider removing the following classifiers in favor of a SPDX license expression:
0.649         License :: OSI Approved :: GNU Affero General Public License v3 or later (AGPLv3+)
0.649         See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
...
0.649 error: Cannot update time stamp of directory 'src/openedx_learning.egg-info'
0.649 
0.649 hint: This usually indicates a problem with the package or the build environment.
```

bradenmacdonald:
@kdmccormick Yeah, don't worry about it. Likely only affects a subset of the devs who have it bind-mounted, which is not going to be many. And it's easy enough to fix.

Issue #469, opened 2026-01-28
Create an `openedx_catalog` app with a representation for CourseRuns


ormsbee:
The Learning Pathways project would benefit from having a representation of CourseRuns in this repo, since the first Pathways will be configured using CourseRuns.

As a migration step, we would push data into this new model from openedx-platform's CourseOverview. At some later point, we can reverse the dependency relationship.

Some high level thoughts/ramblings:

1. We probably want a very small table because it's going to be joined against *a lot* in the future. Common metadata should go in 1:1 models that we optionally  join on. We want to avoid CourseOverview's "just one more field" syndrome, and try to divide up supplementary data into logical pieces like `CourseSchedule`.
2. For phase one, we only need the minimal set of course data needed for Learning Pathways to query.
3. We could make the initial model more general, e.g. `LearningContext`, even if we only push the course keys to it for now (and then subclass or 1:1 for CourseRun). I'm not sure if this is really the way we want to go though, because `LearningContext` feels like it's only a catalog concept by coincidence because of how the system evolved.
4. We should use actual opaque key Django fields.
5. I'm guessing we'll want to use RBAC (opeendx-authz)?

bradenmacdonald:
> We could make the initial model more general, e.g. LearningContext, even if we only push the course keys to it for now (and then subclass or 1:1 for CourseRun). I'm not sure if this is really the way we want to go though, because LearningContext feels like it's only a catalog concept by coincidence because of how the system evolved.

Yeah. If we think of LearningContext as "a context in which we will display learnable content (to learners OR authors) and be able to track user state/grades" (i.e. any context where we render XBlocks), then "content library", "blog post", "course", and perhaps "pathway" are all learning contexts.

_If_ we define this "catalog" app as something like "core data/models to facilitate learner discovery and enrollment into enrollable things", then courses and pathways would be in scope for it, but other learning contexts like content libraries and the (theoretical) "blog post" context would not.

PR #479, opened 2026-02-10
An openedx_catalog app with a representation for CourseRuns [FC-0117]

bradenmacdonald:
An implementation of #469.

Related platform PR: https://github.com/openedx/openedx-platform/pull/38023

## Notes

- This adds a new dependency on `edx-organizations` (which also requires Pillow for the `logo` field)
- This _uses_ `opaque-keys`, which was already an indirect dependency but not used directly
- Before this PR, only course runs have a `display_name`, and catalog courses do not have names nor exist in the core platform at all. This PR adds a `title` (display name) field to the new core `CatalogCourse` model to support various use cases, including [the proposed new studio home page](https://github.com/openedx/openedx-learning/issues/469#issuecomment-3843853272). See the code for how this can be backfilled and how runs can always override the name for each run.
- This PR does not really include a CRUD API (neither python nor REST) for manipulating `CatalogCourse`/`CourseRun` objects; just a minimal API that platform code can use to keep them in sync with `CourseOverview`.
- However, this PR _does_ include complete Django admin views that admins and developers can use for provisioning course runs, deleting course runs, testing data migrations, etc.
    * Screenshot 2026-02-21 at 4 43 26 PM
    * Screenshot 2026-02-21 at 4 44 19 PM
- This current does NOT allow course runs to have a different course code or org code from other runs of the same catalog course. (This restriction could be loosened in the future.) However, the org code capitalization may vary among runs.
- Currently, it is the `course_overviews` app that syncs data from modulestore -> `CourseOverview` -> `openedx_catalog`. It would be more robust and future-friendly to instead sync data directly from `SplitModulestoreCourseIndex` -> `openedx_catalog` , but it's harder to get information like `display_name` and `language` in the latter case as that's not available in the `SplitModulestoreCourseIndex` table. It could be retrieved from Mongo though.
- This PR bakes into the data model the assumption that all runs of the same catalog course are in the same language. [Discussion on Slack](https://openedx.slack.com/archives/C057J2D1WU9/p1772041595020869)
- Because these models are updated by `CourseOverview` based on the `course_published` signal, any test cases in platform that want to use these models have to be sure to enable that signal, which is disabled for test by default.
- ⚠️ Because we rely on signals to update `CourseRun` when `CourseOverview` is updated, and CourseOverview itself is updated after courses are already created, it's _possible_ that an error will occur and the `CourseRun` will never get created or updated. This will result in an error in the logs, but will not block course creation etc., so the error is likely to go unnoticed with the system as it exists today.

## Architecture Diagram

See [ARCHITECTURE.md](https://github.com/openedx/openedx-core/blob/braden/catalog/src/openedx_catalog/ARCHITECTURE.md).

## Open Questions for Future Follow-Up

- [ ] Should we make the new `CourseRun` a [`SoftDeletableModel`](https://django-model-utils.readthedocs.io/en/3.1.2/models.html#softdeletablemodel) to support course deletion without data loss? (Maybe we can't now, because soft-deleting it in that one table wouldn't affect the other tables that the system actually references. But in the future we could add this.)
- [ ] Is "Course Schedule" a core concept that should live in this catalog app? I think yes, but it could also have a draft/publish workflow and ultimately live in `openedx_content`
- [ ] Is "Course Visibility" a core concept that should be in this catalog app? (We actually have it defined as `catalog_visibility`, `visible_to_staff_only`, and `course_visibility` which all have different effects and different enum values, and also need to support "use system default")
- [ ] Integrate authz ?
- [ ] Check if we should deprecate [edx-organization's `OrganizationCourse`](https://github.com/openedx/edx-organizations/blob/3b306d18787a956e1dcd29a63c03709233546668/organizations/models.py#L54)
- [ ] test what happens to enrollments if course is deleted then re-created with different capitalization (not even using this PR)

### DIFF
--- a/src/openedx_core/__init__.py
+++ b/src/openedx_core/__init__.py
@@ -6,4 +6,4 @@ The public APIs belong to the specific apps (openedx_content, openedx_tagging, e
 """
 
 # The version for the entire repository
-__version__ = "0.35.0"
+__version__ = "0.36.0"


src/openedx_catalog/models/course_run.py
```diff
@@ -0,0 +1,233 @@
...
+    )
+    # The catalog course stores the 'org' and 'course_code' fields, which must match the ones in the course ID.
+    # Note: there is no need to load this relationship to get 'org' or 'course_code'; get them from `course_id` instead.
+    catalog_course = models.ForeignKey(
+        CatalogCourse,
+        on_delete=models.PROTECT,
+        null=False,
+        related_name="runs",
+    )
+    run = case_sensitive_char_field(
```
kdmccormick: since we have `org_code` and `course_code`, why not `run_code`?

as an aside, I like the word "code" for this short sluggy things that combine together to make opaquekeys. We've called it several other things in other places ("slug", "url_name", "id", "key") but I'd been keen to retcon them all into "code" -- `library_code`, `collection_code`, `block_code`, `container_code`, etc.

src/openedx_catalog/models/catalog_course.py
```diff
@@ -0,0 +1,241 @@
...
+    def url_slug(self):  # Do we need this? Would an opaque key be better?
+        """
+        An ID that can be used to identify this catalog course in URLs or APIs.
+        In the future, this may be an editable SlugField, so don't assume that
+        it never changes.
+        """
+        # '+' is a bad separator because it can mean " " in URLs.
+        # '-', '.', and '_' cannot be used since they're allowed in the org code
+        # So for now we use ':', and in the future we may make the whole slug customizable.
+        return f"{self.org_code}:{self.course_code}"
```
kdmccormick: whether or not to make it an OpaqueKey key... great question 😛 

My take on OpaqueKeys is that their implementation is away more complicated than it needs to be, _buuuuut_, they are still better than parsing everything by hand all over the place, which is what will naturally happen over time whenever we have a key-like thing floating around.

As a compromise, what about returning a string that we could retroactively turn into an opaquekey if we see the need? All it would need is a prefix, like:

```suggestion
        return f"catalog-course:{self.org_code}:{self.course_code}"
```

src/openedx_catalog/models/course_run.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+
+    # Use this field for relationships within the database:
+    id = models.BigAutoField(
+        primary_key=True,
+        verbose_name=_("Primary Key"),
+        help_text=_("The internal database ID for this course. Should not be exposed to users nor in APIs."),
+        editable=False,
+    )
+    course_id = CourseKeyField(
```
kdmccormick:
```suggestion
    course_key = CourseKeyField(
```

Could we call this `course_key`? `*_id` in Django usually a foreign key, and I think it's unfortunate we've used `course_id` in so many places. I'd love to standardize on `*_key` for anything that's an OpaqueKey instance.

kdmccormick (changes requested):
Partial review as I'm signing off for tonight, but overall really solid. I appreciate the comments and validation a lot. My only concerns so far are some superficial naming stuff.

src/openedx_catalog/models/catalog_course.py
bradenmacdonald: Done! 1b6477c137f23532147edf4b9449791ded56c715

src/openedx_catalog/models/course_run.py
bradenmacdonald: Yeah, I should have done that in the first place. Done: 88aef751beba8c969eead17459b1b9023bb9ba19

src/openedx_catalog/signals.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Nothing stops users from changing Organization.short_name entries in the Django admin, but any changes other than
+    capitalization fixes will result in totally invalid data, as CatalogCourse will be related to an Organization that
+    no longer matches the "org" part of the related CourseRun's course IDs.
+    """
+    if not instance.pk:
+        return  # It's a brand new Organization; we don't care
+
+    prev_org_code = Organization.objects.get(pk=instance.pk).short_name
+    new_org_code = instance.short_name
```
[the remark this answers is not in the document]
bradenmacdonald: I think the only such situation is this:
* An Organization "MITx" is renamed to "MIT" -> meanwhile all the associated course runs have keys like `course-v1:MIT+foo+bar`.  Don't throw an error, because this is now "more correct". 

This situation shouldn't really be possible if you are using these models correctly, but the "org.short_name matches course_key.org" rule is not actually enforced by the database, because we can't write constraints across table boundaries. So it is possible if you are using the `.update()` manager API or raw/custom SQL, or anything else that bypasses the checks in course_run.clean()

The other advantage is that it states an exact, example course key in the ValidationError.

That said, I don't think the situation described above is likely to occur so I'd be fine with simplifying this to your suggestion.

src/openedx_catalog/api_impl.py
```diff
@@ -0,0 +1,236 @@
...
+        defaults={"display_name": display_name},
+    )
+
+    if not created:
+        log.warning('Expected to create CourseRun for "%s" but it already existed.', str(course_key))
+
+    return new_run
+
+
+def delete_course_run(course_key: CourseKey) -> None:
```
kdmccormick: What's the use case for this function?

src/openedx_catalog/README.rst
```diff
@@ -0,0 +1,22 @@
...
+
+Motivation
+----------
+
+The existing ``CourseOverview`` model in ``openedx-platform`` is derived from various places, but mostly from the metadata fields of the root ``Course`` object stored in modulestore (MongoDB) for each course. As we slowly transition toward storing course content fully in Learning Core (in ``openedx_content``), we want to move to storing all course data and metadata in these sort of MySQL models. We're creating this new ``CourseRun`` model in ``openedx_catalog`` to support these goals:
+
+1. Provide a core model to represent each course, for foreign key purposes.
+2. To allow provisioning placeholder courses before any content even exists.
+3. To be much simpler and more performant than ``CourseOverview`` was (far fewer fields generally, fewer legacy fields, integer primary key).
+4. Perhaps to provide a transition mechanism, a pointer than can point either to modulestore content or learning core content, as we transition content storage.
```
kdmccormick:
```suggestion
4. Perhaps to provide a transition mechanism, a pointer than can point either to modulestore or openedx_content, as we transition content storage.
```

src/openedx_catalog/signals.py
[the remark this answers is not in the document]
bradenmacdonald: Done: a27856f2669630b877c58fded126a308cf874562

kdmccormick (approved):
Just tested again, looks great. Other than display_name->title, just one note.

PR #482, opened 2026-02-11
(WIP) feat: implement Pathways

Agrendalath:
This is the initial implementation of [Pathways](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5381160985/Learning+Pathways).

_Private-ref: [BB-10364](https://tasks.opencraft.com/browse/BB-10364)_


src/openedx_content/applets/pathways/models/enrollment.py
```diff
@@ -0,0 +1,139 @@
...
+        constraints = [
+            models.UniqueConstraint(
+                fields=["pathway", "email"],
+                name="oel_pathway_enrollallow_uniq",
+            ),
+        ]
+
+
+# TODO: Create receivers to automatically create audit records.
+class PathwayEnrollmentAudit(models.Model):
```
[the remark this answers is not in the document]
Agrendalath: @bradenmacdonald, we ported some logic from `openedx-platform`, so it should be possible to introduce a reusable approach at some point, but moving course enrollments away from `openedx-platform` (even partially) would be a significant task, as there are some edge cases to consider/deprecate (like processing the refunds for the verified enrollment track, which is also tied to the certificates API).

src/openedx_content/applets/pathways/models/pathway.py
```diff
@@ -0,0 +1,92 @@
...
+from django.utils.translation import gettext_lazy as _
+
+# from organizations.models import Organization
+from openedx_django_lib.fields import case_insensitive_char_field
+from openedx_django_lib.validators import validate_utc_datetime
+
+from ..keys import PathwayKeyField
+
+
+class Pathway(models.Model):
```
ormsbee: If you inherit from `PublishableEntityMixin`, it will add some useful convenience methods.

docs/decisions/0023-pathways.rst
```diff
@@ -0,0 +1,149 @@
...
+
+**PathwayStepVersion** (uses ``PublishableEntityVersionMixin``):
+
+- ``step`` — FK back to ``PathwayStep``.
+- ``context_key`` — ``LearningContextKeyField`` referencing the learning context (e.g., ``course-v1:OpenedX+DemoX+DemoCourse``). TODO: replace it with proper foreign keys (e.g., to ``CourseRun``) before finalizing the data model.
+- ``step_type`` — explicit type label (e.g., ``course``, ``pathway``) stored for query efficiency and validation rather than derived from the key at query time.
+- ``order`` — position within the pathway (0-indexed).
+- ``criteria`` — optional CEL expression overriding the pathway's ``default_step_criteria``.
+- ``group`` — optional label for grouping steps in pathway-level expressions (e.g., "core", "elective").
+
```
ormsbee: I think the way we'd want to group this is more like container parent-child relationships where:

`PathwayVersion` references `PathwaySteps`, i.e. there's a `PathwayVersionPathwayStep` model that joins `PathwayVersion` to `PathwayStep` and has an optional ordering field. PathwaySteps don't  reference Pathways directly. I mean, I guess we could even just call them Steps? 

That would let us let PathwaySteps be semantically meaningful in a standalone way, e.g. "This represents what it means to finish Intro Accounting", which can be used in various ways by different Pathways if we wanted. It also makes it easier for us to ensure that we don't have weird ordering conflicts between different PathwaySteps, since storing it on the PathwayVersion makes it awkward to put a constraint to prevent corrupt ordering states (two PathwaySteps that have the same order num).

mphilbrick211:
Hi @Agrendalath! Is this pull request still in progress?

Agrendalath:
@mphilbrick211, this is replaced by https://github.com/openedx/openedx-core/pull/761 and its linked PRs. Closing.

PR #501, opened 2026-03-16
feat: New `history log` api functions [FC-0123]

ChrisChV:
## Description

- Implements python api functions to get and build the history log of an entity:
    - `get_entity_draft_history`: Return `DraftChangeLogRecords` for a `PublishableEntity` since its last publication.
    - `get_entity_version_contributors`: returns the users who authored changes between two published versions of an entity
    - `get_entity_publish_history`:  Return all `PublishLogRecords` for a `PublishableEntity`
    - `get_entity_publish_history_entries`: Return the `DraftChangeLogRecords` associated with a specific `PublishLog`
    - `get_descendant_component_entity_ids`: BFS traversal of a container hierarchy, returning the IDs of all descendant component entities (used to scope queries when fetching history for      
  container children)
- Which edX user roles will this change impact? "Developer".

## Supporting information

- Github issue: https://github.com/openedx/frontend-app-authoring/issues/2805
- Used in: 
    - https://github.com/openedx/openedx-platform/pull/38178
    - https://github.com/openedx/frontend-app-authoring/pull/2948
- Internal ticket: https://tasks.opencraft.com/browse/FAL-4330

## Testing instructions

Follow the testing instructions in https://github.com/openedx/frontend-app-authoring/pull/2948

## Deadline

Before the Verawood cut.

## Other information

N/A


openedx-webhooks:
Thanks for the pull request, @ChrisChV!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

src/openedx_content/applets/publishing/api.py
```diff
@@ -584,6 +585,45 @@ def get_published_version(publishable_entity_or_id: PublishableEntity | int, /)
...
 
 
+def get_entity_draft_history(
+    publishable_entity_or_id: PublishableEntity | int, /
+) -> QuerySet[DraftChangeLogRecord]:
+    """
+    Return DraftChangeLogRecords for a PublishableEntity since its last publication,
+    ordered from most recent to oldest.
+
+    If the entity has never been published, all DraftChangeLogRecords are returned.
```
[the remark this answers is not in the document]
ormsbee: I think the current code does work in this case, to be clear. And I realize that it's not something that's likely to come up in the UX. I just want to make sure that the behavior is well defined and tested for in that edge case.

rpenido (commented):
Looking good @ChrisChV!

ormsbee:
Per this [slack thread](https://openedx.slack.com/archives/C06JG6RFE4V/p1774553980118459), I think we're going to want to extend the `PublishLog` model to include a nullable field that captures what it was that the user clicked "publish" on. Right now, if there is a Subsection -> Unit -> Component setup where only the component has changed, the data in the `PublishLog` does not allow us to differentiate which level the user pressed "Publish" at, and this seems to be a requirement of the UX.

Does this sound correct to you @sdaitzman, @marcotuts?

FYI @crathbun428, @jmakowski1123

src/openedx_content/applets/publishing/api.py
```diff
@@ -556,6 +561,271 @@ def get_published_version(publishable_entity_or_id: PublishableEntity | int, /)
...
+    if prev_published_at:
+        draft_qs = draft_qs.filter(draft_change_log__changed_at__gt=prev_published_at)
+
+    # Find the baseline: the version that was published in the previous publish group
+    # (None if this is the first publish for this entity).
+    baseline_version_id = prev_pub_record.new_version_id if prev_pub_record else None
+
+    # If reset_drafts_to_published() was called within this publish window, there
+    # will be a DraftChangeLogRecord where new_version == baseline. Use the most
+    # recent such record as the new lower bound so discarded entries are excluded.
```
ormsbee, line 757: I know I was the one who flagged this because I wanted to make sure the code noted it, but I actually don't think we should filter out discarded changes. It's a log, and "discard changes" is a legit thing that someone in that chain of edits did. I think we should be explicit that it happened, especially if people are going here because they're thinking, "I _know_ I was working on this, but what happened to my edits!?!" Then they could see "Dave discarded changes" and know who to <strike>strangle</strike> talk to.

src/openedx_content/applets/containers/api.py
```diff
@@ -879,3 +881,62 @@ def get_container_children_entities_keys(container_version: ContainerVersion) ->
     return list(
         container_version.entity_list.entitylistrow_set.values_list("entity__key", flat=True).order_by("order_num")
     )
+
+
+def get_descendant_component_entity_ids(container: Container) -> list[int]:
+    """
+    Return the entity IDs of all leaf (non-Container) descendants of ``container``.
```
ormsbee, line 887: Code in the `containers` applet shouldn't be querying Draft models directly, or ideally even be aware of components.

Where is this going to get called from?

ChrisChV: > Code in the containers applet shouldn't be querying Draft models directly, or ideally even be aware of components.

I've refactored the function to not use `Draft`

> Where is this going to get called from?

This function is called in `edx-platform`, in [get_library_container_draft_history](https://github.com/openedx/openedx-platform/pull/38178/changes#diff-5a248b18ae17c86d454776cb48e0372a8cbdc9610b3886f175a116f9b280e681R524) and in [get_library_container_publish_history](https://github.com/openedx/openedx-platform/pull/38178/changes#diff-5a248b18ae17c86d454776cb48e0372a8cbdc9610b3886f175a116f9b280e681R578).  This is what I have implemented, as I mentioned in [Slack](https://openedx.slack.com/archives/C06JG6RFE4V/p1776095502040809?thread_ts=1775833403.964179&cid=C06JG6RFE4V). It might change later after implement the `direct` field.

src/openedx_content/applets/publishing/api.py
```diff
@@ -566,6 +571,274 @@ def get_published_version(
...
+    if prev_published_at:
+        reset_filter["draft_change_log__changed_at__gt"] = prev_published_at
+
+    last_reset_at = (
+        DraftChangeLogRecord.objects
+        .filter(**reset_filter)
+        .order_by("-draft_change_log__changed_at")
+        .values_list("draft_change_log__changed_at", flat=True)
+        .first()
+    )
```
ormsbee, line 765: Per @sdaitzman, we should represent the Discard Changes action as another kind of draft update, so we don't need to filter them out here.

src/openedx_content/applets/containers/api.py
```diff
@@ -879,3 +880,48 @@ def get_container_children_entities_keys(container_version: ContainerVersion) ->
...
+
+    Intermediate containers (e.g. Subsections, Units) are never included in the
+    result; only leaf component entities are returned.
```
ormsbee, line 901: [Question] Why do we only focus on Components and exclude intermediate containers? Don't we need to include changes to intermediate Containers to build a historical view of the parent Container?

src/openedx_content/applets/containers/api.py
```diff
@@ -879,3 +880,48 @@ def get_container_children_entities_keys(container_version: ContainerVersion) ->
     return list(
         container_version.entity_list.entitylistrow_set.values_list("entity__key", flat=True).order_by("order_num")
     )
+
+
+def get_descendant_component_entity_ids(container: Container) -> list[int]:
```
ormsbee, line 895: I know you're framing this as descendants using container traversal, but publishing follows the (very closely related) concept of dependencies. Unless you absolutely need to frame this in terms of containers, it might make more sense to do dependency traversal similar to how we [traverse dependencies for publishing](https://github.com/openedx/openedx-core/blob/d0dc9cd6ac3dcbbbd083c19036684a8de88d7b46/src/openedx_content/applets/publishing/api.py#L390-L420). It will likely be faster as well.

One of the reasons I mention this is that we will eventually have dependencies that live outside of the container->child relationship, like [media assets](https://github.com/openedx/openedx-core/issues/499), and eventually course-related things like grading policies.

ormsbee (changes requested):
Partial review. Will add more in the morning. Thank you.

src/openedx_content/applets/containers/api.py
ChrisChV: This implementation follows the logic discussed in https://github.com/openedx/frontend-app-authoring/issues/2805#issuecomment-4092175050

> For Containers, both lists follow the same structure and contain:
>    - All direct edits made to the container itself.
>    - All edits included in each publish group, for all descendant components.
>
> E.g. for a section, an edit to a component grouped under a prior publish shows one edit, to that component (not to any in-between levels, or to the direct child subsection that is an ancestor of the component)

ormsbee: So does a rename of a Unit never show up at all?

ChrisChV: It is shown in the unit's history log:

`All direct edits made to the container itself.`

The logic I understood here is to show the important changes, which are the components, in the log. If we show only intermediate changes, for example, in a section's history log, it can display so many changes that it might overwhelm the user. 

The current implementation keeps it compact and with relevant information.

ormsbee: I guess I would expect it to show non-side-effect changes, i.e. ones where the actual version_num of the record changes. Because otherwise, we'd miss things like "Deleted this Unit", or "Removed this Subsectino" which are really big changes.

src/openedx_content/applets/publishing/api.py
```diff
@@ -566,6 +571,274 @@ def get_published_version(
...
+        entity_id = PublishableEntity.PublishableEntityID(publishable_entity_or_id)
+    else:
+        entity_id = publishable_entity_or_id.id
+
+    qs = (
+        DraftChangeLogRecord.objects
+        .filter(entity_id=entity_id)
+        .select_related(
+            "draft_change_log__changed_by",
+            "entity__component__component_type",
```
ormsbee: Please get rid of this line here, and add it on as a `.select_related()` when this gets called from `get_library_component_draft_history()`, like:

```python
    history_qs = content_api.get_entity_draft_history(component.publishable_entity)
    records = list(history_qs.select_related("entity__component__component_type")
```

Since `get_library_component_draft_history()` is aware of components anyway and is actively traversing them, I think it's better to keep that knowledge out of the publishing applet as much as possible.

src/openedx_content/applets/publishing/api.py
```diff
@@ -566,6 +571,274 @@ def get_published_version(
...
+        entity_id = PublishableEntity.PublishableEntityID(publishable_entity_or_id)
+    else:
+        entity_id = publishable_entity_or_id.id
+
+    return (
+        PublishLogRecord.objects
+        .filter(entity_id=entity_id)
+        .select_related(
+            "publish_log__published_by",
+            "entity__component__component_type",
```
ormsbee: Likewise, please remove this line and add it back in as a `select_related` call from `get_library_component_publish_history()`

ormsbee (changes requested):
Please mark the new API calls as "[ 🛑 UNSTABLE ]" in their docstrings, like the containers APIs are.

Otherwise, two very small change requests to keep component knowledge out of the publishing applet. I realize that we may have to revise this pending further UX refinement, but this is really excellent work—very clearly written and well commented. Great job!

src/openedx_content/applets/publishing/api.py
ChrisChV: Updated: https://github.com/openedx/openedx-core/pull/501/commits/5fa8b6b880deb68e22720776c84641a469700b3a

src/openedx_content/applets/publishing/api.py
ChrisChV: Updated: https://github.com/openedx/openedx-core/pull/501/commits/5fa8b6b880deb68e22720776c84641a469700b3a

ChrisChV:
> Please mark the new API calls as "[ 🛑 UNSTABLE ]" in their docstrings, like the containers APIs are.

Added in https://github.com/openedx/openedx-core/pull/501/commits/fd887b6a9b95dc39fd7a11f3784fecb61773491d

Also, I updated the version of the `openedx-core`: https://github.com/openedx/openedx-core/pull/501/commits/9339c314fee959985b41df78a3297384469af78a

ormsbee (approved):

PR #506, opened 2026-03-18
Tbain/253 add tags count

tbain:
## Description

This implements https://github.com/openedx/modular-learning/issues/253 , the task to add tag usage counts to the tags table under the taxonomies table. The frontend piece is where the results of this aggregation work is displayed is part of a separate pr to openedx/frontend-app-authoring. This change adds a  subquery annotation onto the django query for retrieving tags. The original implementation of the counts for tags only counted raw usage of each tag. This feature/PR aggregatea sum of any tag and child tag usage with sibling de-duplication for the same usage (e.g. when two sibling nodes are used against the same course, module, etc. we still only need to count that as '1' for any parent/grandparent nodes) as specified in the AC for the issue above, so it was replaced with this more complicated bit of logic that sums across tag usage based on various courses, sections, modules, and libraries that might use a tag.  

- update: 
The count logic is done in-memory, since we saw noticeable performance issues with trying to stay in the QuerySet/Django paradigm for calculating the counts. This makes the code a little less straightforward, since we break it out into a somewhat odd in-memory python application of the logic, but it still works as intended and resolves as many performance pain points as possible while still adhering to the counting requirements that end up necessitating such code.

- update: 
The count logic was moved out to the API level so that all the query data could complete and the logic to add the counts would not affect the ability to process the data as a QuerySet until the very end. This necessitated removing all the usage_count unit tests out of the _models and _api levels to the _view levels, so re-implemented that logic as appropriate (with a lot of AI help to speed it up)

AI Usage Disclosure: Claude was used via intelliJ IDE integration was used through the authoring process to work through complicated logic, craft the foundation of the unit tests, and also simplify it/make it more pythonic/alleviate performance concerns. 

## Supporting information

Github issue with AC: https://github.com/openedx/modular-learning/issues/253

## Testing instructions

Refer to the AC in the Github Issue. Steps to verify this is implemented and working via UX (Note, depends on the frontend part of this ticket): 

1. Navigate to the "Studio home" page
2. Navigate into an existing Course (or create a course and navigate into it)
3. In the "Course Outline" page, add tag(s) from an existing taxonomy to the course, module, or section. Ensure at least one of the tags you add is a sub-tag of a root tag.
4. Navigate back to the "Studio home" page
5. Click the "Taxonomies" tab to navigate to the Taxonomies page
6. Navigate into the Taxonomy that corresponds to the tag you added in step 3
7. Observe that, if a tag is used, there is now an additional column on the table named "Usage Count" that is populated with bubbles that display the count of tags usages, if applicable
8. Ensure that the tag you added in Step 3 properly associates the incremented count from its usage, and ensure that the usage count properly aggregates up the lineage based on the sub tag you selected in step 3

## Other information

Include anything else that will help reviewers and consumers understand the change.
- Does this change depend on other changes elsewhere?
   - this ticket is backwards compatible with the current implementation in frontend-app-authoring, since by default the frontend does not request the counts. 
- Any special concerns or limitations? For example: deprecations, migrations, security, or accessibility.
   - none at this time

### DIFF
--- a/src/openedx_core/__init__.py
+++ b/src/openedx_core/__init__.py
@@ -6,4 +6,4 @@ The public APIs belong to the specific apps (openedx_content, openedx_tagging, e
 """
 
 # The version for the entire repository
-__version__ = "0.39.1"
+__version__ = "0.39.2"
--- a/src/openedx_tagging/api.py
+++ b/src/openedx_tagging/api.py
@@ -530,7 +530,7 @@ def add_usage_counts(taxonomy: Taxonomy, tag_data: TagDataQuerySet) -> TagDataQu
     """
     Add usage counts to the query result.
 
-    Not a simple raw count of each tags uasge. A tag can be directly
+    Not a simple raw count of each tags usage. A tag can be directly
     applied to an object, which can be a course, library, module,
     or something else.
 


bradenmacdonald:
We're trying to stabilize the APIs for Verawood, so if we think we're ultimately going to end up with a second API endpoint for getting the counts, then I'd prefer to split that off separately now, even if we just use the existing implementation exactly as it is in this PR. In that case, it would definitely take less than 2 days, because you don't have to change much (although you could simplify it if you get time). You can also mark that "counts" endpoint as unstable so we can freely evolve it in Willow while keeping the "get tags" endpoint stable.

> I think that this is where it could potentially be valuable to add a filter to the tag count to only fetch ObjectTags where the Object corresponds to a course that is currently running or will be running in the future. I think the rest of the big picture for this is that for the folks who create new course runs every term and have very short terms, the usage count is probably meaningless for them and not very helpful anyways "Was it used 250 times or 275 times?

That all makes a lot of sense, but will require a lot of discussion, because the current API is not aware of "courses" as a concept at all, and I'm a bit reluctant to make the tagging API aware of those things - right now tagging is a very low-level feature that other things build on. If you're even considering functionality like that, then I think it's another reason to move the tag usage counts to a separate endpoint, where it can support more elaborate options/filtering.

@ormsbee Can I get your thoughts?

jesperhodge:
@bradenmacdonald if I understand correctly:

- if we are doing a separate endpoint, that can be done outside of this PR as ticket later (in Verawood? Willow?)
- if we are tackling this a different way - e.g. the way Braden suggested as option 1, or the way Mary suggested - this should be a separate discussion outside of this PR.
- We can "just use the existing implementation exactly as it is in this PR"

Did I get that correctly? So can we consider this PR unblocked in this case?
In that case I would like to move this conversation to a new ticket entirely (possibly a "discovery" ticket) and have @thelmick-unicon / @bradenmacdonald / Axim figure out what release and priority it should have.

bradenmacdonald:
I guess before we consider merging this as is, I'd like to know if the slowness mostly scales with taxonomy size or object tag count or both? If the slowness is only a factor on large taxonomies and it's just ~1s, I think that's OK for now. But if it's slow as the # of object tags increases or it's O(n_tags * n_object_tags) or anything like that, then it'll seem fine now and slow to a crawl in prod once people start using thousands of these things and re-running tagged courses.

bradenmacdonald:
FWIW Claude analyzed the query and says it could be slow. I have not had time to validate this analysis, so take with a grain of salt.

## The generated SQL (at depth=3)

```sql
SELECT ...,
  COALESCE(
    (SELECT COUNT(DISTINCT U0."object_id") AS "total_usage"
     FROM "oel_tagging_objecttag" U0
       INNER JOIN "oel_tagging_tag" U2 ON (U0."tag_id" = U2."id")
       LEFT OUTER JOIN "oel_tagging_tag" U3 ON (U2."parent_id" = U3."id")
       LEFT OUTER JOIN "oel_tagging_tag" U4 ON (U3."parent_id" = U4."id")
     WHERE U0."taxonomy_id" = 1
       AND (U0."tag_id" = outer."id"
            OR U2."parent_id" = outer."id"
            OR U3."parent_id" = outer."id"
            OR U4."parent_id" = outer."id")
    ), 0) AS "usage_count"
FROM "oel_tagging_tag"
WHERE "oel_tagging_tag"."taxonomy_id" = 1
```

## The scaling problem: it will get meaningfully slower

<p><strong>The old query</strong> filtered by <code>tag_id = OuterRef("pk")</code> — it used the FK index on <code style="font-family: monospace; color: rgb(59, 59, 59); background-color: rgba(0, 0, 0, 0.12); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">tag_id</code> and touched only the ~2-3 matching ObjectTag rows per tag. Cost per tag: O(direct_uses).</p><p><strong>The new query</strong> is a correlated subquery that, for <em>each</em> tag in the result set, does this:</p><ol style="padding-inline-start: 2em; color: rgb(59, 59, 59); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(248, 248, 248); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li>Range-scans<span> </span><strong>all</strong><span> </span>ObjectTags for the taxonomy (using the<span> </span><code style="font-family: monospace; color: rgb(59, 59, 59); background-color: rgba(0, 0, 0, 0.12); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">(taxonomy, object_id)</code><span> </span>index)</li><li>JOINs each to the tag table + parent chain (D PK lookups per row — fast)</li><li>Evaluates the OR across 4 different table aliases (can't use any index for this filter)</li><li>Counts distinct<span> </span><code style="font-family: monospace; color: rgb(59, 59, 59); background-color: rgba(0, 0, 0, 0.12); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">object_id</code><span> </span>values</li></ol><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(59, 59, 59); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(248, 248, 248); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Cost per tag: O(all_ObjectTags_in_taxonomy × D)</strong></p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(59, 59, 59); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(248, 248, 248); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">So the total work is roughly <strong>T × O × D</strong> where:</p><ul style="padding-inline-start: 2em; color: rgb(59, 59, 59); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(248, 248, 248); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li>T = tags in the result set</li><li>O = total ObjectTags for this taxonomy</li><li>D = max depth (small, ≤5)</li></ul>

Scenario | Tags (T) | ObjectTags (O) | Old cost (T × ~3) | New cost (T × O)
-- | -- | -- | -- | --
Small | 100 | 50 | 300 | 5,000
Medium | 500 | 1,000 | 1,500 | 500,000
Large | 1,000 | 10,000 | 3,000 | 10,000,000
Very large | 1,000 | 100,000 | 3,000 | 100,000,000

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(59, 59, 59); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(248, 248, 248); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>It scales linearly with ObjectTag count</strong>, but since it's inside a correlated subquery that runs per-tag, the multiplier is the number of tags displayed. This will be painfully slow once a popular taxonomy gets applied to thousands of courses/modules/sections.</p><h2 style="color: rgb(59, 59, 59); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(248, 248, 248); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Why the OR kills performance</h2><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(59, 59, 59); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(248, 248, 248); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The condition <code style="font-family: monospace; color: rgb(59, 59, 59); background-color: rgba(0, 0, 0, 0.12); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">tag_id = X OR parent_id = X OR parent.parent_id = X</code> spans different table aliases. The DB optimizer can't use a single index path — it has to evaluate all conditions per row after the JOINs. There's no composite index that helps here.</p>

tests/openedx_tagging/test_models.py
```diff
@@ -560,7 +567,217 @@ def test_usage_count(self) -> None:
...
+            "    Gastrotrich (Animalia) (used: 0, children: 0)",
+            "    Placozoa (Animalia) (used: 0, children: 0)",
+            "    Porifera (Animalia) (used: 0, children: 0)",
+        ]
+
+    def test_usage_count_three_levels_deep_rollup(self) -> None:
+        """
+        Tagging an object with a depth-3 tag (Chordata) should roll up
+        to grandparent (Animalia) and great-grandparent (Eukaryota),
+        verifying the full 3-level lineage query in add_counts_query.
```
bradenmacdonald: The description of this test is not accurate and references a function that was refactored/renamed.

```suggestion
        Tagging an object with a depth-3 tag (Chordata) as well as one
        of that tag's parents (Animalia) should roll up as only a single
        tag usage at the grandparent level, because the parent tag
        (Animalia) is implied by the child tag (Chordata) anyways.
```

tbain:
As per previous comments, moved the _add_counts logic out to the API level just before the return so that the logic to add the usage_counts doesn't interfere with the QuerySet logic until after all the filtering and paging logic is performed. This was necessitated by performance concerns with using something like a QuerySet annotation Case/When adding a multiplicative Big O complexity, so we had to work around that logic and manually add the value derived via Python logic to the list form of the return data rather than the QuerySet form which doesn't support such direct manipulation.

Since we completely moved the counting logic out of the the DB layers up to the API level, this required moving all of the unit testing to the appropriate level and removing it from the lower level unit testing suites. I used AI help to build out as extensive a unit test suite as I could think of in addition to maintaining parity with the tests I had to remove from the lower levels.

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
             existing_tag.refresh_from_db()
 
 
+class TestTaxonomyTagsUsageCount(TestTaxonomyViewMixin):
+    """
+    Tests the usage_count rollup logic in the taxonomy tags view
```
jesperhodge: I would like the tests to act as a sort of documentation of expected behavior, in general.
For anyone not deep into the Tag Usage Count logic, this will not be understandable.

Can we add the definition I wrote here that clarifies the behavior?

"
A tag can be directly applied to an object, which can be a course, library, module, or something else.
A tag can also be indirectly applied: when some of its children are applied to an object, it is considered automatically applied.
So, if tag "Chemistry" and tag "Physics" are applied once each to different objects, their parent tag "Natural Science" is considered indirectly applied to 2 objects.
Deduplication: A tag can only be applied to a single object once. So if two child tags are applied to the same object, the parent tag is only applied to it once, because no tag can be applied to the same object twice.
"

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+    #    |- Protista
+
+    def setUp(self):
+        super().setUp()
+        self.taxonomy = Taxonomy.objects.create(name="Usage Count Taxonomy")
+        self.taxonomy_url = TAXONOMY_TAGS_URL.format(pk=self.taxonomy.pk)
+
+    def test_usage_count_rollup(self):
+        """
+        Test that usage counts correctly roll up from children to parents,
```
jesperhodge: I'd like the test description also to be very clear. You can orient yourself at the definition I wrote above.

I would only use the term "roll up" if it's coupled with an explanation on what rollup is ("When descendant tags are applied to an object, the parent tag is considered automatically applied to the object, which may result in an increase of the usage count value of the parent" or something like that)

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+        # Animalia: applied to obj1 (via Arthropoda, Chordata, Cnidaria) and obj2 (via Arthropoda).
+        # Should be 2, because it counts '1' per object regardless of how many children are applied.
+        assert results["Animalia"]["usage_count"] == 2
+
+        # Eukaryota: same logic as Animalia -> count: 2
+        assert results["Eukaryota"]["usage_count"] == 2
+
+    def test_usage_count_rollup_multi_level(self):
+        """
+        Test that usage counts correctly roll up across more than two levels
```
jesperhodge: How do we know what's "correct" vs "incorrect"? Please specify.

tests/openedx_tagging/test_views.py
jesperhodge: Also please specify what the expected behavior here actually is (when is it "correct"?)

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+        self.client.force_authenticate(user=self.staff)
+
+        # --- Request all tags with counts ---
+        response = self.client.get(self.taxonomy_url + "?include_counts&full_depth_threshold=100")
+        assert response.status_code == status.HTTP_200_OK
+
+        results = {tag["value"]: tag for tag in response.data["results"]}
+
+        # --- Verification ---
+        # Arthropoda: applied to obj1, obj2 -> count: 2
```
jesperhodge, line 2653: I _love_ these comments that explain exactly what is expected.

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+        assert results["Arthropoda"]["usage_count"] == 2
+
+        # Chordata: applied to obj1 -> count: 1
+        assert results["Chordata"]["usage_count"] == 1
+
+        # Cnidaria: applied to obj1 -> count: 1
+        assert results["Cnidaria"]["usage_count"] == 1
+
+        # Animalia: applied to obj1 (via Arthropoda, Chordata, Cnidaria) and obj2 (via Arthropoda).
+        # Should be 2, because it counts '1' per object regardless of how many children are applied.
```
jesperhodge, line 2663: _Awesome_ explanation.

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+
+        self.client.force_authenticate(user=self.staff)
+
+        # --- Request tags with counts ---
+        response = self.client.get(self.taxonomy_url + "?include_counts&full_depth_threshold=100")
+        assert response.status_code == status.HTTP_200_OK
+        results = {tag["value"]: tag for tag in response.data["results"]}
+
+        # --- Verification ---
+        # Mammalia: obj1 -> 1
```
jesperhodge, line 2698: I wouldn't expect the reader of this test to necessarily have read the test above.
So, can you write these comments in the same way you did above so that anyone can understand?
I recommend just having AI do that so you don't need to rewrite every line yourself

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+
+        # --- Verification ---
+        assert results["Animalia"]["usage_count"] == 1
+        assert results["Fungi"]["usage_count"] == 1
+        # Eukaryota should have 2 because it's used on obj1 (via Animalia) and obj2 (via Fungi)
+        assert results["Eukaryota"]["usage_count"] == 2
+
+    def test_usage_count_max_depth_rollup(self):
+        """
+        Verify usage_count rollup up to the maximum depth of 7,
```
jesperhodge: It's great that you test this deeply.

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+        # Chordata: obj1 (via Mammalia), obj2 -> 2
+        assert results["Chordata"]["usage_count"] == 2
+        # Animalia: obj1 (via Chordata), obj2 (via Chordata) -> 2
+        assert results["Animalia"]["usage_count"] == 2
+        # Eukaryota: obj1 (via Animalia), obj2 (via Animalia) -> 2
+        assert results["Eukaryota"]["usage_count"] == 2
+
+    def test_usage_count_one_level_root_and_child_rollup(self):
+        """
+        Verify that usage counts roll up even when querying only a single level.
```
jesperhodge: Clarify a bit more?

tests/openedx_tagging/test_views.py
```diff
@@ -2446,6 +2446,425 @@ def test_delete_tag_in_taxonomy_without_subtags(self):
...
+            "    Placozoa (used: 0, children: 0)",
+            "    Porifera (used: 0, children: 0)",
+            "  Monera (used: 0, children: 0)",
+            "  Plantae (used: 1, children: 0)",
+            "  Protista (used: 0, children: 0)",
+        ]
+
+    def test_usage_count_sibling_and_ancestor_deduplication(self):
+        """
+        Test deduplication when multiple children of the same parent are applied to the same object.
```
jesperhodge, line 2977: This description sounds good to me, very understandable.

src/openedx_tagging/rest_api/v1/views.py
```diff
@@ -847,17 +849,54 @@ def get_queryset(self) -> TagDataQuerySet:
         )
         if depth == 1:
             # We're already returning just a single level. It will be paginated normally.
+            if include_counts:
+                return self._add_counts(results)
```
jesperhodge: Very minor nit to be clear what this returns:

```suggestion
                results_with_counts = self._add_counts(results)
                return results_with_counts
```

src/openedx_tagging/rest_api/v1/views.py
```diff
@@ -847,17 +849,54 @@ def get_queryset(self) -> TagDataQuerySet:
...
             return results
         elif full_depth_threshold and len(results) < full_depth_threshold:
             # We can load and display all the tags in this (sub)tree at once:
             self.pagination_class = DisabledTagsPagination
+            if include_counts:
+                return self._add_counts(results)
```
jesperhodge:
```suggestion
                results_with_counts = self._add_counts(results)
                return results_with_counts
```

src/openedx_tagging/rest_api/v1/views.py
```diff
@@ -847,17 +849,54 @@ def get_queryset(self) -> TagDataQuerySet:
...
             return results
         else:
             # We had to do a deep query, but we will only return one level of results.
             # This is because the user did not request a deep response (via full_depth_threshold) or the result was too
             # large (larger than the threshold).
             # It will be paginated normally.
-            return results.filter(parent_value=parent_tag_value)
+            filtered_results = results.filter(parent_value=parent_tag_value)
+            if include_counts:
+                return self._add_counts(filtered_results)
```
jesperhodge:
```suggestion
                results_with_counts = self._add_counts(results)
                return results_with_counts
```

src/openedx_tagging/rest_api/v1/views.py
jesperhodge: This is a bit subjective, some might not like the suggestion. If @bradenmacdonald you like it please thumbs up this.
If Braden does not thumbs up this please do not implement it @tbain .

src/openedx_tagging/rest_api/v1/views.py
jesperhodge: This is a bit subjective, some might not like the suggestion. If @bradenmacdonald you like it please thumbs up this.
If Braden does not thumbs up this please do not implement it @tbain .

src/openedx_tagging/rest_api/v1/views.py
jesperhodge: This is a bit subjective, some might not like the suggestion. If @bradenmacdonald you like it please thumbs up this.
If Braden does not thumbs up this please do not implement it @tbain .

jesperhodge (approved):
Hey @tbain this is amazing work. I tested it for correctness and performance, and I love the very extensive tests - that's just great.

Please regard every comment from me as a nit, they are very minor, and I think should not block the PR from being merged at all. That said, it would be nice to have the improvements for the comments in here; but merging this very soon has priority due to the Verawood release.

Can you make sure to bump the version (just a patch I think, not a minor or major version bump)?

Probably squashing can be done while merging.

I'll approve but final approve should come from Braden or Dave

src/openedx_tagging/rest_api/v1/views.py
bradenmacdonald: I don't have a strong opinion here but I guess I do like your suggestion a bit more, as it's more clear.

src/openedx_tagging/rest_api/v1/views.py
```diff
@@ -847,17 +849,54 @@ def get_queryset(self) -> TagDataQuerySet:
...
+
+            return filtered_results
+
+    def _add_counts(self, tag_data: TagDataQuerySet) -> TagDataQuerySet:
```
bradenmacdonald: Nit: this could be a top-level function in `src/openedx_tagging/api.py`.

Nit: Also, along the lines of @jesperhodge 's comments, it would be good to explain the de-duplication here or at lest say "refer to test case X for examples and details". (Or put clear details here and have the test cases refer back here).

src/openedx_tagging/models/base.py
```diff
@@ -480,7 +479,7 @@ def get_filtered_tags(  # pylint: disable=too-many-positional-arguments
     def _get_filtered_tags_free_text(
         self,
         search_term: str | None,
-        include_counts: bool,
+        include_counts: bool,  # pylint: disable=unused-argument
```
bradenmacdonald: Why are we keeping these `include_counts` parameters if they don't work? I think you can just remove them (from here and from the function that calls these ones).

There is no code referencing `include_counts` in openedx-platform as far as I can tell, so even though it's a breaking change, it's fine to include such changes in our 0.x version of openedx-core.

tbain: Yeah, I wasn't sure about keeping these or not. I know it's probably a "you aren't going to need it" type thing, but for completions sake, I felt like it was worth continuing them down to this level since they are a param in the query. I'm perfectly fine removing them though, I was 50/50 either way.

src/openedx_core/__init__.py
```diff
@@ -6,4 +6,4 @@
 """
 
 # The version for the entire repository
-__version__ = "0.38.2"
+__version__ = "0.38.3"
```
jesperhodge: Should be `0.39.2` now I believe

src/openedx_tagging/api.py
```diff
@@ -525,3 +524,49 @@ def unmark_copied_tags(object_id: str) -> None:
     Update copied object tags on the given object to mark them as "not copied".
     """
     ObjectTag.objects.filter(object_id=object_id).update(is_copied=False)
+
+
+def add_usage_counts(taxonomy: Taxonomy, tag_data: TagDataQuerySet) -> TagDataQuerySet:
+    """
+    Add usage counts to the query result.
+
+    Not a simple raw count of each tags uasge. A tag can be directly
```
jesperhodge: `usage`

PR #507, opened 2026-03-19
fix: return correct taxonomy depth up to max depth

jesperhodge:
Fixes https://github.com/openedx/modular-learning/issues/257.


bradenmacdonald:
@jesperhodge Thanks for this fix. I think my new PR https://github.com/openedx/openedx-core/pull/510 fixes the same issue and produces the same result for the specific bug you found, but also resolves all the other inconsistencies that you've mentioned here.

If for whatever reason we can't merge it on time, I'm open to merging this as a fallback.

PR #521, opened 2026-03-30
Improve validation of openedx_content to prevent creation of corrupt states

bradenmacdonald:
This PR implements tests for four of the issues identified in https://github.com/openedx/openedx-core/issues/463 .

1. `set_draft_version()` should reject a `PublishableEntityVersion` that belongs to a different `PublishableEntity`.
2. `publish_from_drafts()` should reject drafts that don't belong to the specified `LearningPackage`.
3. `create_publishable_entity_version()` should reject `dependencies` that are from a different `LearningPackage`.
4. `publish_all_drafts()` and `publish_from_drafts()` should not be callable from within a `bulk_draft_changes_for()` context.

Another issue is in a separate PR for simplicity: https://github.com/openedx/openedx-core/pull/569

Two issues are unaddressed as they are API design choices: `create_publishable_entity_version()` and `create_publishable_entity()` can both create either a useless or an invalid state if they are used on their own without corresponding [Component/Container/etc.] versions.

Platform: no corresponding changes are required in openedx-platform as far as I can tell.

### DIFF
--- a/src/openedx_content/applets/publishing/api.py
+++ b/src/openedx_content/applets/publishing/api.py
@@ -511,6 +511,9 @@ def publish_from_drafts(
                 # Skip duplicates that we might get from expanding dependencies.
                 if draft.pk in published_draft_ids:
                     continue
+                # Validate Learning Package here where it won't require any extra queries
+                if draft.entity.learning_package_id != learning_package_id:
+                    raise ValidationError("Draft entity is from a different learning package.")
 
                 try:
                     old_version = draft.entity.published.version
--- a/src/openedx_content/applets/publishing/api.py
+++ b/src/openedx_content/applets/publishing/api.py
@@ -932,6 +932,15 @@ def set_draft_version(
         # block is bookkeeping in our DraftChangeLog.
         draft.version_id = publishable_entity_version_pk
 
+        # Validate the entity
+        if publishable_entity_version_pk is not None:
+            if draft.entity.id != PublishableEntityVersion.objects.only("entity_id").get(
+                pk=publishable_entity_version_pk
+            ).entity_id:
+                raise ValidationError(
+                    "Entity mismatch - the specified PublishableEntityVersion does not match the PublishableEntity"
+                )
+
         # Check to see if we're inside a context manager for an active
         # DraftChangeLog (i.e. what happens if the caller is using the public
         # bulk_draft_changes_for() API call), or if we have to make our own.
--- a/src/openedx_content/applets/publishing/api.py
+++ b/src/openedx_content/applets/publishing/api.py
@@ -515,7 +515,10 @@ def publish_from_drafts(
                     continue
                 # Validate Learning Package here where it won't require any extra queries
                 if draft.entity.learning_package_id != learning_package_id:
-                    raise ValidationError("Draft entity is from a different learning package.")
+                    raise ValidationError(
+                        f"Draft entity (id={draft.entity.id}) is from learning package "
+                        f"{draft.entity.learning_package_id}; expected learning package {learning_package_id}."
+                    )
 
                 try:
                     old_version = draft.entity.published.version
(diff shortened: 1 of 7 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

tests/openedx_content/applets/publishing/test_api.py
```diff
@@ -415,11 +428,11 @@ def test_reset_drafts_to_published(self) -> None:
                 pub_version_num
             )
 
-    def test_reset_drafts_to_published_bulk(self) -> None:
-        """bulk_draft_changes_for creates only one DraftChangeLog."""
-        with publishing_api.bulk_draft_changes_for(self.learning_package_1.id):
-            self.test_reset_drafts_to_published()
-        assert DraftChangeLog.objects.count() == 1
```
bradenmacdonald, line 422: Since it's hard to tell from the diff what's going on here:

This `test_reset_drafts_to_published_bulk` just called `test_reset_drafts_to_published()` inside `bulk_draft_changes_for` but that inner test does some setup, a publish, more setup, then a revert, so this was causing an error - we don't want to allow publishes inside `bulk_draft_changes_for`.

The refactored test is almost exactly the same, but is parametrized and when `bulk=True` it wraps the pre-publish setup in a bulk transaction, and the post-publish setup+revert in a second bulk transaction.

I think this is an even better test now because we can assert that the net effect of the second bulk transaction is zero, as it reverts to the published version.

My local git client shows the diff more cleanly than GitHub:
Image

src/openedx_content/applets/publishing/api.py
```diff
@@ -503,6 +513,9 @@ def publish_from_drafts(
                 # Skip duplicates that we might get from expanding dependencies.
                 if draft.pk in published_draft_ids:
                     continue
+                # Validate Learning Package here where it won't require any extra queries
+                if draft.entity.learning_package_id != learning_package_id:
+                    raise ValidationError("Draft entity is from a different learning package.")
```
ormsbee: Nit: In these sorts of cases, I think it's useful to put the expected and actual package IDs, for easier debugging.

src/openedx_content/applets/publishing/api.py
```diff
@@ -919,6 +932,15 @@ def set_draft_version(
         # block is bookkeeping in our DraftChangeLog.
         draft.version_id = publishable_entity_version_pk
 
+        # Validate the entity
+        if publishable_entity_version_pk is not None:
+            if draft.entity.id != PublishableEntityVersion.objects.only("entity_id").get(
+                pk=publishable_entity_version_pk
+            ).entity_id:
+                raise ValidationError(
+                    "Entity mismatch - the specified PublishableEntityVersion does not match the PublishableEntity"
```
ormsbee: For this one, I think adding a repr() of the objects involved would be helpful when debugging, since if this ever happens, things have gotten seriously weird.

src/openedx_content/applets/publishing/api.py
```diff
@@ -13,7 +13,7 @@
 from typing import ContextManager, Optional, cast
 
 from django.contrib.auth import get_user_model
-from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
```
ormsbee, line 16: [Comment]: I've generally used `ValueError` instead of `ValidationError` for invalid arguments, though I realize that doesn't make sense for things like checking that you're not in a bulk change context before publishing. Should I switch over, do you think?

ormsbee (changes requested):
Minor request on more detail in the errors being raised.

src/openedx_content/applets/publishing/api.py
bradenmacdonald: For these errors that are more developer-facing (you coded something wrong, not the user gave invalid input), I honestly don't think it makes much difference. I'm on the fence and happy to change to `ValueError` too if you'd like.

The python docs say

> ValueError: Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception

So I guess it depends if you consider `ValidationError` to be more precise. We could also define a custom exception that's even more specific, but I don't think it's worth it if we expect these to be pretty rare.

bradenmacdonald:
@ormsbee I added in that extra logging you requested.

ormsbee (approved):

ormsbee:
@bradenmacdonald, @kdmccormick: Just a heads up that something here is causing test failures in the modulestore migrator code.

PR #527, opened 2026-04-01
The Big Keys Rework (OEP-68)

kdmccormick:
This applies the guidance about Code, OpaqueKeys, and refs from https://github.com/openedx/openedx-proposals/pull/773. Please see the individual commits for details.


src/openedx_django_lib/fields.py
```diff
@@ -114,6 +117,29 @@ def immutable_uuid_field() -> models.UUIDField:
...
+
+def code_field(**kwargs) -> MultiCollationCharField:
+    """
+    Field to hold a 'code', i.e. a slug-like local identifier.
+    """
+    return case_sensitive_char_field(
+        max_length=255,
+        blank=False,
+        validators=[
+            RegexValidator(
```
bradenmacdonald, line 141: Django now supports doing regex validations via database constraints, like: 

https://github.com/openedx/openedx-core/blob/5391abc5233b4dfd8c675e25dc8400c468285c2f/src/openedx_catalog/models/catalog_course.py#L238-L239

This has the advantage that it's enforced at the database level, with perfect consistency, which is much better because Django validators are not normally called on `.update()` or `.save()`. But if you define a check constraint, not only does it run at the database level, but Django _also_ runs it as a python-level validator automatically, so you get all the same advantages of a validator.

The only thing is, I'm not sure offhand if there's any way to specify constraints like this at the field level, only at the model level 😞

src/openedx_content/applets/containers/models.py
```diff
@@ -171,6 +172,12 @@ class Container(PublishableEntityMixin):
     olx_tag_name: str = ""
     _type_instance: ContainerType  # Cache used by get_container_type()
 
+    # This foreign key is technically redundant because we're already locked to
+    # a single LearningPackage through our publishable_entity relation. However,
+    # having this foreign key directly allows us to make indexes that efficiently
+    # query by other Container fields within a given LearningPackage.
+    learning_package = models.ForeignKey(LearningPackage, on_delete=models.CASCADE)
```
bradenmacdonald, line 179: I know we already have an identical field for `Component`. I really wish we had some way to enforce that these non-normalized fields are kept up to date.

kdmccormick:
Closed in favor of smaller PRs, listed on https://github.com/openedx/openedx-core/issues/322

PR #534, opened 2026-04-07
Add type annotations for primary key types by model [FC-0117]

bradenmacdonald:
This PR has some experimental implementations of https://github.com/openedx/openedx-core/issues/436 as well as moving closer toward the consistent terms preferred in OEP-68.

### Second Approach: fully-typed `.id`, avoid `.pk`

(Revised from my initial attempt)

* `PublishableEntity`, `LearningPackage`, `Container`, and `Component` all get a fully-typed `.id` field.
   - `PublishableEntity.id` is of type `PublishableEntity.PK` which is a sub-type of `int`.
   - `Container.id` is of type `Container.PK` which is a sub-type of `PublishableEntity.PK`.
   - `Component.id` is of type `Component.PK` which is a sub-type of `PublishableEntity.PK`.
   - `LearningPackage.id` is of type `LearningPackage.PK` which is a sub-type of `int`.
   - Thanks to django-stubs, these types generally get propagated and used correctly pretty much everywhere.
   - Example: <br>Screenshot 2026-04-07 at 2 18 09 PM
* Likewise, `CatalogCourse` and `CourseRun` get typed.
* Unfortunately, it is impossible to change the type of `.pk` on models that inherit `PublishableEntityMixin` (or any other models that use a `OneToOneField` as their `primary_key`, at least without modifying django-stubs or implementing a mypy plugin. It will always have type `Any`, circumventing all type definitions. However, although we cannot override its type, we can mark it as `deprecated`. As a result, and for consistency, accessing `.pk` on any of these will raise a deprecation warning in pytest/console and show a warning in your IDE:
     Screenshot 2026-04-07 at 2 11 26 PM
* I did _not_ (yet?) add a deprecation warning for `CatalogCourse.pk` and `CourseRun.pk`, because (a) they are typed correctly, and (b) mypy crashes when I do 😬.

### Prior Approach: fully-typed `.pk`, avoid `.id`

<details><summary>Expand for pros/cons/details of this earlier approach</summary>
In this earlier approach:

* `PublishableEntity`'s primary key is the `.pk` field, and it has type `PublishableEntity.PK` which is a sub-type of `int`.
   - Example: Screenshot 2026-04-07 at 12 59 30 PM
   - In the approach I took, accessing `some_entity.id` will show a deprecation warning, telling you to use `pk` instead. And since this is an actual field called `pk`, any DB lookups need to be changed from e.g. `entity__id=...` to `entity__pk=...`
      pk
   - A slightly less disruptive approach would be to keep the ID field called `id` and make an alias called `entity_pk`, then use deprecation warnings to encourage the use of `entity_pk` over `id` and `pk`; this would be more consistent with what I did for containers and components, but it's slightly annoying to type `entity_pk` instead of `pk` everywhere.
   - Keeping the ID field called `id` and making `pk` a fully-typed alias is unfortunately not possible, as django-stubs forcibly overrides the type of `pk` unless you have a concrete field named `pk` as I did here.

* `Container` and `Component` get `Container.PK` and `Component.PK` types which are sub-types of `PublishableEntity.PK`
* `Container` and `Component` cannot have correctly-typed `.pk` accessors due to django-stubs limitations, so I created a `.container_pk` and `.component_pk` accessor for them:
    container_pk
   - Details: because `Container` and `Component` have a `OneToOneField(PublishsableEntity, primary_key=True)`, django-stubs will always override their `pk` attribute to have type `Any`, and it's impossible to change this without modifying django-stubs or writing our own mypy plugin. _But_, we can use the `@deprecated` decorator to annotate the attribute with a deprecation warning which does show up in the IDE and in the pytest logs wherever `.pk` is used.
   - I think this is clear and explicit, but would something shorter be better (`.pk_` ?)

</details>

-----------------------------------------

### General Notes

One downside of using deprecation warnings as I have here is that Django itself uses `.pk`, so in the case of `Container` and `Component` (but not `PublishableEntity`) where we've marked `.pk` as deprecated, even if our code doesn't use `.pk` itself, you'll see warnings like this when running pytest: `django/db/models/fields/related.py:812: DeprecationWarning: Use component_pk instead`. It is possible to suppress these deprecation warnings, and I have done so (see `tox.ini`) but it's an extra step required to use this library well and I'm not a big fan of it.


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

bradenmacdonald:
@ormsbee @kdmccormick Can I get your thoughts on this? It also has implications for OEP-68.

src/openedx_content/applets/components/models.py
```diff
@@ -123,6 +132,24 @@ class Component(PublishableEntityMixin):
...
+
+    @property
+    @deprecated("Use .id instead")
+    def pk(self):
+        """Mark the .pk attribute as deprecated"""
+        # Note: Django-Stubs forces mypy to identify the `.pk` attribute of this model as having 'Any' type (due to our
+        # use of a OneToOneField primary key), and this is impossible for us to override, so we prefer to use
+        # `.id` which we can control fully.
+        # Since Django uses '.pk' internally, we have to make sure it still works, however. So the best we can do is
+        # override this with a deprecated marker, so it shows a warning in developer's IDEs like VS Code.
```
ormsbee, line 148: Ouch. Very timely to know this given the [OEP in flight](https://github.com/openedx/openedx-proposals/pull/773) on identifiers, where we recommend using `.pk`.

bradenmacdonald: Yes, as you can read above in the edited description, I actually changed my approach here, as it turns out it's much easier to override the type of `.id` than `.pk` in general. So I'm wondering if we should just recommend `.id` / `_id` more.

src/openedx_content/applets/publishing/models/id_fields.py
```diff
@@ -0,0 +1,16 @@
...
+from django.db import models
+
+PublishableEntityPK = NewType("PublishableEntityPK", int)
+
+
+class PublishableEntityPKField(models.BigAutoField):
+    """
+    Superficial subclass of models.BigAutoField that tells mypy/django-stubs to
+    use `PublishableEntityPK` instead of `int`
+    """
```
bradenmacdonald: Note: as far as I can tell, it is not possible to make this generic, e.g. to declare a `TypedPKField[arg]` where `arg` can be `PublishableEntityPK`, `ContainerPK`, `ComponentPK`, etc.

This is because when django-stubs evaluates the types of this field (specifically when it looks at the `_pyi_private_set_type` etc. in the `.pyi` file), it "reads the declared type of _pyi_private_get_type straight off the class's TypeInfo and returns it verbatim — no expand_type / no TypeVar substitution." and it ends up injecting an _unbound_ TypeVar `arg` as the type of the field values, not evaluating `arg` to get the actual type we want. So later you get errors like `expected 'arg' but got 'PublishableEntity.PK'`.

So: one downside of this approach is the necessity of declaring these `PublishableEntityPKField` style classes for each major "type" of primary key. (Although for `Container`/`Component` it's not necessary as they don't have their own separate ID fields.)

Edit: see later top-level comment on this PR. I did find an approach that is much nicer and uses a shared, generic field, by using one additional layer of indirection.

bradenmacdonald: I had a thought that maybe it would be possible to make this "typed ID field" into a generic field by using a _boxed_ ID type instead of using `NewType` to define specialized `int` sub-types. And I explored that approach with Claude. It does seem like it's possible to do it generically using such an approach. However, because these things are used as primary keys in various ways, you end up needing so much code that's tied to Django and django-stubs internals to force everything to work properly that I don't think it's a good approach: https://github.com/openedx/openedx-core/compare/braden/typed-pks-alternate

src/openedx_content/applets/components/models.py
kdmccormick: oof, good find. I've opened a thread here: https://github.com/openedx/openedx-proposals/pull/773#discussion_r3052394999

bradenmacdonald:
_If_ we want to also introduce typed keys for Unit/Section/Subsection subclasses, that's easy to do: https://github.com/openedx/openedx-core/pull/535 . Would welcome thoughts on this as I'm a bit on the fence.

src/openedx_content/applets/components/models.py
kdmccormick: If we standardize on `.id`, then I presume we can drop this deprecation warning? I think most developers naturally use `.id` anyway.

src/openedx_content/applets/containers/models.py
```diff
@@ -161,6 +165,8 @@ class Container(PublishableEntityMixin):
     entities for different learners or at different times.
     """
 
+    PK: TypeAlias = ContainerPK
```
kdmccormick: If we standardize on `id`, I think it'd make sense to call the types `Container.ID`, `ContainerID`, etc.

src/openedx_content/applets/components/models.py
bradenmacdonald: I found that `.pk` was being used in quite a few places in this codebase, and the deprecation warnings made it easy for me to track them down and replace them with `.id`.

src/openedx_content/applets/containers/models.py
bradenmacdonald: Sure, updated to `Container.ID`/`ContainerID` etc.

bradenmacdonald:
OK, good news! I found a nicer way to do this. It still requires a custom field type per model, but we can put the definitions entirely within the model class itself, and there is now one global `id_fields` module in `openedx_django_lib` rather than needing `id_fields.py` and `id_fields.pyi` alongside each app[let]'s `models`.

e.g. within the `class LearningPackage(models.Model):` definition:

https://github.com/openedx/openedx-core/blob/959f9e6d4e4c6812d69e6d549fc07681a77fa848/src/openedx_content/applets/publishing/models/learning_package.py#L32-L38

Note:

This could be made even more concise, and copy-pastable without changes:

```python
ID = NewType("ID", int) 
class IDField(TypedBigAutoField[ID]): ...
id = IDField(primary_key=True) 
```
~~But: this creates ambiguous type names (what is "ID", without context?) in the mypy output, and the use of an ellipsis instead of `pass` implies that something is incomplete. So I'm recommending the more verbose version as our new standard boilerplate.~~

Edit: actually this works great, so I think I prefer the second version now. If you have several types called `ID`, mypy distinguished them in the output like so:

```
error: Argument 1 to "update_learning_package" has incompatible type
"openedx_content.applets.publishing.models.learning_package.LearningPackage.ID"; expected
"openedx_content.applets.publishing.models.publishable_entity.PublishableEntity.ID"  [arg-type]
```

kdmccormick:
@bradenmacdonald woohoo! The TypedBigAutoField is a huge improvement. I was fiddling with my own version of TypedBigAutoField last night and decided that it must be impossible :P Nice work on the `__class_getitem__` hack.

> Edit: actually this works great, so I think I prefer the second version now. If you have several types called ID, mypy distinguished them in the output like so:
> ```
> error: Argument 1 to "update_learning_package" has incompatible type
> "openedx_content.applets.publishing.models.learning_package.LearningPackage.ID"; expected
> "openedx_content.applets.publishing.models.publishable_entity.PublishableEntity.ID"  [arg-type]
> ```

I'm a little iffy on this--this is a pretty painful error message to read. I think I'd prefer more boilerplate in exchange for more-legible messages:

```
error: Argument 1 to "update_learning_package" has incompatible type LearningPackageID; expected PublishableEntityID  [arg-type]
```

bradenmacdonald:
@kdmccormick OK, good call. I changed that back. It now looks like: `error: Argument 1 to "get_entities_with_unpublished_changes" has incompatible type "LearningPackageID"; expected "PublishableEntityID"  [arg-type]`

kdmccormick (approved):
LGTM , just one question about next steps.

EDIT: GitHub lost my comment 🤦🏻 I'll rewrite it after my meeting....

bradenmacdonald:
@kdmccormick Here is the corresponding platform PR which shows the required changes (pretty minimal) plus demonstrates using the new field to make `StagedContent` also use fully-typed IDs.

https://github.com/openedx/openedx-platform/pull/38308

src/openedx_django_lib/id_fields.py
```diff
@@ -0,0 +1,57 @@
...
+
+    ```
+    class MyModel(models.Model):
+        MyModelID = NewType("MyModelID", int)
+        type ID = MyModelID
+
+        class IDField(TypedBigAutoField[ID]):
+            pass
+
+        id = IDField(primary_key=True)
```
kdmccormick, line 21: Not for this PR, but as a possible followup: at the expense of some more boilerplate, what do you think about extending this same system to model relationships?
```suggestion
        # Boilerplate for strongly-typed model relationships.
        # Replace "MyModel" with your model's name.
        MyModelID = NewType("MyModelID", int)
        type ID = MyModelID
        class IDField(TypedBigAutoField[ID]):
            pass
        class ForeignKey(TypedForeignKey[ID]):
            def __init__(self, *args, **kwargs):
                super().__init__("MyModel", *args, **kwargs)
        class OneToOneField(TypedOneToOneField[ID])
            def __init__(self, *args, **kwargs):
                super().__init__("MyModel", *args, **kwargs)

        id = IDField(primary_key=True)
        
        # To make references to other strongly-typed models: 
        my_related_object = MyOtherModel.ForeignKey(on_delete=..., ...)
        my_one_to_one_object = MyAdditionalModel.OneToOneField(on_delete=..., ...)
        # etc.
```

maybe there's even some way to shove `super().__init__("MyModel", *args, **kwargs)` into `TypedForeignKey/TypedOneToOneField`, so the body of all three field classes can be `pass`.

bradenmacdonald: @kdmccormick It's not necessary! The django-stubs plugin should already be providing the correct types for all the foreign key `_id` fields 😎

Please let me know if you observe any exceptions or if I'm wrong.

kdmccormick: 🤯

src/openedx_content/migrations/0006_typed_ids.py
```diff
@@ -0,0 +1,29 @@
...
+        ("openedx_content", "0005_containertypes"),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            state_operations=[
+                migrations.AlterField(
+                    model_name="publishableentity",
+                    name="id",
+                    field=models.BigAutoField(primary_key=True, serialize=False),
```
bradenmacdonald, line 24: Just noting here that changing to these fully-typed primary keys does not result in any migration if the model previous had an explicit `id` field. But if it had only the implicit, auto-created field, then adding an explicit one results in this kind of migration. By default such migrations will re-create the table in order to re-create the ID column, even though nothing has changed, so it's important to catch that and tweak the migration like this to not actually make any DB changes.

bradenmacdonald:
I made two final tiny fixes as a result of testing this in openedx-platform: 085393d718420eb3a1fbdba5c2153d98a5aab29e and bd6867105d87e79a9eca4cf03fb06768f9866728 . I'm going to squash this now to get all the checks passing.

kdmccormick (approved):
Just reaffirming

kdmccormick:
(don't forget to bump the version)

bradenmacdonald:
> Just reaffirming

Thanks, I was just holding off on the merge in hopes of getting two PRs merged into one version bump, but I don't think the other is ready yet as I hoped so I'm just going to merge this now.

PR #536, opened 2026-04-08
Fix integrity error handling

### LINKED ISSUE #261: Fix API integrity on rename
- When renaming a tag to a name that's duplicate, I just want to get a 4xx error, not an HTML integrity error page
- The API generally should hide internal errors from the end user and just send 500 if something unexpected / unhandled happens

Image

jesperhodge:
Fixes https://github.com/openedx/modular-learning/issues/261

Idea right now:

* Validate against duplicate values before they reach the database and cause an IntegrityError (this is not implemented here yet)
* Handle any uncaught internal errors, including integrity errors, throughout content-tagging. For security reasons we do not want to expose internal error information directly so we return a generic 500 response.
* I made a custom error handling mixin so that this is handled for the content_tagging api in general.

Question: This would be better handled at the CMS level?

Answer: There is ongoing work to fix this up, starting with a proposal. See https://openedx.slack.com/archives/CHYH0BDTR/p1775053660743269 and https://github.com/openedx/openedx-platform/pull/38246 in particular.

### DIFF
--- a/src/openedx_tagging/rest_api/v1/exception_handlers.py
+++ b/src/openedx_tagging/rest_api/v1/exception_handlers.py
@@ -31,8 +31,9 @@ def _custom_exception_handler(exc: Exception, context: dict) -> Response | None:
     if is_expected_exception:
         return exception_handler(exc, context)
 
-    # Making really sure that any unexpected exception gets logged
-    log.error(exc, stack_info=True, exc_info=True)
+    # DRF always calls exception handlers from within an `except:` block, so we can assume
+    # that `log.exception` will automatically insert those exception details and a stack trace.
+    log.exception("Unexpected exception while handling API request")
 
     if settings.DEBUG:
         description_with_traceback = f"{exc.__class__.__name__}: {str(exc)}\n\nTraceback:\n{traceback.format_exc()}"
--- a/src/openedx_tagging/rest_api/v1/exception_handlers.py
+++ b/src/openedx_tagging/rest_api/v1/exception_handlers.py
@@ -1,0 +1,57 @@
+"""
+Tagging REST API exception handling utilities.
+"""
+
+import logging
+import traceback
+
+from django.conf import settings
+from django.http import Http404
+from rest_framework import status
+from rest_framework.exceptions import APIException, PermissionDenied
+from rest_framework.response import Response
+from rest_framework.views import exception_handler
+
+log = logging.getLogger(__name__)
+
+
+def _custom_exception_handler(exc: Exception, context: dict) -> Response | None:
+    """
+    Return standard DRF errors for APIException and a generic 500 otherwise.
+    This exception handler should eventually be replaced by a more top-level
+    exception handler in the openedx-platform repo after the ADR for it is accepted:
+    https://github.com/openedx/openedx-platform/pull/38246
+    """
+    # For exceptions expected by DRF return the standard DRF error response:
+    # Instances of APIException, subclasses of APIException, Django's Http404 exception,
+    # and Django's PermissionDenied exception.
+    is_expected_exception = isinstance(
+        exc, (APIException, Http404, PermissionDenied)
+    )
+    if is_expected_exception:
+        return exception_handler(exc, context)
+
+    # Making really sure that any unexpected exception gets logged
+    log.error(exc, stack_info=True, exc_info=True)
+
+    if settings.DEBUG:
+        description_with_traceback = f"{exc.__class__.__name__}: {str(exc)}\n\nTraceback:\n{traceback.format_exc()}"
+
+        return Response(
+            {"detail": description_with_traceback},
+            status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+        )
+
+    return Response(
+        {"detail": "An unexpected error occurred while processing the request."},
+        status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+    )
+
+
+class TaggingExceptionHandlerMixin:
+    """
+    Scope custom exception handling to tagging API views only.
+    """
+
+    def get_exception_handler(self):
+        return _custom_exception_handler
--- a/src/openedx_tagging/rest_api/v1/serializers.py
+++ b/src/openedx_tagging/rest_api/v1/serializers.py
@@ -13,6 +13,7 @@ from rest_framework.reverse import reverse
 from openedx_tagging.data import TagData
 from openedx_tagging.import_export.parsers import ParserFormat
 from openedx_tagging.models import ObjectTag, Tag, TagImportTask, Taxonomy
+from openedx_tagging.models.utils import RESERVED_TAG_CHARS
 from openedx_tagging.rules import ObjectTagPermissionItem
 
 from ..utils import UserPermissionsHelper
@@ -59,7 +60,7 @@ class UserPermissionsSerializerMixin(UserPermissionsHelper):
     @property
     def _request(self) -> Request:
         """
-        Returns the current request from the serialize context.
+        Returns the current request from the serializer context.
         """
         return self.context.get('request')  # type: ignore[attr-defined]
 
@@ -217,6 +218,27 @@ class ObjectTagUpdateBodySerializer(serializers.Serializer):  # pylint: disable=
     tagsData = serializers.ListField(child=ObjectTagUpdateByTaxonomySerializer(), required=True)
 
 
+def validate_tag_value(value, context):
+    """
+    Validate this tag value is unique within the current taxonomy context and
+    does not contain forbidden characters.
+    """
+    taxonomy_id = context.get("taxonomy_id")
+    if taxonomy_id is not None:
+        # Check if tag value already exists within this taxonomy. If so, raise a validation error.
+        queryset = Tag.objects.filter(taxonomy_id=taxonomy_id, value=value)
+        if queryset.exists():
+            raise serializers.ValidationError(
+                f'Tag value "{value}" already exists in this taxonomy.', code='unique'
+            )
+
+    # validator checks there are no forbidden characters ">" or ";":
+    for char in value:
+        if char in RESERVED_TAG_CHARS:
+            raise serializers.ValidationError('Tag values cannot contain "\t" or ">" or ";" characters.')
+    return value
+
+
 class TagDataSerializer(UserPermissionsSerializerMixin, serializers.Serializer):  # pylint: disable=abstract-method
     """
     Serializer for TagData dicts. Also can serialize Tag instances.
@@ -236,6 +258,12 @@ class TagDataSerializer(UserPermissionsSerializerMixin, serializers.Serializer):
     sub_tags_url = serializers.SerializerMethodField()
     can_change_tag = serializers.SerializerMethodField()
     can_delete_tag = serializers.SerializerMethodField()
+
+    def validate_value(self, value):
+        """
+        Runs validations for the tag value.
+        """
+        return validate_tag_value(value, self.context)
 
     def get_sub_tags_url(self, obj: TagData | Tag):
         """
@@ -303,6 +331,12 @@ class TaxonomyTagCreateBodySerializer(serializers.Serializer):  # pylint: disabl
     parent_tag_value = serializers.CharField(required=False)
     external_id = serializers.CharField(required=False)
 
+    def validate_tag(self, value):
+        """
+        Run validations for the tag value.
+        """
+        return validate_tag_value(value, self.context)
+
 
 class TaxonomyTagUpdateBodySerializer(serializers.Serializer):  # pylint: disable=abstract-method
     """
@@ -311,6 +345,12 @@ class TaxonomyTagUpdateBodySerializer(serializers.Serializer):  # pylint: disabl
 
     tag = serializers.CharField(required=True)
     updated_tag_value = serializers.CharField(required=True)
+
+    def validate_updated_tag_value(self, value):
+        """
+        Run validations for the updated tag value.
+        """
+        return validate_tag_value(value, self.context)
 
 
 class TaxonomyTagDeleteBodySerializer(serializers.Serializer):  # pylint: disable=abstract-method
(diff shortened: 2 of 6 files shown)


openedx-webhooks:
Thanks for the pull request, @jesperhodge!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

kdmccormick (changes requested):
@jesperhodge I agree that exception details should not be passed to the end user on a production site. However, it's useful behavior for developers. I'd expect the stack trace to be passed upwards when `DEBUG==True`, and a generic 500 to be shown when `DEBUG==False`.

Did you test with `DEBUG==False`? You can do django settings override, or just run the site with `tutor local ...`.

kdmccormick:
To you other point:

> Validate against duplicate values before they reach the database and cause an IntegrityError (this is not implemented here yet)

yep, I agree. I think it's fine to either check for conflicts before renaming the tag, or use a very targeted `except IntegrityError` to catch the conflict.

jesperhodge:
I did have a look at https://github.com/openedx/openedx-platform/pull/38246, an ADR in review that aims to standardize errors platform-wide. (Also reviewed it and left my thoughts.)

I decided against implementing the proposed format more extensively here at this point, since there is still a lot to discuss around that.

Rather, I think it makes sense that after that ADR is finalized we just implement a top-level CMS error handler, and this temporary solution in this PR can be removed at that time.

kdmccormick:
To close the loop on this: 
> Did you test with DEBUG==False? You can do django settings override, or just run the site with tutor local ....

Jesper discussed in Slack that he did test with DEBUG==False, and he and Braden confirmed that this error detail leakage is a standing issue with DRF in general.

> Rather, I think it makes sense that after that ADR is finalized we just implement a top-level CMS error handler, and this temporary solution in this PR can be removed at that time.

That approach sounds good to me @jesperhodge . Could you just include a link to the ADR PR in the temporary handlers so that we remember to remove them later?

src/openedx_tagging/rest_api/v1/exception_handlers.py
```diff
@@ -0,0 +1,57 @@
...
+from django.http import Http404
+from rest_framework import status
+from rest_framework.exceptions import APIException, PermissionDenied
+from rest_framework.response import Response
+from rest_framework.views import exception_handler
+
+log = logging.getLogger(__name__)
+
+
+def custom_exception_handler(exc, context):
```
kdmccormick:
```suggestion
def _custom_exception_handler(exc: Exception, context: dict) -> Response:
```
Please use type annotations, and prefix with an underscore unless this is intended for use elsewhere.

src/openedx_tagging/rest_api/v1/exception_handlers.py
```diff
@@ -0,0 +1,57 @@
...
+    # For exceptions expected by DRF return the standard DRF error response:
+    # Instances of APIException, subclasses of APIException, Django's Http404 exception,
+    # and Django's PermissionDenied exception.
+    is_expected_exception = isinstance(
+        exc, (APIException, Http404, PermissionDenied)
+    )
+    if is_expected_exception:
+        return exception_handler(exc, context)
+
+    # if django settings have DEBUG=True
```
kdmccormick:
```suggestion
```
No need to state the obvious :)

src/openedx_tagging/rest_api/v1/exception_handlers.py
```diff
@@ -0,0 +1,57 @@
...
+    if settings.DEBUG:
+        log.exception(exc)
```
kdmccormick: Logging unexpected exceptions is just as important, if not more important, on a prod instance, so I'd move this outside the `if`.

kdmccormick: [`log.exception`](https://docs.python.org/3/library/logging.html#logging.Logger.exception) is only to be called in an exception handler (that is, a literal python `except:` block, not a DRF exception handler). The first argument is supposed to be an additional message, not an exception itself.

I think best the approach would be to call `log.error` using the `detail` string you have constructed below.

src/openedx_tagging/rest_api/v1/serializers.py
```diff
@@ -323,6 +363,25 @@ class TaxonomyTagDeleteBodySerializer(serializers.Serializer):  # pylint: disabl
...
+        """
+        Make sure all tags are valid and exist before attempting deletion, to avoid partial deletes.
+        """
+        # Iterate through the list and make one bulk request that checks whether every tag.value exists
+        taxonomy_id = self.context.get("taxonomy_id")
+        existing_tags = set(
+            Tag.objects.filter(taxonomy_id=taxonomy_id, value__in=tags_list)
+            .values_list("value", flat=True)
+        )
+        missing_tags = [tag for tag in tags_list if tag not in existing_tags]
```
kdmccormick:
```suggestion
        missing_tags = set(tags_list) - existing_tags
```
IMO, this is easier to read when written using set arithmetic

tests/openedx_tagging/test_views.py
```diff
@@ -2173,6 +2175,153 @@ def test_update_tag_in_taxonomy_with_different_methods(self):
...
+            "updated_tag_value": updated_tag_value
+        }
+
+        response = self.client.put(
+            self.small_taxonomy_url, update_data, format="json"
+        )
+
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+        # Check that the error message indicates the duplicate value issue
+        self.assertIn("Tag value \"Updated Tag\" already exists in this taxonomy", str(response.data))
```
kdmccormick:
```suggestion
        assert "Tag value \"Updated Tag\" already exists in this taxonomy" in str(response.data)
```
It's preferable to use the `assert` keyword over `self.assert*` whenever available. ruff will begin enforcing this soon.

tests/openedx_tagging/test_views.py
```diff
@@ -2173,6 +2175,153 @@ def test_update_tag_in_taxonomy_with_different_methods(self):
...
+            assert_generic_500_response(response)
+
+    def test_passes_stack_trace_upwards_when_settings_debug_true(self):
+        """
+        Test that when an unexpected error occurs, if we're in debug mode, the stack trace is included in the response
+        to help with debugging, instead of just a generic error message.
+        """
+        self.client.force_authenticate(user=self.staff)
+        # simulate debug mode by patching the settings.DEBUG value to True
+        with patch("django.conf.settings.DEBUG", True):
```
kdmccormick:
```suggestion
        with override_settings(DEBUG=True)
```
standard practice is to use override_settings for this https://docs.djangoproject.com/en/6.0/topics/testing/tools/#django.test.override_settings

tests/openedx_tagging/test_views.py
```diff
@@ -2173,6 +2175,153 @@ def test_update_tag_in_taxonomy_with_different_methods(self):
...
+
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+        # Check that the error message indicates the duplicate value issue
+        self.assertIn("Tag value \"Updated Tag\" already exists in this taxonomy", str(response.data))
+
+    def test_should_handle_unexpected_errors_gracefully(self):
+        """
+        Test that if any unexpected error occurs during the processing of the request,
+        the API handles it gracefully and returns a 500 error with a helpful error message,
+        instead of crashing or exposing any sensitive information.
```
kdmccormick:
```suggestion
        the API converts it to a generic 500 response without exposing sensitive info.
```

Sorry to nitpick your wording, just want to highlight a few things about the framing of this:

* There's nothing wrong with crashing when there's an unexpected error--it's basically what we're doing, anyway. I would characterize a 5xx response as a presentation of a crash.
* There error message is not helpful, but that's on purpose.

tests/openedx_tagging/test_views.py
```diff
@@ -2173,6 +2175,153 @@ def test_update_tag_in_taxonomy_with_different_methods(self):
...
+        def get_request(params):
+            return self.client.get(
+                self.small_taxonomy_url, params, format="json"
+            )
+
+        def put_request(update_data):
+            return self.client.put(
+                self.small_taxonomy_url, update_data, format="json"
+            )
+
```
kdmccormick: Nit, non-blocking: This test is very comprehensive, but I find its implementation to be overengineered. In general, I find that our standard DRYness patterns (helper methods, always using constants, etc.) tend to increase brittleness and decrease readability when applied to test code. 

Consider the fully-inlined version of this test method, which would be easily skimmable and come out to a similar number of lines or fewer.

```python
    EXPECTED_ERROR_MESSAGE = "An unexpected error occurred while processing the request."
    ...
    ...
    response = self.client.patch(self.small_taxonomy_url, update_data, format="json")
    assert response.status == 500
    assert EXPECTED_ERROR_MESSAGE in str(response.data)
```

tests/openedx_tagging/test_views.py
```diff
@@ -4,12 +4,14 @@
 from __future__ import annotations
```
kdmccormick, line 4: Thanks for the comprehensive tests.

kdmccormick (changes requested):
Thanks for thinking this through and coming up with a stop-gap solution.

tests/openedx_tagging/test_views.py
bradenmacdonald: > DRYness patterns (helper methods, always using constants, etc.) tend to increase brittleness and decrease readability when applied to test code.

Big +1 on that, I've really come to appreciate this over time. I'd much rather have tests that are long and verbose but super easy to follow + understand that tests that are concise but difficult to follow. Plus, not using constants and hard-coding strings etc. into your test can actually help you catch more errors and better understand when things change in your codebase.

src/openedx_tagging/rest_api/v1/exception_handlers.py
```diff
@@ -0,0 +1,57 @@
...
+    # Instances of APIException, subclasses of APIException, Django's Http404 exception,
+    # and Django's PermissionDenied exception.
+    is_expected_exception = isinstance(
+        exc, (APIException, Http404, PermissionDenied)
+    )
+    if is_expected_exception:
+        return exception_handler(exc, context)
+
+    # Making really sure that any unexpected exception gets logged
+    log.error(exc, stack_info=True, exc_info=True)
```
kdmccormick:
```suggestion
    # DRF always calls exception handlers from within an `except:` block, so we can assume
    # that `log.exception` will automatically insert those exception details and a stack trace.
    log.exception("Unexpected exception while handling API request")
```

Sorry, I led you astray last time when I said that we needed to use `log.error` instead of `log.exception`. My rationale was that the `log.exception` method should only be used within an `except:` block, but it turns out that we **_are_** in a [caller's except block ](https://github.com/encode/django-rest-framework/blob/385df20e3efd442107fd952fbb1d681bed3905b9/rest_framework/views.py#L514-L515)whenever we're in this method (you might have realized this when you wrote the code originally--I've suggested a comment to make it clear for future devs).

Along the same lines, there's no need to redundantly pass `exc` to the logger. You can just pass a generic message, which will be added above the details of `exc`.

kdmccormick (commented):
Just one last request, and then we're good to merge. Thanks!

kdmccormick:
Squashing as:
```
fix(tagging): Handle rename conflicts; Better 500 handling

This fixes two issues: 
1. Renaming a tag such that it conflicts with an existing tag should result
   in a 4xx and a helpful error message. 
2. Unexpected errors while handling requests should result in a simple
   500 error response without internal details (unless in debug mode)
   and a log message with exception details. This is default Django behavior.
   However DRF returns internal exception details, which is confusing to
   users and bad security practice.  This PR restores the default Django behavior
   for the Tagging API in particular. In the future, we'd like to generalize this handler
   to work for all our DRF APIs.

Fixes: https://github.com/openedx/modular-learning/issues/261
```

kdmccormick (approved):

src/openedx_tagging/rest_api/v1/serializers.py
```diff
@@ -217,6 +218,27 @@ class ObjectTagUpdateBodySerializer(serializers.Serializer):  # pylint: disable=
...
+    Validate this tag value is unique within the current taxonomy context and
+    does not contain forbidden characters.
+    """
+    taxonomy_id = context.get("taxonomy_id")
+    if taxonomy_id is not None:
+        # Check if tag value already exists within this taxonomy. If so, raise a validation error.
+        queryset = Tag.objects.filter(taxonomy_id=taxonomy_id, value=value)
+        if queryset.exists():
+            raise serializers.ValidationError(
+                f'Tag value "{value}" already exists in this taxonomy.', code='unique'
```
kdmccormick, line 232: Sorry @jesperhodge , I only just thought of this after merging:

Do these validation strings appear in the end user UI? If so, they need to be internationalized.

jesperhodge: Can you give me an example of how we do that?
As far as I see, we are not internationalizing a single error response in the whole repository.

I think since this is ubiquitous that would call for a new issue that has a bit wider scope and would include internationalizing other error messages like [this](https://github.com/openedx/openedx-core/blob/5391abc5233b4dfd8c675e25dc8400c468285c2f/src/openedx_tagging/models/base.py#L200). What do you think?

Is there any prior art in edx-platform?

jesperhodge: Hmm I had a quick look at edx-platform.

It seems like that in edx-platform, ValidationError messages forwarded to the frontend on edx-platform are also not internationalized, for example https://github.com/openedx/openedx-platform/blob/77293cdf8e97eddcb3db32624b690dff8fe5d0df/lms/djangoapps/badges/models.py#L15

kdmccormick: @jesperhodge 

> As far as I see, we are not internationalizing a single error response in the whole repository.

Ack, you're right, and it's worse than that--nothing in this repo is internationalized. And I guess we've noticed that before, too: https://github.com/openedx/openedx-core/issues/483 . No need for you to do anything here--we'll do a big sweep of internationalization at some point in the future. Thanks for checking on that.

> would include internationalizing other error messages like [this](https://github.com/openedx/openedx-core/blob/5391abc5233b4dfd8c675e25dc8400c468285c2f/src/openedx_tagging/models/base.py#L200). What do you think?

I think the rule of thumb is "If the user will see it, it must be internationalized" rather than a blanket "all exceptions strings do/don't need to be internationalized".

> Is there any prior art in edx-platform?

Yes, we use the standard [gettext](https://docs.djangoproject.com/en/6.0/topics/i18n/translation/) function for all translation. Usually we alias it to `_`:

```python
from django.utils.translation import gettext as _

my_message = _("translate me")
my_template = _("hello {name}, translate me")
````

kdmccormick: > It seems like that in edx-platform, ValidationError messages forwarded to the frontend on edx-platform are also not internationalized, for example https://github.com/openedx/openedx-platform/blob/77293cdf8e97eddcb3db32624b690dff8fe5d0df/lms/djangoapps/badges/models.py#L15

That message is internationalized. The `_` function returns the localized string, if it available.

jesperhodge: > That message is internationalized. The _ function returns the localized string, if it available.

Oh! That's good

PR #539, opened 2026-04-13
feat: add `direct` boolean field to PublishLogRecord [FC-0123]

ChrisChV:
Github issue: https://github.com/openedx/openedx-core/issues/533

Track whether each publish record was explicitly requested by the user (direct=True) or was a side-effect of publishing a parent/dependency (direct=False). Historical records get direct=None since intent cannot be determined retroactively.

* Add `direct = BooleanField(null=True, blank=True, default=False)` to `PublishLogRecord`.
* In `publish_from_drafts()`, `collect direct_draft_ids` before the loop and set `direct=True/False` per record accordingly.


openedx-webhooks:
Thanks for the pull request, @ChrisChV!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

ChrisChV:
Hi @ormsbee, could you review this PR? This is the new `direct` field

ormsbee:
Sure thing. I should be able to look this over in the morning.

src/openedx_content/applets/publishing/models/publish_log.py
```diff
@@ -111,6 +111,18 @@ class PublishLogRecord(models.Model):
     # the values may drift away from each other.
     dependencies_hash_digest = hash_field(blank=True, default='', max_length=8)
 
+    # True if this entity was explicitly requested to be published by the user
+    # (e.g. they clicked "publish" on this entity or selected it for bulk publish).
+    # False if it was indirectly published as a child/dependency of a directly
+    # published entity (e.g. a Component published because its parent Unit was
+    # published).
+    # None for historical records created before this field was added.
```
ormsbee:
```suggestion
    # The "direct" field captures user intent during the publishing process. It
    # is True if the user explicitly requested to publish the entity represented
    # by this PublishLogRecord—i.e. they clicked "publish" on this entity or
    # selected it for bulk publish.
    # 
    # This field is False if this entity was indirectly published either as a
    # child/dependency or side-effect of a directly published entity.
    #
    # If this field is None, that means that this PublishLogRecord was created
    # before we started capturing user intent (pre-Verawood release), and we
    # cannot reliably infer what the user clicked on. For example, say we had a
    # Subsection > Unit > Component arrangement where the Component had an
    # unpublished change. The user is allowed to press the "publish" button at
    # the Subsection, Unit, or Component levels in the UI. Before we started
    # recording this field, the resulting PublishLogs would have looked
    # identical in all three cases: a version change PublishLogRecord for
    # Component, and side-effect records for the Unit and Subsection Therefore,
    # we cannot accurately backfill this field.
    # 
    # Here are some examples to illustrate how "direct" gets set and why:
    #
    # Example 1: The user clicks "publish" on a Component that's in a Unit.
    #
    #   The Component has direct=True, but the side-effect PublishLogRecord for
    #   the Unit has direct=False. Likewise, any side-effect records at higher
    #   levels (subsection, section) also have direct=False.
    # 
    # Example 2: The user clicks "publish" on a Unit, where both the Unit and
    # Component have unpublished changes:
    # 
    #   In this case, the Unit has direct=True, and the Component has
    #   direct=False. The draft status of the Component is irrelevant. The user
    #   asked for the Unit to the published, so the Unit's PublishLogRecord is
    #   the only thing that gets direct=True.
    #
    # Example 3: The user clicks "publish" on a Unit that has no changes of its
    # own (draft version == published version), but the Unit contains a Component
    # that has has changes.
    # 
    #   Again, only the PublishLogRecord for the Unit has direct=True. The
    #   Component's PublishLogRecord has direct=False. Even though the Unit's
    #   published version_num does not change (i.e. it is purely a side-effect
    #   publish), the user intent was to publish the Unit (and anything it
    #   contains), so the Unit gets direct=True.
    #
    # Example 4: The user selects multiple entities for bulk publishing.
    # 
    #   Those exact entities that the user selected get direct=True. It does not
    #   matter if some of those entities are children of other selected items or
    #   not. Other entries like dependencies or side-effects have direct=False.
    # 
    # Example 5: The user selects "publish all".
    # 
    #   Selecting "publish all" in our system currently translates into "publish
    #   all the entities that have a draft version that is different from its
    #   published version". Those entities would get PublishLogRecords with
    #   direct=True, while all side-effects would get records with direct=False.
    #   So if a Unit's draft and published versions match, and one of its
    #   Components has unpublished changes, then "publish all" would cause the
    #   Component's record to have direct=True and the Unit's record to have
    #   direct=False.
    #    
    #   All PublishLogRecords in the PublishLog have direct=True. The "publish
    #   all" operation is indistinguishable from bulk publishing and selecting
    #   every single item.
```

src/openedx_content/applets/publishing/models/publish_log.py
```diff
@@ -111,6 +111,18 @@ class PublishLogRecord(models.Model):
...
+    direct = models.BooleanField(
+        null=True,
+        blank=True,
+        default=None,
```
ormsbee: It would be nice if we could actually backfill this with `None` but default new values to `False`, since that would mean we wouldn't have to touch `_create_side_effects_for_change_log()`, which is one of the more complex functions already.

tests/openedx_content/applets/publishing/test_api.py
```diff
@@ -1429,6 +1458,100 @@ def test_publish_all_layers(self) -> None:
...
+        containers_api.create_container_version(
+            unit.id, 1, title="Direct Unit", entities=[component],
+            created=self.now, created_by=None,
+        )
+        publish_log = publishing_api.publish_from_drafts(
+            self.learning_package.id,
+            Draft.objects.filter(entity=unit.publishable_entity),
+        )
+        assert publish_log.records.get(entity=unit.publishable_entity).direct is True
+        assert publish_log.records.get(entity=component).direct is False
```
ormsbee, line 1509: Please add a bit more to this test to assert that publishing the Unit still marks the Unit's record as `direct=True`, even if there is no version change on the Unit (i.e. only the Component's version is actually changing in the publish).

tests/openedx_content/applets/publishing/test_api.py
```diff
@@ -935,6 +935,35 @@ def test_simple_publish_log(self) -> None:
...
+        entity = publishing_api.create_publishable_entity(
+            self.learning_package_1.id, "direct_entity",
+            created=self.now, created_by=None,
+        )
+        publishing_api.create_publishable_entity_version(
+            entity.id, version_num=1, title="Direct Entity",
+            created=self.now, created_by=None,
+        )
+        publish_log = publishing_api.publish_all_drafts(self.learning_package_1.id)
+        assert publish_log.records.get(entity=entity).direct is True
```
ormsbee: Shouldn't this test have multiple entities?

ormsbee (changes requested):
Some minor requests and a really verbose suggested comment to try to avoid later confusion about the data model. Please let me know if anything in there does not align with your understanding of this feature.

src/openedx_content/applets/publishing/models/publish_log.py
ChrisChV: Updated: https://github.com/openedx/openedx-core/pull/539/commits/bb9698b365b2be767ae9ce2d9189d3cdb6b41f37

src/openedx_content/applets/publishing/models/publish_log.py
ChrisChV: Right. https://github.com/openedx/openedx-core/pull/539/commits/bb9698b365b2be767ae9ce2d9189d3cdb6b41f37

tests/openedx_content/applets/publishing/test_api.py
ChrisChV: Updated: https://github.com/openedx/openedx-core/pull/539/commits/bb9698b365b2be767ae9ce2d9189d3cdb6b41f37

tests/openedx_content/applets/publishing/test_api.py
ChrisChV: Updated: https://github.com/openedx/openedx-core/pull/539/commits/bb9698b365b2be767ae9ce2d9189d3cdb6b41f37

ormsbee:
@ChrisChV: I forgot when reviewing yesterday, but could you please add this field to the Django admin view for PublishLogs (there's an inline view of PublishLogRecords that already has all the other fields, so it should be easy to add).

ormsbee (changes requested):
Please add the new field to the Django admin view for Components.

ChrisChV:
>  add this field to the Django admin view for PublishLogs

Done :+1: Sorry for the late update here.

> Please add the new field to the Django admin view for Components.

Are you referring to PublishLogs? I'm not sure where to put it in the Components.

ormsbee:
> Are you referring to PublishLogs? I'm not sure where to put it in the Components.

Yeah, sorry, it was just a repeat of my last request, but I'm getting my wires crossed looking at different PRs. Anyhow, yes, please add this new field to the inline display of PublishLogRecords in the PublishLog admin.

ormsbee:
... which I see you've already done. :smile:

ormsbee (approved):

PR #542, opened 2026-04-15
feat!: Collection.key -> Collection.collection_code

kdmccormick:
## Description

Also, standardize internal usage of collection_key to collection_code.
This helps clarify that Collection.key is *not* an OpaqueKey, but is rather
a local slug, which can be combined with other identifiers to form a fully-
qualified LibraryCollectionKey instance.

BREAKING CHANGE: Collection.key has been renamed to Collection.collection_code.

BREAKING CHANGE: Collection.collection_code now validates that its contents
matches '[A-Za-z0-9\-\_\.]+'.  This was already effectively true, because
LibraryCollectionKey can only be built with slug-like parts, but we now
we explicitly raise ValiationError from create_collection.

Backup-restore still write and reads the collection_code to/from TOML files as `key` for backwards compatibility. This may change in a future "v2" restore format.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

Part of:
* https://github.com/openedx/openedx-core/issues/322

Full series of PRs:
1. #542
2. #544
3. #545
4. #546
5. #547


src/openedx_content/applets/components/models.py
```diff
@@ -1,5 +1,5 @@
 """
-The model hierarchy is Component -> ComponentVersion -> Media.
+The model hierarchy is Component -> ComponentVersion -> Content.
```
ormsbee: It looks like this undoes my doc changes in https://github.com/openedx/openedx-core/pull/518

src/openedx_content/applets/units/models.py
```diff
@@ -1,6 +1,7 @@
 """
 Models that implement units
 """
+from __future__ import annotations
```
ormsbee, line 4: Question: Is this just because we want to standardize on importing annotations?

src/openedx_content/migrations/0008_rename_collection_key_to_code_in_db.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('openedx_content', '0007_rename_collection_key_to_code_in_python'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='collection',
+            name='collection_code',
+            field=openedx_django_lib.fields.MultiCollationCharField(db_collations={'mysql': 'utf8mb4_bin', 'sqlite': 'BINARY'}, max_length=500),
```
ormsbee: Ack. I take it this is required to get Django to rename the underlying `_key` column? Is there a reason why this operation isn't just tacked onto the previous migration?

src/openedx_content/applets/components/models.py
kdmccormick: Oops, rebase fail. I'll put those back.

src/openedx_content/applets/backup_restore/toml.py
```diff
@@ -220,7 +220,10 @@ def toml_collection(collection: Collection, entity_keys: list[str]) -> str:
 
     collection_table = tomlkit.table()
     collection_table.add("title", collection.title)
-    collection_table.add("key", collection.key)
+    # Write both names so that older software (which reads 'key') stays compatible
+    # with archives produced after the Collection.key -> Collection.collection_code rename.
+    collection_table.add("key", collection.collection_code)
+    collection_table.add("collection_code", collection.collection_code)
```
ormsbee: I think that we should just continue to serialize it to `key` for now (i.e. keep the backup archive format completely unchanged), and potentially change the names in a v2 of the serialization format that can handle import/export. The fact that we have to serialize both ways anyway for backwards compatibility means that we're not really making things any clearer to people reading the file format--and may actually be making things more confusing.

ormsbee (changes requested):

src/openedx_content/migrations/0008_rename_collection_key_to_code_in_db.py
kdmccormick: It's what it took to get Django to recognize it as a rename rather than a field-removal and a field-addition. But I think I could squish it into one now.

src/openedx_content/applets/units/models.py
kdmccormick: I think I added this while fighting with mypy, not sure if it ended up being needed. It's harmless either way, but I could remove it.

ormsbee: I'm fine with it. I just added some for the pydantic modeling stuff because I like being able to specify structures top-down in the file. I was mostly just trying to understand why it was there, since it was the only change to the file.

src/openedx_django_lib/fields.py
```diff
@@ -114,6 +117,29 @@ def immutable_uuid_field() -> models.UUIDField:
...
+    return case_sensitive_char_field(
+        max_length=255,
+        blank=False,
+        validators=[
+            RegexValidator(
+                CODE_REGEX,
+                # Translators: "letters" means latin letters: a-z and A-Z.
+                _('Enter a valid "code name" consisting of letters, numbers, underscores, hyphens, or periods.'),
+                "invalid",
+            ),
```
bradenmacdonald, line 146: (copied from other PR): Django now supports doing regex validations via database constraints, like: 

https://github.com/openedx/openedx-core/blob/5391abc5233b4dfd8c675e25dc8400c468285c2f/src/openedx_catalog/models/catalog_course.py#L238-L239

This has the advantage that it's enforced at the database level, with perfect consistency, which is much better because Django validators are not normally called on `.update()` or `.save()`. But if you define a check constraint, not only does it run at the database level, but Django _also_ runs it as a python-level validator automatically, so you get all the same advantages of a validator.

The only thing is, I'm not sure offhand if there's any way to specify constraints like this at the field level, only at the model level 😞. So I would suggest implementing a helper function alongside `code_field` like `code_field_check(field_name)` that can be used to add the DB constraints too.

src/openedx_django_lib/fields.py
```diff
@@ -114,6 +117,29 @@ def immutable_uuid_field() -> models.UUIDField:
     )
 
 
+# Alphanumeric, hyphens, underscores, periods
+CODE_REGEX = re.compile(r"^[a-zA-Z0-9\-\_\.]+\Z")
```
ormsbee:
```suggestion
CODE_REGEX = re.compile(r"^[a-zA-Z0-9_.-]+\Z")
```

src/openedx_django_lib/fields.py
kdmccormick: good idea, done

src/openedx_content/applets/backup_restore/toml.py
kdmccormick: makes sense. done.

src/openedx_content/migrations/0008_rename_collection_key_to_code_in_db.py
kdmccormick: squashed, and same done for the other PRs

src/openedx_content/applets/components/models.py
kdmccormick: done

ormsbee (approved):
Great stuff! Thank you!

src/openedx_content/applets/collections/api.py
```diff
@@ -44,35 +44,37 @@ def create_collection(
...
     Get a Collection by ID
     """
-    return Collection.objects.get_by_key(learning_package_id, collection_key)
+    return Collection.objects.get_by_code(learning_package_id, collection_code)
 
 
 def update_collection(
     learning_package_id: LearningPackage.ID,
-    key: str,
+    collection_code: str,
```
bradenmacdonald, line 69: Not at all a blocker, but something to consider for the future: our APIs are a bit inconsistent about how we identify the "thing" we're modifying in each API call.

My preference is this style where you can pass in an instance or a primary key:

https://github.com/openedx/openedx-core/blob/10eae2da8d5e8cfc2d5b9a451687eaa9b4f40530/src/openedx_catalog/api_impl.py#L63-L64

https://github.com/openedx/openedx-core/blob/10eae2da8d5e8cfc2d5b9a451687eaa9b4f40530/src/openedx_content/applets/containers/api.py#L454-L455

But this collection API is now using a composite (`lp_id`, `collection_code`) pattern:

https://github.com/openedx/openedx-core/blob/0614fcb80d7d4896d507c03fd91e063cb3ebe760/src/openedx_content/applets/collections/api.py#L67-L69

And other APIs accept only an ID:

https://github.com/openedx/openedx-core/blob/10eae2da8d5e8cfc2d5b9a451687eaa9b4f40530/src/openedx_content/applets/components/api.py#L155-L156

And other APIs accept only an instance:

https://github.com/openedx/openedx-core/blob/10eae2da8d5e8cfc2d5b9a451687eaa9b4f40530/src/openedx_tagging/api.py#L436-L437

src/openedx_content/migrations/0008_rename_collection_key_to_collection_code.py
```diff
@@ -0,0 +1,73 @@
...
+from django.db import migrations, models
+
+import openedx_django_lib.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('openedx_content', '0007_publishlogrecord_direct'),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
```
bradenmacdonald, line 18: This might be another Claude artifact. This migration does not actually depend on the `AUTH_USER_MODEL`.

bradenmacdonald (approved):
I didn't test this but the code changes look great. Thanks!

src/openedx_content/applets/collections/api.py
kdmccormick: I agree, this would be great to fix soon

src/openedx_content/migrations/0008_rename_collection_key_to_collection_code.py
kdmccormick: I presume we only need it for migrations which touch an FK to User?

bradenmacdonald: Yes, I looked and every migration in core that has this dependency has a reference to `settings. AUTH_USER_MODEL` elsewhere in the list of changes.

kdmccormick: Noted on https://github.com/openedx/openedx-core/issues/322 for followup

ChrisChV:
Hi @kdmccormick, is the refactoring of this field on the `edx-platform` in progress? I am working on https://github.com/openedx/openedx-core/pull/501, and several errors are appearing in `edx-platform` about this field, preventing us from testing the new features.

kdmccormick:
@ChrisChV Yes, it is in progress. I should have an openedx-platform PR open later today which you could base your changes on.

bradenmacdonald:
@ChrisChV Until that other PR is done, you need to use a slightly old version of `openedx-core`, or apply a temporary revert commit that undoes this PR.

ChrisChV:
@bradenmacdonald Sure, I'll use a revert for that. Thanks!

PR #543, opened 2026-04-15
Emit events when a learning package is modified [FC-0117]

bradenmacdonald:
Implements https://github.com/openedx/openedx-core/issues/462 .

This updates openedx-code to emit the following events:

**1. `LEARNING_PACKAGE_ENTITIES_CHANGED`**

https://github.com/openedx/openedx-core/blob/7ad1e799a495e927662c04a172e743920b6ff6f1/src/openedx_content/applets/publishing/signals.py#L90-L95

**2. `LEARNING_PACKAGE_ENTITIES_PUBLISHED`**

https://github.com/openedx/openedx-core/blob/7ad1e799a495e927662c04a172e743920b6ff6f1/src/openedx_content/applets/publishing/signals.py#L125-L129

**3. `LEARNING_PACKAGE_COLLECTION_CHANGED`**

https://github.com/openedx/openedx-core/blob/7ad1e799a495e927662c04a172e743920b6ff6f1/src/openedx_content/applets/collections/signals.py#L48-L49

4. `LEARNING_PACKAGE_CREATED` / `LEARNING_PACKAGE_UPDATED` / `LEARNING_PACKAGE_DELETED`. Not sure we have much use case for these yet, but they seem like important low-level hooks.

### TODOs

Test that the performance of event handlers responding to `set_collections` is OK in practice and change it to async if needed.

### Notes

Tagging events are much more complicated so will be in their own PR. The proposed spec is at https://github.com/openedx/openedx-core/issues/557

While working on this, I also opened https://github.com/openedx/openedx-events/pull/570 which would make working with these kind of events easier in the future, and let us slightly simplify this code in a few places.

These events are defined locally in `openedx_content` for now but we could move them to openedx-events in the future?

### Platform PR

https://github.com/openedx/openedx-platform/pull/38437

The platform PR is approx -500 lines of non-test code, and +265 lines of test code, and I think it's much cleaner, so I'm really happy with the downstream effects of this PR. Part of this is being able to take advantage of the "dependencies" / "side effect" tracking properly, which the original library events weren't able to do at the time.


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

src/openedx_content/applets/publishing/signals.py
```diff
@@ -0,0 +1,82 @@
...
+
+@define
+class ChangeLogRecord:
+    """A single change that was made to a PublishableEntity"""
+
+    entity_id: PublishableEntity.ID
+
+    old_version: int | None
+    """
+    The old version_num of this entity (not the PublishableEntityVersion ID).
```
bradenmacdonald, line 46: Working on this PR has made me want to add fully-typed IDs for `PublishableEntityVersion` and some kind of `VersionNum` type, because we do mix and match those a bit in the code, but they're very different things.

(I made `PublishableEntity.id` fully-typed but hadn't yet done `___Version`)

tests/openedx_content/applets/publishing/test_signals.py
```diff
@@ -0,0 +1,72 @@
...
+    event = captured[0]  # capture_events(...) context manager already asserted there's only one event.
+    assert event.signal is api.signals.LEARNING_PACKAGE_ENTITIES_CHANGED
+    assert event.kwargs["learning_package"].id == learning_package.id
+    assert event.kwargs["learning_package"].title == "Test LP 📦"
+    assert event.kwargs["changed_by"].user_id is None
+    assert event.kwargs["change_log"].draft_change_log_id == expected_draft_change_log_id
+    assert event.kwargs["change_log"].changes == [
+        api.signals.ChangeLogRecord(entity_id=entity.id, old_version=None, new_version=NEW_VERSION_NUM),
+    ]
+    assert event.kwargs["metadata"].time == now_time
```
bradenmacdonald, line 60: Unfortunately, openedx-events [doesn't yet support typing](https://github.com/openedx/openedx-events/issues/563) so `event.kwargs` is just a generic `dict`.

src/openedx_content/applets/publishing/signals.py
```diff
@@ -0,0 +1,82 @@
...
+    """Identifies which user triggered the event."""
+
+    user_id: int | None
+
+
+@define
+class ChangeLogRecordData:
+    """A single change that was made to a PublishableEntity"""
+
+    entity_id: PublishableEntity.ID
```
bradenmacdonald, line 42: Should I include `container_type`, `component_type`, `container_code`, and `component_code` in these ChangeLogRecords? I think so, but I currently implemented this entirely within the `publishing` app, and doing so would require either making `publishing` aware of containers/components or moving the signals out of publishing to become generic `openedx_content` signals. Perhaps I could even make `signals` its own applet?

bradenmacdonald: @ormsbee @kdmccormick Any opinions on this question ^ ?

Edit: I think I have something in mind that'll work, so no worries / no rush here.

kdmccormick: @bradenmacdonald sorry for the slow response. My hunch is to make a `signals` applet, given that this might not be the only instance where we want to respond with Data objects with cross-applet information. Kinda like how `backup_restore` works with data from across all the lower applets. Curious what you come up with, though.

bradenmacdonald: I do also like the `signals` or `events` applet approach, but the challenge is that there aren't good hooks into `publishing`/`collections` to use, so to support that still requires `publishing`/`collections`/`tagging` to emit internal signals, then the new `events` applet to receive them + enrich them and re-emit them as public signals, but then we're still missing the context info like library key, so platform has to receive and re-emit them yet again. I'm playing around with this on the platform side, but I'm thinking for now it's simpler to just have one set of low-level events emitted from Learning Core and have them re-emitted in the platform, which we can later change to be re-emitted from Learning Core itself.

bradenmacdonald: This is actually a perfect example of why having a single consolidated `EntityType` in the `publishing` app would be excellent, as opposed to `ComponentType`/`ContainerType` that `publishing` isn't aware of.

Before @kdmccormick's changes, the `PublishableEntity.key` _could_ also have been included in events and used to basically provide the container/component type (e.g. `xblock.v1:problem:my-key` vs. `unit:my-unit`) albeit in a way that requires parsing and is not ideal - but is efficient in terms of avoiding JOINs/queries.

ormsbee: An additional thing I'll note is that not all LearningPackages will have a library key. Leaving aside a future with courses, we currently create a detached LearningPackage when a restore happens, and then hook it up to a library after the fact.

Maybe we could do some kind of dependency injection thing where `publishing` is about to fire off a draft-change or publish event, and the different apps are given an opportunity to add their own `select_related` bits to the queryset?

bradenmacdonald: Yeah, I was thinking that too. But (a) even using dependency injection, it creates a bit of a messy circular data flow between publishing and containers/components, and (b) the event types are defined statically, so even if we can inject extra select_related into the queryset, the event data doesn't define additional fields unless we say "there's a random dict of extra data on each event" or move the events to a higher level.

What I'm leaning at the moment:
1) Just have `publishing` emit a single batch event for most changes, and platform can listen to that, annotate/enrich it with context+container+component info, and emit high-level non-batch events for each modified entity (in a celery task). An additional query here with various new JOINs is not going to hurt performance much, as it's still a big batch event and it can be done async when N_entities_modified > 1.
2) Later, we can look at moving "Learning Context" / library / course into openedx_core, and possibly consolidating ComponentType+ContainerType into EntityType. Then openedx_core can emit the same events directly. But that would be a future optimization and not necessary.

Although this isn't "optimal" in terms of reducing the number of events, I do think it has the cleanest separation of concerns given our current architecture. The publishing app just focuses on entities and learning packages which are its main primitives, and higher-level code does higher-level events.

kdmccormick: @bradenmacdonald yup, I agree that (1) the best approach for now given the constraints we have, and everything in (2) is worth considering for Willow+

src/openedx_content/applets/collections/api.py
```diff
@@ -32,6 +35,42 @@
...
+    collection: Collection,
+    *,
+    created: bool = False,
+    metadata_modified: bool = False,
+    deleted: bool = False,
+    entities_added: list[PublishableEntity.ID] | None = None,
+    entities_removed: list[PublishableEntity.ID] | None = None,
+    user_id: int | None = None,
+) -> None:
+    """Helper for emitting the event when a collection has changed."""
```
bradenmacdonald, line 49: This helper would not be necessary if https://github.com/openedx/openedx-events/pull/570 is accepted, and then we could just use that.

src/openedx_content/applets/collections/signals.py
```diff
@@ -0,0 +1,73 @@
...
+    entities_added: list[PublishableEntity.ID] = field(factory=list)
+    entities_removed: list[PublishableEntity.ID] = field(factory=list)
+
+
+LEARNING_PACKAGE_COLLECTION_CHANGED = OpenEdxPublicSignal(
+    event_type="org.openedx.content.collections.lp_collection_changed.v1",
+    data={
+        "learning_package": LearningPackageEventData,
+        "changed_by": UserAttributionEventData,
+        "change": CollectionChangeData,
```
bradenmacdonald, line 44: I am using multiple `x: XData` values in the `data` dict per event. I think this is a much cleaner and DRYer pattern than forcing all events to have a single key-value pair in their data dictionaries, but most (all?) event definitions in openedx-events have only a single key-value pair in the dict even when it doesn't make sense, and [I don't know why](https://github.com/openedx/openedx-events/issues/562).

Edit: based on the additional info I posted there, it should be completely fine. But any automatic parsing of the signal documentation may have incomplete type info (or rather, slightly more incomplete than it already is, since it already lacks the dict key info for all events).

src/openedx_content/applets/publishing/signal_handlers.py
```diff
@@ -0,0 +1,42 @@
...
+
+@receiver(post_delete, sender=LearningPackage)
+def _emit_learning_package_deleted(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """
+    Emit ``LEARNING_PACKAGE_DELETED`` after a ``LearningPackage`` is deleted.
+
+    This fires for any deletion: single-object ``.delete()``, bulk
+    ``QuerySet.delete()`` (Django calls ``post_delete`` once per row), or
+    deletions performed via the Django admin. There is currently no official API
+    for deleting Learning Packages, but you can orhpan them by deleting any
```
ormsbee:
```suggestion
    for deleting Learning Packages, but you can orphan them by deleting any
```
(typo)

tests/openedx_content/conftest.py
```diff
@@ -0,0 +1,15 @@
+"""Shared fixtures for collection tests."""
```
bradenmacdonald:
```suggestion
"""Shared fixtures for openedx_content tests."""
```

src/openedx_content/applets/publishing/signal_handlers.py
```diff
@@ -0,0 +1,42 @@
...
+from django.db import transaction
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
+
+from .models.learning_package import LearningPackage
+from .signals import LEARNING_PACKAGE_DELETED, LearningPackageEventData
+
+
+@receiver(post_delete, sender=LearningPackage)
+def _emit_learning_package_deleted(sender, instance, **kwargs):  # pylint: disable=unused-argument
```
ormsbee: [Question]: Is the underscore prefix here just meant to indicate that people shouldn't import this?

src/openedx_content/applets/publishing/api.py
```diff
@@ -924,6 +952,51 @@ def _create_side_effects_for_change_log(change_log: DraftChangeLog | PublishLog)
     update_dependencies_hash_digests_for_log(change_log)
 
 
+def _emit_event_for_change_log(
+    change_log: PublishLog | DraftChangeLog, time_stamp: datetime, user_id: int | None
```
ormsbee:
```suggestion
    change_log: PublishLog | DraftChangeLog, timestamp: datetime, user_id: int | None
```

Nit: In the few other places in the code where we refer to timestamps, we don't have the underscore.

src/openedx_content/applets/collections/api.py
```diff
@@ -32,6 +35,42 @@
...
+            change=signals.CollectionChangeData(
+                collection_id=collection.id,
+                collection_code=collection.collection_code,
+                created=created,
+                metadata_modified=metadata_modified,
+                deleted=deleted,
+                entities_added=entities_added or [],
+                entities_removed=entities_removed or [],
+            ),
+        )
```
ormsbee: This is elegant, but I think that for this kind of thing, the more immediate binding behavior of a partial will be more intuitive for folks than an inner fn that won't evaluate its args until the on_commit is invoked.

I think this still works for Collection deletion because the fields stick around in the object even after the underlying database entry is removed, but it seems like this is susceptible to subtle bugs if folks are expecting to be able to do more queries here, e.g. if someone wants to do queries to related models in the inner fn.

src/openedx_content/applets/collections/signals.py
```diff
@@ -0,0 +1,73 @@
...
+
+@define
+class CollectionChangeData:
+    """Summary of changes to a collection, for event purposes"""
+
+    collection_id: int
+    collection_code: str
+    created: bool = False
+    """The collection is newly-created, or un-deleted. Some entities may be added simultaneously."""
+    metadata_modified: bool = False
```
ormsbee, line 28: Is this meant to capture title and description changes as well?

ormsbee (changes requested):
Trying to do this in pieces to give a quicker feedback cycle. This is for Collections.

src/openedx_content/applets/publishing/signal_handlers.py
bradenmacdonald: Ah, this function name was generated by Claude; I don't think the _ conveys anything useful so I'll just remove it.

tests/openedx_content/applets/collections/test_signals.py
```diff
@@ -0,0 +1,528 @@
...
+
+    A disabled collection is invisible to consumers, so there is nothing to notify about.
+    """
+    with capture_events(expected_count=0):
+        api.create_collection(
+            lp1.id,
+            collection_code="col1",
+            title="Collection 1",
+            created_by=None,
+            enabled=False,
```
ormsbee, line 73: Nit: So technically, this means that you can get a deletion event for a collection that never emitted a create event, right?

src/openedx_content/applets/collections/signals.py
bradenmacdonald: The description is on the line below, but yeah this `metadata_modified` is meant to _only_ reflect that:

> """The collection's title/description has changed. Does not indicate whether or not entities were added/removed."""

src/openedx_content/applets/publishing/api.py
bradenmacdonald: Changed to `timestamp`

src/openedx_content/applets/collections/api.py
bradenmacdonald: Good point. Better to use `partial` in all these cases, so I've done that now.

src/openedx_content/applets/publishing/signals.py
```diff
@@ -0,0 +1,239 @@
...
+``title`` of the package.
+
+💾 This event is only emitted after the enclosing database transaction has
+been committed. If the transaction is rolled back, no event is emitted.
+
+⏳ This event is emitted synchronously.
+"""
+
+
+LEARNING_PACKAGE_DELETED = OpenEdxPublicSignal(
```
ormsbee, line 138: @bmtcril, @mariajgrimaldi, @robrap: We're declaring an OpenEdxPublicSignal outside of openedx-events here. A couple of questions:

1. Is this going to have any weird consequences w.r.t. the message bus? Should we be making it a simple Django signal for now?
2. @bradenmacdonald  is introducing a Sphinx doc convention of doing docstring-style comments for module variables. Will this cause any problems? Should we also be putting the [normal annotations](https://docs.openedx.org/projects/openedx-events/en/open-release-sumac.master/reference/in-line-code-annotations-for-an-event.html) above the signal? (I'm not even clear if the normal annotations for signals are picked up if they're defined outside of openedx-events.)

src/openedx_content/applets/publishing/signals.py
```diff
@@ -0,0 +1,239 @@
...
+``QuerySet.delete()``), because it is fired by a Django ``post_delete`` signal
+on the ``LearningPackage`` model.
+
+Note: at the time this event is emitted, the ``LearningPackage`` and all of
+its related content (entities, versions, drafts, publishes, etc.) have already
+been removed from the database. Handlers cannot look up the learning package
+by ID — they only get the ``id`` and ``title`` that are captured in the
+``LearningPackageEventData`` payload.
+
+🗑️ Unlike other ``publishing`` events, the effects of this deletion are
```
ormsbee, line 159: I'm guessing that we'll eventually want a pre-delete hook as well, but that's definitely puntable.

src/openedx_content/applets/publishing/signals.py
```diff
@@ -0,0 +1,239 @@
...
+completely irreversible and the LearningPackage cannot be restored/un-deleted.
+
+💾 This event is only emitted after the enclosing database transaction has
+been committed. If the transaction is rolled back, no event is emitted.
+
+⏳ This event is emitted synchronously.
+"""
+
+
+LEARNING_PACKAGE_ENTITIES_CHANGED = OpenEdxPublicSignal(
```
ormsbee: Nit: Is it worth calling out in the name that it's more specifically when the Entities' draft state changes?

src/openedx_content/api.py
```diff
@@ -9,7 +9,10 @@
 """
 
 # These wildcard imports are okay because these api modules declare __all__.
-# pylint: disable=wildcard-import
+# pylint: disable=wildcard-import,unused-import
+
+# Signals are kept in a separate namespace:
+from . import api_signals as signals
```
ormsbee: Nit: Why do this import? The signal receiver import happens in `apps.py` anyway, no?

ormsbee (commented):
Minor comments on the rest of the app code. I don't consider any of these blocking. I'm going to look through the rest of the tests now.

src/openedx_content/api.py
bradenmacdonald: This is for the public API, e.g. in my platform PR:

```python
from openedx_content.api import signals as content_signals
...
@receiver(content_signals.LEARNING_PACKAGE_ENTITIES_CHANGED)
def entities_updated(...):
    ...
```

or even in this PR:

https://github.com/openedx/openedx-core/blob/a53166df0bbe7aacd834fa45d58f60f28fb700ba/tests/openedx_content/applets/publishing/test_signals.py#L46

tests/openedx_content/applets/collections/test_signals.py
bradenmacdonald: Hmm, I should add a test case for that but technically no, having enabled=False should be the same as never being created at all as far as events are concerned, and then actually deleting it should not emit any further events. But enabling it will emit a CREATED event and then re-disabling it ("soft deleting") should emit a deleted event. Let me know if you disagree.

src/openedx_content/api.py
ormsbee: Oh, I see. I thought it would be expected that clients import `api_signals`. But if they're expected to import `api.signals`, it seems like the aggregating thing should be just `signals.py` to match what we do with `models.py` and `admin.py`, and then `api.py` should re-export that?

src/openedx_content/applets/publishing/signals.py
bradenmacdonald: Yeah, you're right - the more explicit the better. Some of the other openedx-events have quite ambiguous names and documentation and that's been a pain.

Do you like `LEARNING_PACKAGE_ENTITIES_DRAFTS_CHANGED`? Or since that's getting rather long we could consider `ENTITIES_DRAFTS_CHANGED`. (Or `_DRAFT_CHANGED` singular?)

src/openedx_content/api.py
bradenmacdonald: I already find it annoying to almost always need a second import of `api_models` in addition to `api` everywhere we use these APIs, so I do prefer including `signals` in the `api` instead of a separate `api_signals`. But if you think there's a good argument for not always loading them and having them in a separate import, let me know.

I definitely don't mind renaming `api_signals` to `signals`; I just wanted to signal that things in it were public, but maybe that's obvious if you look at it.

src/openedx_content/applets/publishing/signals.py
ormsbee: I have a mild preference for `ENTITIES_DRAFT_CHANGED` with a corresponding `ENTITIES_PUBLISHED`.

bradenmacdonald: Or maybe `DRAFT_ENTITIES_CHANGED` and `ENTITIES_PUBLISHED` ? I think that reads better.

src/openedx_content/api.py
ormsbee: Yeah, I think it's fine to:

1. Do the aggregation in `signals.py` and then;
2. Import `signals.py` into `api.py` with the expectation that all external clients will import from `api.signals`

I guess that if we ever do have internal signals for some reason, we can tweak things with an intermediate module later before we re-export.

ormsbee (approved):
No blockers. I would prefer the signal name be changed and the signal aggregation be done in `signals.py` instead of `api_signals.py` before being re-exported, but I have no additional concerns. Thank you.

src/openedx_content/applets/publishing/signals.py
ormsbee: I'd prefer `ENTITIES_DRAFT_CHANGED` to keep it closer to DraftChangeLog and PublishLog.

src/openedx_content/api.py
bradenmacdonald: Updated `api_signals` -> `signals` with a lot more explanation in e42144bd7aa13b18bf9801456c6200a272cf2289

src/openedx_content/applets/publishing/signals.py
bradenmacdonald: Signals renamed in 26683ff3286fd28a52be0592d55d4ba962a68ce6:

* `ENTITIES_DRAFT_CHANGED`
* `ENTITIES_PUBLISHED`
* `COLLECTION_CHANGED`

tests/openedx_content/applets/collections/test_signals.py
bradenmacdonald: Added a fix+test for that edge case: 5ff0496d0e75742e285bc28e3ebe741a2918762f

kdmccormick:
@bradenmacdonald I won't review this unless you'd like me to.

PR #544, opened 2026-04-15
feat!: Component.local_key -> Component.component_code

kdmccormick:
## Description

BREAKING CHANGE: Component.local_key has been renamed to
Component.component_code.

BREAKING CHANGE: Component.component_code now validates against
[A-Za-z0-9\-\_\.]+  and has a max_length of 255. Previously local_key used
key_field() (no regex validation, max_length=500).

BREAKING CHANGE: Function parameters renamed from local_key to component_code
in create_component(...) and create_component_and_version(...).

BREAKING CHANGE: Functions
get_component_by_key(...)/component_exists_by_key(...), renamed to
get_component_by_code(...)/component_exists_by_code(...), and parameters
renamed from local_key to component_code.

Backup-restore format is unchanged. component_type and component_code
will continue to be parsed from the entity key. This may change in
a future "v2" backup format.

Full series of PRs:
1. #542
2. #544
3. #545
4. #546
5. #547

### DIFF
--- a/src/openedx_content/applets/backup_restore/serializers.py
+++ b/src/openedx_content/applets/backup_restore/serializers.py
@@ -59,21 +59,49 @@ class EntityVersionSerializer(serializers.Serializer):  # pylint: disable=abstra
 class ComponentSerializer(EntitySerializer):  # pylint: disable=abstract-method
     """
     Serializer for components.
-    Contains logic to convert entity_key to component_type and local_key.
+
+    Extracts component_type and component_code from the [entity.component]
+    section if present (archives created in Verawood or later). Falls back to
+    parsing the entity key for archives created in Ulmo.
     """
+
+    component = serializers.DictField(required=False)
 
     def validate(self, attrs):
         """
-        Custom validation logic:
-        parse the entity_key into (component_type, local_key).
+        Custom validation logic: resolve component_type and component_code.
+
+        Archives created in Verawood or later supply an [entity.component]
+        section with ``component_type`` (e.g. "xblock.v1:problem") and
+        ``component_code`` (e.g. "my_example"). Archives created in Ulmo only
+        have the entity ``key`` in the format
+        ``"{namespace}:{type_name}:{component_code}"``, so we fall back to
+        parsing that for backwards compatibility.
         """
-        entity_key = attrs["key"]
-        try:
-            component_type_obj, local_key = components_api.get_or_create_component_type_by_entity_key(entity_key)
-            attrs["component_type"] = component_type_obj
-            attrs["local_key"] = local_key
-        except ValueError as exc:
-            raise serializers.ValidationError({"key": str(exc)})
+        component_section = attrs.pop("component", None)
+        if component_section:
+            # Verawood+ format: component_type and component_code are explicit.
+            component_type_str = component_section.get("component_type", "")
+            component_code = component_section.get("component_code", "")
+            try:
+                namespace, type_name = component_type_str.split(":", 1)
+            except ValueError as exc:
+                raise serializers.ValidationError(
+                    {"component": f"Invalid component_type format: {component_type_str!r}. "
+                                  "Expected '{namespace}:{type_name}'."}
+                ) from exc
+            component_type_obj = components_api.get_or_create_component_type(namespace, type_name)
+        else:
+            # Ulmo (legacy) format: parse the entity key.
+            entity_key = attrs["key"]
+            try:
+                component_type_obj, component_code = (
+                    components_api.get_or_create_component_type_by_entity_key(entity_key)
+                )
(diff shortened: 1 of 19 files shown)


src/openedx_content/applets/components/models.py
```diff
@@ -176,37 +177,37 @@ def pk(self):
...
+            # component_code "week_1".
             models.UniqueConstraint(
                 fields=[
                     "learning_package",
                     "component_type",
-                    "local_key",
+                    "component_code",
                 ],
                 name="oel_component_uniq_lc_ct_lk",
             ),
```
bradenmacdonald, line 200: Optional: add `code_field_check` here too?

bradenmacdonald (approved):
Thanks, looks great! I did not test this, just read the code.

src/openedx_content/applets/components/models.py
kdmccormick: yes, that was a miss, will do

src/openedx_content/applets/backup_restore/toml.py
```diff
@@ -108,6 +108,15 @@ def _get_toml_publishable_entity_table(
...
     entity_table.add("published", published_table)
 
+    if hasattr(entity, "component"):
+        component = entity.component
+        component_table = tomlkit.table()
+        # Write component_type and component_code explicitly so that restore
+        # (Verawood and later) does not need to parse the entity key.
+        component_table.add("component_type", str(component.component_type))
+        component_table.add("component_code", component.component_code)
+        entity_table.add("component", component_table)
```
ormsbee: Question: What happens when an Ulmo instance reads this?

ormsbee (commented):
Just a question about backwards compatibility. I'm not sure it's a blocker, but I would like to understand it better.

src/openedx_content/applets/backup_restore/toml.py
ormsbee: I mean, I get that it won't understand the new component_code piece. But does it successfully import components that have an explicit `[component]` section, since that wasn't being written to the TOML previously?

kdmccormick: Hm, I'm currently hitting an issue where any archive which I unzip or re-zip fails to restore, regardless of whether I edit the contents or not, with the message `Errors encountered during restore:
package.toml: Missing learning package file.` Archive that are never unzipped restore fine. I'm seeing this on the Ulmo sandbox, and I'm zipping using macOS's built in "Compress" utility. I'll try it later locally and see if I can get logs, and will try re-zipping with other methods....

kdmccormick: Haven't figured out this ⬆️ issue yet, but I did verify through unit tests that `[component]` and `[container]` seem safe to add without breaking Ulmo: https://github.com/openedx/openedx-core/pull/552

kdmccormick: > Hm, I'm currently hitting an issue where any archive which I unzip or re-zip fails to restore, regardless of whether I edit the contents or not, with the message Errors encountered during restore: package.toml: Missing learning package file.

ah, duh, I was making a ZIP where collection, entities and package.toml were nested within a top-level rather than being top-level items themselves.

with that figured out, I verified manually that `[component]` and `[container]` are safe to add 👍🏻

kdmccormick: Ohp, just realized I should be testing `[entity.component]` and `[entity.container]`. Testing now

kdmccormick: See discussion on https://github.com/openedx/openedx-core/pull/552

src/openedx_content/applets/backup_restore/serializers.py
```diff
@@ -59,24 +60,55 @@ class EntityVersionSerializer(serializers.Serializer):  # pylint: disable=abstra
...
+
+    Raises ValueError if the entity_key is not in the expected format.
+
+    Historical note: In Ulmo, this function was part of the public API. This was
+    inappropriate because the exact format of entity_keys is just a convention
+    rather than something API callers should count on. That said, it is safe to
+    assume that in all "v1" archives, the components' entity keys are safe to
+    parse into (namespace, type, code). So, we have moved this parsing logic
+    from the public API to just this internal halper function.  Future devs,
+    please to not make new external guarantees about the format of entity keys
```
kdmccormick:
```suggestion
    please do not make new external guarantees about the format of entity keys
```

src/openedx_content/applets/components/models.py
kdmccormick: done

ormsbee (approved):

PR #545, opened 2026-04-15
feat!: Add Container.container_code field

kdmccormick:
## Description

Adds a container_code field (code_field) and a learning_package FK to the
Container model. Also adds a UniqueConstraint on (learning_package,
container_code), which is stricter than Component's constraint (no type scoping
-- container codes must be unique across all container types within a given
LearningPackage).

For existing containers, container_code is backfilled from the entity key via a
data migration. Future containers will have container_code set explicitly by the
caller.

BREAKING CHANGE: In create_container() and create_container_and_version(), the
`key` argument has been renamed to `container_code`. The same applies to
create_unit_and_version(), create_subsection_and_version(), and
create_section_and_version().

Backup-restore format is unchanged. container_code will be set to the value of
the entity key. This may change in a future "v2" backup format.

Full series of PRs:
1. #542
2. #544
3. #545
4. #546
5. #547

### DIFF
--- a/src/openedx_content/applets/containers/api.py
+++ b/src/openedx_content/applets/containers/api.py
@@ -140,6 +140,7 @@ def create_container(
     created: datetime,
     created_by: int | None,
     *,
+    container_code: str,
     container_cls: type[ContainerModel],
     can_stand_alone: bool = True,
 ) -> ContainerModel:
@@ -152,6 +153,8 @@ def create_container(
         key: The key of the container.
         created: The date and time the container was created.
         created_by: The ID of the user who created the container
+        container_code: A local slug identifier for the container, unique within
+            the learning package (regardless of container type).
         container_cls: The subclass of container to create (e.g. `Unit`)
         can_stand_alone: Set to False when created as part of containers
 
@@ -170,7 +173,9 @@ def create_container(
         )
         container = container_cls.objects.create(
             publishable_entity=publishable_entity,
+            learning_package_id=learning_package_id,
             container_type=container_cls.get_container_type(),
+            container_code=container_code,
         )
     return container
 
@@ -341,6 +346,7 @@ def create_container_and_version(
     learning_package_id: LearningPackage.ID,
     key: str,
     *,
+    container_code: str,
     title: str,
     container_cls: type[ContainerModel],
     entities: EntityListInput | None = None,
@@ -354,6 +360,8 @@ def create_container_and_version(
     Args:
         learning_package_id: The learning package ID.
         key: The key.
+        container_code: A local slug identifier for the container, unique within
+            the learning package (regardless of container type).
         title: The title of the new container.
         container_cls: The subclass of container to create (e.g. Unit)
         entities: List of the entities that will comprise the entity list, in
@@ -371,6 +379,7 @@ def create_container_and_version(
             key,
             created,
             created_by,
+            container_code=container_code,
             can_stand_alone=can_stand_alone,
             container_cls=container_cls,
         )
--- a/src/openedx_content/applets/containers/api.py
+++ b/src/openedx_content/applets/containers/api.py
@@ -136,11 +136,10 @@ class ParsedEntityReference:
 
 def create_container(
     learning_package_id: LearningPackage.ID,
-    key: str,
+    container_code: str,
     created: datetime,
     created_by: int | None,
     *,
-    container_code: str,
     container_cls: type[ContainerModel],
     can_stand_alone: bool = True,
 ) -> ContainerModel:
@@ -150,11 +149,10 @@ def create_container(
 
     Args:
         learning_package_id: The ID of the learning package that contains the container.
-        key: The key of the container.
-        created: The date and time the container was created.
-        created_by: The ID of the user who created the container
         container_code: A local slug identifier for the container, unique within
             the learning package (regardless of container type).
+        created: The date and time the container was created.
+        created_by: The ID of the user who created the container
         container_cls: The subclass of container to create (e.g. `Unit`)
         can_stand_alone: Set to False when created as part of containers
 
@@ -164,18 +162,21 @@ def create_container(
     assert issubclass(container_cls, Container)
     assert container_cls is not Container, "Creating plain containers is not allowed; use a subclass of Container"
     with atomic():
+        # By convention, a Container's entity_key is set to its container_code.
+        # Do not bake this assumption into other systems. We may change it at some point.
+        entity_key = container_code
         publishable_entity = publishing_api.create_publishable_entity(
             learning_package_id,
-            key,
+            entity_key,
             created,
             created_by,
             can_stand_alone=can_stand_alone,
         )
         container = container_cls.objects.create(
             publishable_entity=publishable_entity,
+            container_code=container_code,
             learning_package_id=learning_package_id,
             container_type=container_cls.get_container_type(),
-            container_code=container_code,
         )
     return container
 
@@ -344,9 +345,8 @@ def create_container_version(
 
 def create_container_and_version(
     learning_package_id: LearningPackage.ID,
-    key: str,
-    *,
     container_code: str,
+    *,
     title: str,
     container_cls: type[ContainerModel],
     entities: EntityListInput | None = None,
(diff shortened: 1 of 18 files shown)


src/openedx_content/applets/containers/api.py
```diff
@@ -341,6 +346,7 @@ def create_container_and_version(
     learning_package_id: LearningPackage.ID,
     key: str,
     *,
+    container_code: str,
```
bradenmacdonald, line 349: The fact that this has `key` and `container_code` is just from how the PRs are split up right?

kdmccormick: Yep

bradenmacdonald: OK, I'm not sure why we have them as separate arguments here instead of just using `container_code = key` and then later renaming the `key` arg to `container_code`, but it really doesn't matter if it's just for this PR and it all get sorted out in the end.

src/openedx_content/applets/containers/models.py
```diff
@@ -171,6 +172,12 @@ class Container(PublishableEntityMixin):
     olx_tag_name: str = ""
     _type_instance: ContainerType  # Cache used by get_container_type()
 
+    # This foreign key is technically redundant because we're already locked to
+    # a single LearningPackage through our publishable_entity relation. However,
+    # having this foreign key directly allows us to make indexes that efficiently
+    # query by other Container fields within a given LearningPackage.
+    learning_package = models.ForeignKey(LearningPackage, on_delete=models.CASCADE)
```
bradenmacdonald, line 179: Very much non-blocking:

I wish we had some way to enforce the integrity of these denormalized relationships.

One idea could be in the future to write a [system check](https://docs.djangoproject.com/en/6.0/topics/checks/) that inspects the database and verifies everything. System checks which depend on the database do not normally run on startup (for performance) but can by run manually with the `./manage.py check` command.

src/openedx_content/applets/containers/models.py
```diff
@@ -194,6 +206,14 @@ def pk(self):
         # override this with a deprecated marker, so it shows a warning in developer's IDEs like VS Code.
         return self.id
 
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["learning_package", "container_code"],
+                name="oel_container_uniq_lp_cc",
+            ),
```
bradenmacdonald, line 214: Optional: Suggest adding code constraint here too.

src/openedx_content/applets/units/api.py
```diff
@@ -33,6 +33,7 @@ def create_unit_and_version(
     learning_package_id: LearningPackage.ID,
     key: str,
     *,
+    container_code: str,
```
bradenmacdonald:
```suggestion
    container_code: str,
    *,
```

This would make more sense to me, and then pass `key=container_code, container_code=container_code` or just `container_code=container_code` in the part below.

But if this is definitely just a temporary thing until the next PR merges, no problem.

tests/openedx_content/applets/publishing/test_api.py
```diff
@@ -1129,6 +1129,7 @@ def test_parent_child_side_effects(self) -> None:
             "my_container",
             created=self.now,
             created_by=None,
+            container_code="my_container",
```
bradenmacdonald: I mentioned this on the `key` removal PR, but do we really want to change this code from a positional second arg to a kwarg? Leaving it as positional second arg will make this change simpler on the platform side and even reduce the diff in this test code.

Non-blocking.

bradenmacdonald (approved):
I'm fine with this, assuming it's immediately followed by the proper removal of the `key` field, but I'd prefer to replace `key` more 1 for 1 as mentioned. Did not run the code, just read.

src/openedx_content/migrations/0010_add_container_code.py
```diff
@@ -0,0 +1,88 @@
...
+
+    For existing containers, container_code is set to the entity key (the
+    only identifier available at this point). Future containers will have
+    container_code set by the caller.
+    """
+    Container = apps.get_model("openedx_content", "Container")
+    for container in Container.objects.select_related("publishable_entity__learning_package").all():
+        container.learning_package = container.publishable_entity.learning_package
+        container.container_code = container.publishable_entity.key
+        container.save(update_fields=["learning_package", "container_code"])
```
ormsbee, line 22: [Comment, no action needed] It's really unfortunate that we can't use `F()` across relations. I feel like this is one of the few cases where I'd be tempted to use raw SQL.

src/openedx_content/migrations/0010_add_container_code.py
```diff
@@ -0,0 +1,88 @@
...
+                on_delete=django.db.models.deletion.CASCADE,
+                to="openedx_content.learningpackage",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="container",
+            name="container_code",
+            field=openedx_django_lib.fields.MultiCollationCharField(
+                db_collations={"mysql": "utf8mb4_bin", "sqlite": "BINARY"},
+                max_length=255,
```
ormsbee, line 69: It's not necessary to explicitly add `null=False` here?

ormsbee (commented):
Just a sanity check question, but otherwise LGTM.

src/openedx_content/migrations/0010_add_container_code.py
kdmccormick: @ormsbee oof, thank you for catching that. All of collection_code, container_code, and component_code are missing `null=False` on their definition and migrations. I'll fix them all.

kdmccormick: @ormsbee actually the Django docs say that `null=False, blank=False` is the default on any CharField, so this should be fine as-is.

src/openedx_content/applets/containers/models.py
kdmccormick: Done

src/openedx_content/applets/containers/api.py
kdmccormick: When I did it this way originally, it was back when I thought we'd need to let API callers customize the `entity_refs` of components and containers. Since we're not worried about that, I agree, it's better just to rename the arg in-place from `key` to `container_code`. Making that change now.

kdmccormick: Done

src/openedx_content/applets/containers/models.py
kdmccormick: This is interesting. Do you mind opening a ticket with this idea?

kdmccormick:
Thanks for the review @bradenmacdonald . I've applied all your suggestions except the denormalization check. Could you take another look?

src/openedx_content/applets/containers/models.py
bradenmacdonald: Sure, done: https://github.com/openedx/openedx-core/issues/556

src/openedx_content/applets/containers/models.py
```diff
@@ -179,6 +186,11 @@ class Container(PublishableEntityMixin):
         editable=False,
     )
 
+    # container_code is an identifier that is local to the learning_package.
+    # Unlike component_code, it is unique across all container types within
+    # the same LearningPackage.
```
bradenmacdonald, line 191: Will requiring this code to be unique across all containers in a [course] bite us later when we try to migrate old courses to Learning Core, since they may have e.g. a section and unit with the same `container_code` ?

bradenmacdonald (approved):
Looks great! I read through the code, but didn't test; let me know if you want any testing on these PRs.

kdmccormick:
@bradenmacdonald all set for now, but if you if you have time on Wednesday I could use a hand testing the openedx-platform PR.

src/openedx_content/applets/containers/models.py
kdmccormick: Hard to say. 

On one hand, we'll likely have multiple course runs per learning package. So, we'll need to handle collisions between container block url_names anyway. I'm imagining that there'll have to be a table which maps CourseRunUsages to Entities, rather than a simple mapping function between `{component,container}_{type,code}` and CourseRunUsageKeys.

On the other hand, it's weird to me that container_code and component_code have different uniqueness constraints. I have a feeling that the inconsistency will make consuming code more complicated rather than having any benefit to us.

We're riding the line now, but maybe we can sneak a fix to this in before the cutoff.

bradenmacdonald: Yeah, I'd prefer to make container_code work like component_code if we can.

PR #546, opened 2026-04-15
feat!: Package and Entity `keys` are now opaque `refs`

kdmccormick:
## Description

The `{LearningPackage,PublishableEntity}.key` fields were always meant to be externally-supplied refs whose format does not matter. However, they were effectively being used as parseable psuedo-keys for identifying containers and components, with more assumptions to their structure being baked in over time, both within and outside openedx-core.

This commit renames them to `{package,entity}_ref`, and removes all parsing of them from the API and from the internals (except for where required for backcompat with Ulmo ZIP backups). Instead, the pairings of (component_type,component_code) and (container_type,container_code) are used to identify components and containers; their respective entity_refs are derived from those *by convention* but the publishing app has no conception of this convention.

BREAKING CHANGES to **openedx_django_lib**:
* Renamed key_field(...) -> ref_field(...) in openedx_django_lib.

BREAKING CHANGES to **models** API:
* Renamed field LearningPackage.key -> LearningPackage.package_ref.
* Renamed field PublishableEntity.key -> PublishableEntityKey.entity_ref
* Renamed field PublishableEntityMixin.key -> PublishableEntityMixin.entity_ref
* Renamed property PublishableEntityVersion.key -> PublishableEntityVersion.entity_ref

BREAKING CHANGES to **publishing** API:
* In create_learning_package(...), renamed param key -> package_ref
* In update_learning_package(...), renamed param key -> package_ref
* In learning_package_exists(...), renamed param key -> package_ref
* In create_publishable_entity(...), renamed param key -> entity_ref
* Renamed get_publishable_entity_by_key(...) -> get_publishable_entity_by_ref(...), and renamed param key -> entity_ref
* Renamed get_learning_package_by_key(...) -> get_learning_package_by_ref(...), and renamed param key -> package_ref

BREAKING CHANGES to **components** APIs:
* Removed function get_or_create_component_type_by_entity_key()

BREAKING CHANGES to **container** APIs:
* Renamed get_container_by_key(...) -> get_container_by_code(...), and renamed param key -> container_code
* Rename get_container_children_entities_keys(...) -> get_container_children_entity_refs(...)
* Renamed get_container_children_entities_keys(...) -> get_container_children_entity_refs(...)

BREAKING CHANGES to **collections** API:
* In get_entity_collections(...), renamed param entity_key -> entity_ref

BREAKING CHANGES to **media** API:
* In look_up_component_version_media(...):
  * Renamed param learning_package_key -> learning_package_ref
  * Renamed param component_key -> entity_ref
* In add_assets_to_component management command:
  * Renamed argument learning_package_key -> learning_package_ref
  * Renamed argument component_key -> entity_ref

BREAKING CHANGES to **backup_restore** API:
* BREAKING CHANGE In load_learning_package:
  * Renamed  param key -> package_ref
  * In return dict, within sub-dict ["lp_restored_data"]:
    * Renamed ["key"] -> ["package_ref"]
    * Renamed ["archive_lp_key"] -> ["archive_package_ref"]
    * Renamed ["archive_org_key"] -> ["archive_org_code"]
    * Renamed ["archive_slug"] -> ["archive_package_code"]
  * If archive_package_ref cannot be parsed into "{_prefix}:{org_code}:{package_code}", then archvie_org_code and archive_package_code will both be None. Callers should handle this case. (Previously, a ValueError would be raised by backup-restore code.)

Backup-restore still write and reads entity_ref and package_ref to/from TOML files as `key`
for backwards compatibility. This may change in a future "v2" restore format.

Full series of PRs:
1. #542
2. #544
3. #545
4. #546
5. #547

### DIFF
--- a/src/openedx_content/applets/backup_restore/zipper.py
+++ b/src/openedx_content/applets/backup_restore/zipper.py
@@ -473,6 +473,19 @@ def generate_staged_package_ref(archive_package_ref: str, user: UserType) -> str
     """
     Generate a staged learning package ref based on the archive's package_ref.
 
+    We can't trust package_ref from the archive directly, because the archive
+    could specify *any* arbitrary package_ref, and the user may or may not be
+    permitted to create an Package using that ref. So, instead, this function
+    generates a unique and semi-human-readable package_ref which is namespaced
+    to the current user and appropriate to provisionally save the package under.
+    The package_ref from the archive can then be presented to the user as a
+    *suggestion*, which they may or may not choose to use.
+
+    Please note that the ref returned by this function is valid for Packages is
+    a generic sense, but it's not a valid Content Library key.  Callers who are
+    restoring a Package for Library usage will need to replace this staged
+    package_ref before being able to render the Library's content.
+
     Arguments:
         archive_package_ref (str): The original package_ref from the archive.
         user (UserType | None): The user performing the restore operation.
@@ -480,9 +493,6 @@ def generate_staged_package_ref(archive_package_ref: str, user: UserType) -> str
     Example:
         Input:  "lib:WGU:LIB_C001"
         Output: "lp-restore:dave:WGU:LIB_C001:1728575321"
-
-    The timestamp at the end ensures the ref is unique. Falls back to using
-    the full archive_package_ref when the conventional format is not recognised.
     """
     username = user.username
     org_code, package_code = unpack_package_ref(archive_package_ref)
--- a/src/openedx_content/applets/backup_restore/zipper.py
+++ b/src/openedx_content/applets/backup_restore/zipper.py
@@ -749,7 +749,7 @@ class LearningPackageUnzipper:
     ) -> LearningPackage:
         """Persist all validated entities in two phases: published then drafts."""
 
-        # Important: If not using a specific LP key, generate a temporary one
+        # Important: If not using a specific LP ref/key, generate a temporary one
         # We cannot use the original key because it may generate security issues
         if not self.package_ref:
             # Generate a tmp ref for the staged learning package
@@ -972,7 +972,7 @@ class LearningPackageUnzipper:
             self,
             num_version: int,
             entity_ref: str,
-            component_type,
+            component_type: ComponentType,
             static_files_map: dict[str, List[str]]
     ) -> dict[str, bytes | int]:
         """Resolve static file paths into their binary media content."""
--- a/src/openedx_content/applets/backup_restore/zipper.py
+++ b/src/openedx_content/applets/backup_restore/zipper.py
@@ -202,7 +202,7 @@ class LearningPackageZipper:
                     to_attr="prefetched_media",
                 ),
             )
-            .order_by("key")
+            .order_by("entity_ref")
         )
 
     def get_collections(self) -> QuerySet[Collection]:
@@ -238,25 +238,25 @@ class LearningPackageZipper:
             versions_to_write.append(published_version)
         return versions_to_write, draft_version, published_version
 
-    def get_entity_toml_filename(self, entity_key: str) -> str:
+    def get_entity_toml_filename(self, entity_ref: str) -> str:
         """
         Generate a unique TOML filename for a publishable entity.
         Ensures that the filename is unique within the zip file.
 
         Behavior:
-        - If the slugified key has not been used yet, use it as the filename.
+        - If the slugified ref has not been used yet, use it as the filename.
         - If it has been used, append a short hash to ensure uniqueness.
 
         Args:
-            entity_key (str): The key of the publishable entity.
+            entity_ref (str): The entity_ref of the publishable entity.
 
         Returns:
             str: A unique TOML filename for the entity.
         """
-        slugify_name = slugify(entity_key, allow_unicode=True)
+        slugify_name = slugify(entity_ref, allow_unicode=True)
 
         if slugify_name in self.entities_filenames_already_created:
-            filename = slugify_hashed_filename(entity_key)
+            filename = slugify_hashed_filename(entity_ref)
         else:
             filename = slugify_name
 
@@ -316,7 +316,7 @@ class LearningPackageZipper:
                 )
 
                 if hasattr(entity, 'container'):
-                    entity_filename = self.get_entity_toml_filename(entity.key)
+                    entity_filename = self.get_entity_toml_filename(entity.entity_ref)
                     entity_toml_filename = f"{entity_filename}.toml"
                     entity_toml_path = entities_folder / entity_toml_filename
                     self.add_file_to_zip(zipf, entity_toml_path, entity_toml_content, timestamp=latest_modified)
@@ -403,11 +403,11 @@ class LearningPackageZipper:
             for collection in collections:
                 collection_hash_slug = self.get_entity_toml_filename(collection.collection_code)
                 collection_toml_file_path = collections_folder / f"{collection_hash_slug}.toml"
-                entity_keys_related = collection.entities.order_by("key").values_list("key", flat=True)
+                entity_refs_related = collection.entities.order_by("entity_ref").values_list("entity_ref", flat=True)
                 self.add_file_to_zip(
                     zipf,
                     collection_toml_file_path,
-                    toml_collection(collection, list(entity_keys_related)),
+                    toml_collection(collection, list(entity_refs_related)),
                     timestamp=collection.modified,
                 )
 
@@ -418,10 +418,10 @@ class RestoreLearningPackageData:
     Data about the restored learning package.
     """
     id: int  # The ID of the restored learning package
-    key: str  # The key of the restored learning package (may be different if staged)
-    archive_lp_key: str  # The original key from the archive
-    archive_org_key: str  # The original organization key from the archive
-    archive_slug: str  # The original slug from the archive
+    package_ref: str  # The package_ref of the restored learning package (may be different if staged)
+    archive_package_ref: str  # The original package_ref from the archive
+    archive_org_code: str | None  # The org code parsed from archive_package_ref, or None if unparseable
+    archive_package_code: str | None  # The package code parsed from archive_package_ref, or None if unparseable
     title: str
     num_containers: int
     num_sections: int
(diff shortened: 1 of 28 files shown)


src/openedx_content/applets/subsections/api.py
```diff
@@ -31,14 +31,14 @@ def get_subsection(subsection_id: Subsection.ID, /):
 
 def create_subsection_and_version(
     learning_package_id: LearningPackage.ID,
-    key: str,
```
ormsbee, line 34: I realize this is still in draft, but just a quick note here: The `entity_ref` should replace `key` directly here. One of the big reasons why this arg is positional-only was because we thought it was likely to get renamed at some point and wanted to make it so that we didn't have to update callers when that happened.

kdmccormick: @ormsbee @bradenmacdonald  , I think this is something we need to align on. My thinking was that `entity_ref` would become an optional parameter in the Container and Component APIs, defaulting to and `{container_code}` and `{component_type}:{component_code}`, respectively. That would preserve the current convention by default (entity refs and opaque keys match) but would allow API callers to customize `entity_ref` if they care to.

Dave, it sounds like you'd prefer `entity_ref` to always be an explicit required parameter?

kdmccormick: > I realize this is still in draft

They're only in Draft because I don't have testing instructions posted yet-- review comments are all welcome!

ormsbee: Ah, I see. I have no objection then.

bradenmacdonald: > would allow API callers to customize entity_ref if they care to.

When would they want to?

Unrelated, is too late to switch containers to use the convention `{container_type}:{container_code}` ?

src/openedx_content/migrations/0017_rename_learningpackage_db_column_key_to_package_ref.py
```diff
@@ -0,0 +1,39 @@
...
+import openedx_django_lib.fields
+
+
+class Migration(migrations.Migration):
+    """
+    Rename the underlying DB column for LearningPackage.package_ref from
+    '_key' to 'package_ref'.
+
+    Uses SeparateDatabaseAndState with RunSQL (RENAME COLUMN) for
+    SQLite/MySQL compatibility. The table is named
```
bradenmacdonald: As in the other PR, I don't understand from this explanation why Django's auto-generated migration wouldn't be "SQLite/MySQL" compatible ?

src/openedx_content/applets/subsections/api.py
kdmccormick: @bradenmacdonald Not too late, but I don't want to clog this PR stack up with it. I'm open to doing it in a followup PR.

bradenmacdonald: OK. I'm still wondering if we should just drop the ability to customize `entity_ref`. Do we have a use case in mind?

kdmccormick: @ormsbee , should we allow customization of `entity_ref` right now?

ormsbee: I'm not sure I understand correctly. We already have arbitrarily named sections/subsections/units that don't follow any sort of {containter_type}:{container_code} convention. I think I'd prefer to have `entity_ref` as an opaque identifier whose only real purpose is to guarantee uniqueness to the publishing system and maybe give a hint to users.

If we want to make a more logically structured and uniform lookup like (container_type, container_code), I think we should add an explicit `container_code` field to `Container`, backfill as necessary, and make that the way we look up containers going forward.

src/openedx_content/migrations/0017_rename_learningpackage_db_column_key_to_package_ref.py
kdmccormick: I think what happened here is that Claude wrote the migration file itself instead of invoking `./manage.py makemigrations`, and I figured that it must have done so for a good reason so I trusted the comment, but I challenged it now to just use a regular old `AlterField` and it works fine. Sorry about that.

kdmccormick: I think what happened here is that Claude wrote the migration file itself instead of invoking `./manage.py makemigrations`, and I figured that it must have done so for a good reason so I trusted the comment, but I challenged it now to just use a regular old `AlterField` and it works fine. My bad, sorry about that.

src/openedx_content/applets/subsections/api.py
kdmccormick: @ormsbee what you've described is exactly what I'm going for with these PRs. #545 adds `container_code` and backfills it, and this PR demotes `key` (a canonical identifier) into `entity_ref` (just for publishing uniqueness and hinting).

The question on the table now is: Do we allow Component and Container API callers to specify arbitrary `entity_refs` going forward? Or do we always derive their refs from their respective `_type` and/or `_code` fields, only allowing `entity_ref` to be arbitrarily set if you're creating a bare PublishableEntity?

bradenmacdonald: I think we should

> always derive their refs from their respective `_type` and/or `_code` fields, only allowing `entity_ref` to be arbitrarily set if you're creating a bare PublishableEntity?

unless we have a compelling use case for overriding the entity_ref when creating a container/component. I am not aware of one, but let me know if I'm missing something.

ormsbee: Got it. I agree with @bradenmacdonald in this case that if we don't have an immediate use case, it's better to not expose the capability yet. With the understanding that it's there in our back pocket if we ever do need to add it.

kdmccormick: sounds great

kdmccormick: done

src/openedx_content/applets/components/api.py
```diff
@@ -529,13 +511,13 @@ def _get_component_version_info_headers(component_version: ComponentVersion) ->
...
         # Component
-        "X-Open-edX-Component-Key": component.publishable_entity.key,
+        "X-Open-edX-Component-Key": component.publishable_entity.entity_ref,
         "X-Open-edX-Component-Uuid": component.uuid,
         # Component Version
         "X-Open-edX-Component-Version-Uuid": component_version.uuid,
         "X-Open-edX-Component-Version-Num": str(component_version.version_num),
         # Learning Package
-        "X-Open-edX-Learning-Package-Key": learning_package.key,
+        "X-Open-edX-Learning-Package-Key": learning_package.package_ref,
```
kdmccormick, line 520: @ormsbee I should probably fix these header names, right?

src/openedx_content/applets/containers/api.py
```diff
@@ -344,7 +343,6 @@ def create_container_version(
 
 def create_container_and_version(
     learning_package_id: LearningPackage.ID,
-    key: str,
     *,
     container_code: str,
```
bradenmacdonald, line 347: Since `key` was before the `*`, won't it help ease compatibility and consistency if we move `container_code` to be where `key` was, and allow it to be a positional argument?

src/openedx_content/applets/components/admin.py
```diff
@@ -37,16 +37,16 @@ class ComponentAdmin(ReadOnlyModelAdmin):
...
         "learning_package",
         "uuid",
         "component_type",
-        "key",
+        "component_code",
         "created",
     ]
     list_filter = ("component_type", "learning_package")
-    search_fields = ["publishable_entity__uuid", "publishable_entity__key"]
+    search_fields = ["publishable_entity__uuid", "publishable_entity__entity_ref"]
```
bradenmacdonald, line 49: Optional: Shouldn't this just be `component_code` for searching? Also, unrelated, but do we need to search by UUID?


```suggestion
    search_fields = ["component_code"]
```

src/openedx_content/applets/components/api.py
ormsbee: @kdmccormick: Ah, right. Yeah, it'd be good to fix. It's not like anyone's relying on these values as they are now--they're just debug information at the moment.

ormsbee: I mean, clearly not a blocker and doesn't need to happen for Verawood, but it makes sense to update.

src/openedx_content/applets/containers/api.py
kdmccormick: will do

src/openedx_content/applets/components/api.py
kdmccormick: Fixed

kdmccormick:
This is almost ready, but I'm still wrestling with translation between the backup-restore terms (keys) and the model terms (refs).

EDIT: It's all ready now.

tests/openedx_content/applets/backup_restore/test_restore.py
```diff
@@ -363,23 +363,21 @@ def test_success_metadata_using_user_context(self):
 class RestoreUtilitiesTest(TestCase):
     """Tests for utility functions used in the restore process."""
 
-    def test_generate_staged_lp_key(self):
+    def test_generate_staged_package_ref(self):
         """Test generating a staged learning package key."""
```
bradenmacdonald:
```suggestion
        """Test generating a staged learning package ref."""
```

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -743,16 +751,16 @@ def _save(
 
         # Important: If not using a specific LP key, generate a temporary one
```
bradenmacdonald:
```suggestion
        # Important: If not using a specific LP ref/key, generate a temporary one
```

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -454,35 +454,43 @@ class RestoreResult:
...
+        return None, None
+    _, org_code, package_code = parts[:3]
+    return org_code, package_code
 
 
-def generate_staged_lp_key(archive_lp_key: str, user: UserType) -> str:
+def generate_staged_package_ref(archive_package_ref: str, user: UserType) -> str:
     """
-    Generate a staged learning package key based on the given base key.
+    Generate a staged learning package ref based on the archive's package_ref.
```
bradenmacdonald, line 474: Optional request for additional docstring info: What is the purpose of this staged package ref?

All I see below is

> Important: If not using a specific LP key, generate a temporary one
> We cannot use the original key because it may generate security issues

So I assume the purpose of this function is just to generate a temporary ref that is (A) unique, and (B) somewhat descriptive for debugging purposes ?

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -949,20 +965,21 @@ def _is_version_already_exists(self, entity_key: str, version_num: int) -> bool:
...
-        identifier = (entity_key, version_num)
+        identifier = (entity_ref, version_num)
         return identifier in self.all_published_entities_versions
 
     def _resolve_static_files(
             self,
             num_version: int,
-            entity_key: str,
+            entity_ref: str,
+            component_type,
```
bradenmacdonald:
```suggestion
            component_type: ComponentType,
```
?

bradenmacdonald (approved):
I read through the code and it looks great. Some very minor suggestions.
I did not test anything, but can do so when the platform PR is ready if you'd like.

src/openedx_content/applets/backup_restore/zipper.py
kdmccormick: @bradenmacdonald good idea, I did some research and added to the docstring.

bradenmacdonald: Thank you!

PR #551, opened 2026-04-19
fix: TypeError when restoring published containers

kdmccormick:
## Description

The original version of this code depended upon `version_num` being popped out
of `valid_published` before `valid_published` got expanded. A [recent change](https://github.com/openedx/openedx-core/pull/495)
changed the argument order, leading to:

    TypeError: create_next_container_version() got an unexpected keyword argument
    'version_num'

whenever restoring an archive with any published containers. This commit
restores the original code's order of operations, but a bit more explicitly.

Bumps from 0.39.2 to 0.40.0 (it would be 0.39.3, but the previous
commit had breaking changes and neglected to bump the minor version)

## Testing

On openedx-platform master and openedx-core==v0.39.0, back up any library with at least one published component.

Try to restore it. It'll fail. You should see a TypeError in the logs.

Cherry-pick this commit on top of openedx-core v0.39.0. Try again to restore the archive. It should succeed.

### DIFF
--- a/src/openedx_core/__init__.py
+++ b/src/openedx_core/__init__.py
@@ -6,4 +6,4 @@ The public APIs belong to the specific apps (openedx_content, openedx_tagging, e
 """
 
 # The version for the entire repository
-__version__ = "0.39.2"
+__version__ = "0.40.0"
--- a/tests/openedx_content/applets/backup_restore/test_restore.py
+++ b/tests/openedx_content/applets/backup_restore/test_restore.py
@@ -66,13 +66,19 @@ class RestoreLearningPackageCommandTest(RestoreTestCase):
             published_version = publishing_api.get_published_version(container.publishable_entity.id)
             assert container.created_by is not None
             assert container.created_by.username == "lp_user"
+            # The unit has been published. The other two containers haven't been.
+            # It's important that we test with at least one published container in order to
+            # fully cover _create_container in zipper.py.
             if container.key == "unit1-b7eafb":
                 assert containers_api.get_container_type_code_of(container) == "unit"
                 assert draft_version is not None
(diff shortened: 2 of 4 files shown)


tests/openedx_content/applets/backup_restore/test_restore.py
```diff
@@ -66,13 +66,19 @@ def verify_containers(self, lp):
             published_version = publishing_api.get_published_version(container.publishable_entity.id)
             assert container.created_by is not None
             assert container.created_by.username == "lp_user"
+            # The unit has been published. The other two containers haven't been.
+            # It's important that we test with at least one published container in order to
+            # fully cover _create_container in zipper.py.
```
kdmccormick, line 71:
```suggestion
            # fully cover _save_container in zipper.py.
```
(I typo'd the function name)

kdmccormick:
@bradenmacdonald @ormsbee a quick fix to a bug on openedx-core main, could you take a look?

src/openedx_core/__init__.py
```diff
@@ -6,4 +6,4 @@
 """
 
 # The version for the entire repository
-__version__ = "0.39.2"
+__version__ = "0.40.0"
```
ormsbee, line 9: Since it's a bugfix, shouldn't this be 0.39.3?

ormsbee (approved):
Code looks fine. Just a comment about the version.

bradenmacdonald (approved):
Thanks! Yeah the current backup/restore code doesn't have enough test coverage.

src/openedx_core/__init__.py
kdmccormick: The collection_code change is also released by this, which makes it a breaking release

PR #552, opened 2026-04-19
test: Ensure back-compat of new _code fields in TOML archives

kdmccormick:
[Do Not Merge]

This is a test PR to see how changes to the backup-restore format in #544 and #545 would work on Ulmo instance.

This PR is based on v0.30.2, which is what `release/ulmo` pins.

Context: https://github.com/openedx/openedx-core/pull/544#discussion_r3102780605

### DIFF
--- a/tests/openedx_learning/apps/authoring/backup_restore/fixtures/library_backup/entities/subsection1-48afa3.toml
+++ b/tests/openedx_learning/apps/authoring/backup_restore/fixtures/library_backup/entities/subsection1-48afa3.toml
@@ -9,7 +9,14 @@ version_num = 2
 [entity.published]
 # unpublished: no published_version_num
 
+# Starting with https://github.com/openedx/openedx-core/pull/545,
+# containers will derive their identity from `container_code` instead of
+# from `key` (although archives without a `container_code` will fall back
+# to `key`). We're adding this bit to the test fixture ahead of time to
+# ensure that older (Ulmo) code is forward-compatible with archives
+# containing `container_code`.
 [entity.container.subsection]
+container_code = "subsection1-48afa3"
 
 # ### Versions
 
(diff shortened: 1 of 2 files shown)


kdmccormick:
@ormsbee @bradenmacdonald Could I get your take on which of these formats to use for Verawood? I'm leaning 2 or 3.

All of these could be implemented to be backwards-compatible with Ulmo instances. Existing Ulmo-generated archives would be forwards compatible with Verawood instances, as the Verawood instance would just fall back to parsing `[entity].key`

### Option 1
This is what I have implemented now.
```toml
[entity.component]
component_type = "xblock.v1:html"
component_code = "c22b9f97-f1e9-4e8f-87f0-d5a3c26083e2"
```

### Option 2
This would match how an `[entity.container]` is defined. You could imagine additional component-type-specific data going into this table.
```toml
[entity.component."xblock.v1".html]
component_code = "c22b9f97-f1e9-4e8f-87f0-d5a3c26083e2"
```

### Option 3

**No changes for Verawood.** Just continue parsing`[entity].key`. A future v2 format would implement something like the options above, and it would replace `[entity].key` with `[entity].entity_ref`.

kdmccormick:
For containers, I have similar question. Option 1 doesn't make sense given the existence of `[entity.container.<subclass>]` in the format. So, it's either:

### Option 2 

```toml
[entity.container.subsection]
container_code = "unit-foo-bar-baz"
```

## Option 3

No changes. Just continue using `[entity].key` as the container_code.

ormsbee:
I think (1) or (3). The container_code and container_type are things handled by the `container` applet, and are not specific to a kind of container. Likewise with `component`.

kdmccormick:
Option (1) is unfortunately impossible for containers without breaking Ulmo backcompat, due to how strict this existing validation code is:

https://github.com/openedx/openedx-core/blob/cbc344b9f3c15055dc23f9dd421aab9080050cc5/src/openedx_content/applets/backup_restore/serializers.py#L131-L135

That is, the following TOML
```toml
[entity.container]
container_code = "blah"

[entity.container.subsection]
...
```

will yield `"Container must be a dict with exactly one key."`.

Given that, I think we should keep things simple and just wait until "v2" before making any backup-restore format changes.

As an aside @ormsbee @bradenmacdonald , if/when we do TOML v2, I think we should aim to make Restore fairly permissive as to allow future backwards-compatible additions to the format.

PR #554, opened 2026-04-20
[WIP] Re-implementing Restore to use Pydantic

ormsbee:
This code will import things through the existing Studio interface, but it doesn't report errors properly. It's a parallel implementation (most of the original restore code still exists so that I can copy tests and compare implementation logic).

If you want to test on the command line, the new (temporary) command `lp_load2` will work. You'd invoke it like this:

```
python manage.py cms lp_load2 [path] [package_ref] [username]
```

Starting point for all the logic is in [`backup_restore.api.load_learning_package`](https://github.com/openedx/openedx-core/pull/554/changes#diff-2e119ea0879d0fd914fed8e74c38e158efcd29dbd949a29c9556bb37fa5b1755R33).

New modules:

* `payload.py` is where the extraction of TOML files happens, and things get assembled into a big dict for validation. It would also eventually be used to map the other way. It understands the backup archive's internal format.
* `validation.py` is where that's gets turned into a validated Pydantic model and errors will be accumulated (might just fold this into `api.py`).
* `schema.py` defines the actual Pydantic models. It needs a lot of validation rules (just the bare field stuff exists now).
* `loading.py` is the logic for actually loading validated input into the database. It initializes a `Loader` object with input data and then calls  `loader.load_into()` to push that data into a target `LearningPackage`.

I have a lot of tests to write and some details to work out regarding error reporting, but I think I'm comfortable with this as the basic structure. I intend to keep full compatibility with the existing REST endpoint and just do conversions as needed.

### DIFF
--- a/src/openedx_content/applets/backup_restore/payload.py
+++ b/src/openedx_content/applets/backup_restore/payload.py
@@ -1,0 +1,496 @@
+"""
+This module works with the actual files in our backup archive. It is agnostic to
+the archive container format that the files are bundled in, e.g. a local file
+system directory, a zip file archive, or something more exotic down the line.
+
+Some high level considerations for this module:
+
+1. The error checking is for the file format itself, i.e. extracting  values
+   from the TOML files and statica assets and assembling them for validation.
+   In some cases, this means we do have to look for particular fields to handle
+"""
+
+from __future__ import annotations
+from numbers import Number
+import os.path  # fsspec doesn't work well with Path objects.
+import tomllib
+
+import attrs
+from fsspec import AbstractFileSystem
+
+ROOT_PACKAGE_PATH = "package.toml"
+
+
+@attrs.define(frozen=True)
+class UnvalidatedLearningPackageInput:
+    raw_data: dict
+    errors: list[ExtractionError]
+    fs: AbstractFileSystem
+
+    # Mapping of entity refs to the paths where we found them.
+    entity_path_mapping: dict[str, str]
+
+
+class ExtractionError(Exception):
+    """
+    Any error during the extraction process.
+
+    At the moment, any error is fatal. The point of the different errors is to
+    provide useful debug logging and to let us write tests that look for
+    specific errors.
+    """
+
+    def __init__(self, message, path=None):
+        super().__init__(message)
+        self.message = message
+        self.path = path
+
+    def __str__(self):
+        return f"{self.path}: {self.message}"
+
+
+class InvalidTOMLError(ExtractionError):
+    def __init__(self, file_description, details, path):
+        message = f"Cannot decode TOML for {file_description}: {details}"
+        super().__init__(message, path=path)
+
+
+class TableNotFoundError(ExtractionError):
+    def __init__(self, file_description, table, path):
+        self.table = table
+        message = f"Table [{table}] not found in {file_description}."
+        super().__init__(message, path=path)
+
+
+class FieldsNotInTable(ExtractionError):
+    def __init__(self, file_description, fields, path):
+        self.fields = sorted(fields)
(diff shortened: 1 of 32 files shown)


ormsbee:
@bradenmacdonald, @kdmccormick: The WIP (but functioning) new restore code. There are a lot of missing low-level things and tests, so it's not really ready for review. But please let me know if you have any high level concerns with the overall approach. Thank you.

src/openedx_content/applets/backup_restore/payload.py
```diff
@@ -0,0 +1,253 @@
...
+            media = {
+                os.path.relpath(path, comp_ver_dir): fs.read_text(path)
+                for path in fs.glob(f"{comp_ver_dir}/*")
+                if fs.isfile(path)
+            }
+            # Any static files are encoded as pointers.
+            for static_file_path in fs.glob(f"{comp_ver_dir}/static/**"):
+                if fs.isfile(static_file_path):
+                    rel_path = os.path.relpath(static_file_path, comp_ver_dir)
+                    media[rel_path] = f"fs:{static_file_path}"
```
ormsbee, line 431: @kdmccormick, @bradenmacdonald: I was worried about storing potentially enormous inline assets in a giant JSON doc, so I'm thinking of building the component version media mapping with this convention:
* "text:" will be simple text
* "base64:" indicates base64 encoded data is to follow
* "fs:" indicates that what follows is a path from the filesystem object we're passing around from the input.

So an example:

```json
{
  "block.xml": "text:<problem>(bunch of OLX here)</problem>",
  "static/figure1.webp": "fs:entities/xblock.v1/problem/intro-to-everything/component_versions/v2/static/figure1.webp",
  "static/logo.svg": "base64:SXMgc29tZW9uZSBjdXJpb3VzIGVub3VnaCB0byBhY3R1YWxseSBkZWNvZGUgdGhpcz8="
}
```

bradenmacdonald: Sounds nice, but that won't be backwards compatible right?

It's also very similar to [data URLs](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data); I like the simplicity here but could be worth considering using standard data URLs since there are already lots of libraries for parsing and generating them. (Although the syntax is not as nice; you have these mandatory `,` and `;` which look weird when omitting MIME types.)

```json
{
  "block.xml": "data:,<problem>(bunch of OLX here)</problem>",
  "static/logo.svg": "data:;base64,SXMgc29tZW9uZSBjdXJpb3VzIGVub3VnaCB0byBhY3R1YWxseSBkZWNvZGUgdGhpcz8="
}
```

ormsbee: > Sounds nice, but that won't be backwards compatible right?

This would be for the currently internal format after things get extracted and pieced together, but before Pydantic model validation. So it wouldn't break backwards compatibility with anything.

> It's also very similar to data URLs;

That occurred to me, but I had only ever used those for base64 encoding binary data. I hadn't realized you could use them for plain text in a clear, non-escaped way. Thank you! I'll do that I think.  So if it starts with "data:", I'll parse it like a data URL, and if it's just a path, I'll assume it comes from the archive.

ormsbee:
Note to self: Make restore more forgiving for those that make a zip with a root directory, i.e. when the `package.toml` is in `library-folder-name/package.toml` instead of being at the top level.

kdmccormick:
^ I have hit this a lot myself :) Even if we don't want to allow it in the format, a nicer error message in this case would help.

ormsbee:
Closing in favor of https://github.com/openedx/openedx-core/pull/792

PR #558, opened 2026-04-23
feat: Reinstate support for Unicode component_codes and container_codes

kdmccormick:
[While upgrading platform from core 0.39.2 to core 0.43.0](https://github.com/openedx/openedx-platform/pull/38402#discussion_r3121151211), we realized that we've secretly supported non-ascii component and container codes in V2 content libraries since their release. We do not advertise our support for non-ascii codes (and we don't want to, [yet](https://github.com/openedx/openedx-platform/issues/38413)), but in order to avoid breaking sites that stumbled upon non-ascii codes, we need to loosen the validation & db constraints which were added to openedx-core.

On a code level, we simply add a `unicode=True/False` parameter to the `code_field()`. I've set it True for components and containers, and left it as False for collections for now, since we no of know way in which a user could make non-ascii collection code (other than by editing a ZIP archive).

For migrations, I've added a new migration (0013) which alters the model field validations, removes the too-strict db constraints, and reinstates unicode-friendly db constraints. I've also edited existing migrations 0009 and 0010 for components and containers, respectively, to loosen their db constraints.  Here's my thinking:

* If a dev has already applied 0009 and 0010: Migration 0013 will remove the over-strict db constraints and add the unicode-friendly ones.
* Otherwise, for devs/ops who are skipping 0.43.0 and are applying 0009 and 0010 in the future (the vast majority): The altered 0009 and 0010 will apply unicode-friendly constraints, which won't cause an IntegrityError if there are non-ascii codes in the db. Then, 0013 will remove those constraints, and then just re-apply the same constraints again.

I'm not modifying the overly-strict Django validators in 0009 and 0010, because (I believe) Django will happily "add" and "remove" those validators regardless of what's in the database.

Bumps version to 0.44.0


src/openedx_content/migrations/0013_unicode_container_component_codes.py
```diff
@@ -0,0 +1,57 @@
...
+            field=openedx_django_lib.fields.MultiCollationCharField(db_collations={'mysql': 'utf8mb4_bin', 'sqlite': 'BINARY'}, max_length=255, validators=[django.core.validators.RegexValidator(re.compile('^[\\w.-]+\\Z'), 'Enter a valid "code name" consisting of any letters, numbers, underscores, hyphens, or periods.', 'invalid')]),
+        ),
+        migrations.AddConstraint(
+            model_name='component',
+            constraint=models.CheckConstraint(condition=django.db.models.lookups.Regex(models.F('component_code'), '^[\\w.-]+\\Z'), name='oel_component_code_regex', violation_error_message='Enter a valid "code name" consisting of any letters, numbers, underscores, hyphens, or periods.'),
+        ),
+        migrations.AddConstraint(
+            model_name='container',
+            constraint=models.CheckConstraint(condition=django.db.models.lookups.Regex(models.F('container_code'), '^[\\w.-]+\\Z'), name='oel_container_code_regex', violation_error_message='Enter a valid "code name" consisting of any letters, numbers, underscores, hyphens, or periods.'),
+        ),
```
bradenmacdonald, line 55: This updated constraint is already added in migration 0009, so why does it need to get re-created in this migration?

src/openedx_django_lib/fields.py
```diff
@@ -119,15 +119,21 @@ def immutable_uuid_field() -> models.UUIDField:
...
+    'Enter a valid "code name" consisting of latin letters (A-Z, a-z), numbers, underscores, hyphens, or periods.'
+)
+
+_CODE_VIOLATION_MSG_UNICODE = _(
+    'Enter a valid "code name" consisting of any letters, numbers, underscores, hyphens, or periods.'
 )
 
 
-def code_field(**kwargs) -> MultiCollationCharField:
+def code_field(unicode: bool, **kwargs) -> MultiCollationCharField:
```
bradenmacdonald, line 136: I like that this is mandatory.

src/openedx_content/migrations/0013_unicode_container_component_codes.py
kdmccormick: I tried to spell this out in the PR description. It's for those of us developers who already installed 0.43.0 into platform and applied 0009 and 0010, getting the ascii-only constraints.

I guess I could omit this, and instead just send out a slack blast to the handful of devs who might be in this situation, telling them to back-migrate to 0008. I think I'd also yank 0.43 from PyPI in that case.

kdmccormick: @bradenmacdonald let me know if you'd prefer this : 
> I guess I could omit this, and instead just send out a slack blast to the handful of devs who might be in this situation, telling them to back-migrate to 0008. I think I'd also yank 0.43 from PyPI in that case.

src/openedx_django_lib/fields.py
kdmccormick: same. I'd rather force developers to think about which kind of code they're defining.

kdmccormick: maybe we can make it default to true after doing https://github.com/openedx/openedx-platform/issues/38413

src/openedx_content/migrations/0013_unicode_container_component_codes.py
bradenmacdonald: Oops, sorry I didn't notice the PR description. Since your platform PR was not merged I expect you can count the number of affected developers on one hand. But it is a nice idea to resolve it automatically; if you're keeping it, maybe mention that in a comment in the migration itself too? Either way is good with me!

bradenmacdonald (approved):
Looks good and works well!

src/openedx_content/migrations/0013_unicode_container_component_codes.py
kdmccormick: great, I added comments!

kdmccormick:
@bradenmacdonald I added comments and a couple tests. I'll merge now, but if you have any suggestions on those I'm happy to incorporate them when I make the "Mark things stable / bump to v1.0" PR.

PR #559, opened 2026-04-23
First Stable Release: v1.0

kdmccormick:
This removes the big warning messages about this package being unstable. We're happy for developers to begin using these APIs, *except* for the ones with big red "UNSTABLE" warnings in their docstrings.

Part of:
* https://github.com/openedx/openedx-core/issues/434

Blocks:
* https://github.com/openedx/openedx-platform/pull/38435

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,10 +47,10 @@ We have a few different identifier types in the schema, and we try to avoid ``_i
 See Also
 ~~~~~~~~
 
-The structure of this repo follows [OEP-0049](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0049-django-app-patterns.html) where possible, and also borrows inspiration from:
+The structure of this repo follows `OEP-0049 <https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0049-django-app-patterns.html>`_ where possible, and also borrows inspiration from:
 
-* [Scaling Django to 500 apps](https://2021.djangocon.us/talks/scaling-django-to-500-apps/) (Dan Palmer, DjangoCon US 2021)
-* [Django structure for scale and longevity](https://www.youtube.com/watch?v=yG3ZdxBb1oo) (Radoslav Georgiev, EuroPython 2018)
+* `Scaling Django to 500 apps <https://2021.djangocon.us/talks/scaling-django-to-500-apps/>`_ (Dan Palmer, DjangoCon US 2021)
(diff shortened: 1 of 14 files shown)


src/openedx_content/migrations/0008_rename_collection_key_to_collection_code.py
```diff
@@ -15,7 +15,6 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('openedx_content', '0007_publishlogrecord_direct'),
-        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
```
kdmccormick, line 18: Just a bit of clean-up. Claude added these, they don't need to be here.

bradenmacdonald:
> Are there any function signatures or data classes we want to adjust now, before this goes out in Verawood?

* If we can, I would love to make some of these a bit more consistent: https://github.com/openedx/openedx-core/issues/549
* Most of [the collections API functions](https://github.com/openedx/openedx-core/blob/main/src/openedx_content/applets/collections/api.py) are missing a parameter for specifying which user performed the action - e.g. `update_collection`, `delete_collection`, `restore_collection`, `remove_from_collection`
* [Change `create_component_version_media`](https://github.com/openedx/openedx-core/issues/561) so ComponentVersion is properly immutable.
* (Edit: added): Make the first argument to our API methods consistently `Thing | Thing.ID`: https://github.com/openedx/openedx-core/pull/542#discussion_r3102456222

README.rst
```diff
@@ -11,7 +11,10 @@ Overview
 
 The ``openedx-core`` project holds Django apps which represent core teaching & learning platform concepts.
 
-Each app exposes stable, public API of Python functions and Django models. Some apps additionally provides REST APIs. These APIs are suitable for use in ``openedx-platform`` as well as in community-developed Open edX plugins.
+Each app exposes a public API of Python functions and Django models; some apps also provide REST APIs. These APIs are suitable for use in openedx-platform and in community-developed Open edX plugins.
+
+APIs marked "UNSTABLE" are subject to change at any time. All other APIs are considered stable, and any breaking changes will be announced through the community DEPR (deprecation and removal) process.
```
kdmccormick, line 16: There are none now, but I'll leave this note here as a strategy we can use as we develop new APIs.

ormsbee (approved):

bradenmacdonald (approved):
🚀

PR #564, opened 2026-04-27
feat: Allow IDs or Models to be passed to all publishing APIs

kdmccormick:
Part of: https://github.com/openedx/openedx-core/issues/562

To reduce noise in the implementations, I'm proposing these general patterns:

* Keep the param positional so that callers don't care if it's `model`, `model_id`, or `model_or_id`.
* When the function just needs the ID, name the param `model_id` and immediately invoke `model_id = get_model_id(model_id)`.
* When the function needs the model instance, name the param `model` and immediately invoke `model = get_<model>(model)`.

If this looks good, I'll apply it to the other applets as well.

BREAKING CHANGE: Some API functions which previously accepted `int` now accept `<Model>.ID`. Runtime behavior has not changed, but callers may need to use additional annotations or `typing.cast(...)` in order to get their typechecker to pass.


src/openedx_content/applets/publishing/api.py
```diff
@@ -74,10 +75,12 @@
 ]
 
 
-def get_learning_package(learning_package_id: LearningPackage.ID, /) -> LearningPackage:
+def get_learning_package(learning_package_id: LearningPackage | LearningPackage.ID, /) -> LearningPackage:
     """
     Get LearningPackage by ID.
     """
+    if isinstance(learning_package_id, LearningPackage):
+        return learning_package_id
```
kdmccormick, line 84: Yes, I know this is funny :P But it makes the rest of the code simpler.

src/openedx_content/applets/publishing/api.py
```diff
@@ -188,21 +192,22 @@ def create_publishable_entity(
...
-    entity_id: PublishableEntity.ID,
+    entity_id: PublishableEntity | PublishableEntity.ID,
     /,
     version_num: int,
     title: str,
     created: datetime,
     created_by: int | None,
     *,
-    dependencies: list[int] | None = None,  # PublishableEntity IDs
+    dependencies: Sequence[PublishableEntity.ID] | None = None,
```
kdmccormick, line 228: `list` is invariant (its items must be the exact type, not subclasses thereof), whereas `Sequence` is covariant, thus more likely to behave with mypy the way consumers expect.

kdmccormick:
@bradenmacdonald Could you review? If you like it, I'll do the other applets this way too.

bradenmacdonald:
@kdmccormick I like it!

Could you please add a test for the `get_model_id()` helper that uses `assert_type` to verify that it's returning a typed ID like `LearningPackageID` correctly? It is working perfectly already; I just want to ensure that it stays working and doesn't downgrade to returning `typing.All` in some future refactor.

bradenmacdonald:
@kdmccormick I'm curious why get_model_id() wasn't typechecking properly? I tried out your branch before and it seemed to be working fine.

kdmccormick:
@bradenmacdonald It passes mypy. The issue is that mypy seems to effectively consisder `get_model_id` to return `Any` rather than `<Any subclass of int>`. 

For example.

```python
# given:
my_model_or_id: Model | int

# these pass, as we'd expect
my_id: LearningPackage.ID = get_model_id(my_model_or_id)
my_id: int = get_model = get_model_id(my_model_or_id)

# but this passes too :(
my_id: str = get_model_id(my_model_or_id)
```

It seems that mypy doesn't respect `bound=int` on the return `TypeVar`.

You're welcome to take a whack at it--I'm going to put this down to focus on the timestamp & author stuff.

bradenmacdonald:
@kdmccormick 

The problem you're encountering is that you're defining models in [`test_django_app/models.py`](https://github.com/openedx/openedx-core/blob/main/tests/test_django_app/models.py) but Django-stubs is only auto-detecting models in our main project:

https://github.com/openedx/openedx-core/blob/c3d876d0bd93df02f0bd92fc93d5fc23f177602d/mypy.ini#L12-L13

So it's not properly typing all the relation and ID accessors on those particular models.

I think it's safe to change mypy to use `test_settings` as its Django settings, and then you'll find that your latest two revert commits cb7742eb65f9467c712f5b77c7de1b67a443a9e0 and fd4924db3f4a4448e6716be0dabb3c4eff8bc7c2 aren't necessary.

> \# given:
> `my_model_or_id: Model | int`
> It seems that mypy doesn't respect bound=int on the return TypeVar.

I think you mean given `my_model_or_id: Model | Model.ID` right?

It seems to be working fine for me on 8dc3240fef2798db3d623d74f2c01272b65f3d5f

```python
# tests/openedx_django_lib/test_typing.py

def more_tests(my_model_or_id: LearningPackage | LearningPackage.ID) -> None:
    # these pass, as we'd expect
    my_typed_id: LearningPackage.ID = get_model_id(my_model_or_id)
    my_int_id: int = get_model_id(my_model_or_id)

    # but this passes too :(
    my_str_id: str = get_model_id(my_model_or_id)

def more_tests2(my_model_or_id: MyTypedModel | MyTypedModel.ID) -> None:
    # these pass, as we'd expect
    my_typed_id: MyTypedModel.ID = get_model_id(my_model_or_id)
    my_int_id: int = get_model_id(my_model_or_id)

    # but this passes too :(
    my_str_id: str = get_model_id(my_model_or_id)

```

mypy:

```
tests/openedx_django_lib/test_typing.py:47: error: Incompatible types in assignment (expression has type "LearningPackageID", variable has type "str")  [assignment]
tests/openedx_django_lib/test_typing.py:55: error: Incompatible types in assignment (expression has type "MyTypedModelID", variable has type "str")  [assignment]
```

PR #565, opened 2026-04-28
Remove ability to add media to component versions after initialization

ormsbee:
This implements https://github.com/openedx/openedx-core/issues/561

```
Media associations must now be specified at the time when new
ComponentVersions are created:

* create_component_version
* create_component_and_version
* create_next_component_version

ComponentVersions are intended to be immutable, and having
create_component_version_media meant that we were encouraging a
pattern where people would create a component version first, and
then make modifications to it after the fact.

This bumps the version to 0.46.0.
```


src/openedx_content/applets/media/models.py
```diff
@@ -240,6 +243,16 @@ class Media(models.Model):
...
 
+    # Custom type for our primary key, to make it more type-safe when using in
+    # API calls.
+    MediaID = NewType("MediaID", int)
+    type ID = MediaID
+
+    class IDField(TypedBigAutoField[ID]):  # Boilerplate for fully-typed ID field.
+        pass
+
+    id = IDField(primary_key=True)
```
ormsbee, line 253: @bradenmacdonald: May I get a sanity check from you on this one? I think I understand what this is doing and why, but I'm mostly cargo-culting it from PublishableEntity. This convention should probably go in an ADR somewhere.

bradenmacdonald: @ormsbee Yep, looks right. Nice comment in the migration too.


The documentation such as it is is here:

https://github.com/openedx/openedx-core/blob/c22cc1bd9571faf925a90a9f004575d291d2533f/src/openedx_django_lib/id_fields.py#L9-L25

I believe we might update OEP-68 to include examples of how to do this, after the Verawood cut? CC @kdmccormick

kdmccormick: yes! https://github.com/openedx/openedx-core/issues/538

ormsbee:
I'll have the companion platform PR up shortly...

src/openedx_content/applets/components/api.py
bradenmacdonald, line 135: Nit: might be nice to explain the role of `media` in the docstring here.

src/openedx_content/applets/components/api.py
```diff
@@ -300,8 +286,76 @@ def create_component_and_version(  # pylint: disable=too-many-positional-argumen
...
+):
+    """
+    Internal helper to set the Media for this ComponentVersion.
+
+    Only call this when we're first initializing a ComponentVersion.
+
+    Media can be specified as ``bytes`` for testing convenience, but you will
+    almost always want to create a Media object first in actual app code,
+    because that will give you better control over the MIME type and storage
+    specifics (file vs. database).
```
bradenmacdonald, line 325: Should this note about `bytes` be in the public API methods too? (`create_[next_]component_version`)

src/openedx_content/applets/components/api.py
```diff
@@ -300,8 +286,76 @@ def create_component_and_version(  # pylint: disable=too-many-positional-argumen
...
+    # We allow a range of values to be in paths_to_media_values, but we want to
+    # normalize to media_ids for our bulk insert later.
+    paths_to_media_ids: dict[str, Media.ID] = {}
+
+    for path, media_value in paths_to_media_values.items():
+        match media_value:
+            case int():  # MediaID
+                media_id = media_value
+            case Media():
+                media_id = media_value.id
```
bradenmacdonald, line 354: In this case (getting a `Media` instance), we should validate that `Media.learning_package_id` matches `version.component.learning_package.id`. This was flagged in https://github.com/openedx/openedx-core/issues/463 .

src/openedx_content/applets/components/api.py
```diff
@@ -300,8 +286,76 @@ def create_component_and_version(  # pylint: disable=too-many-positional-argumen
...
+                )
+                media_id = media.id
+            case _:
+                raise ValueError(f"Invalid object for paths_to_media Media: {media_value!r}")
+
+        # Don't allow whitespace, absolute paths, or Windows-style paths
+        normalized_path = path.strip().replace('\\', '/').lstrip('/')
+        paths_to_media_ids[normalized_path] = media_id
+
+    ComponentVersionMedia.objects.bulk_create(
```
bradenmacdonald, line 381: Nice, using `bulk_create` here will improve performance for anything with multiple pieces of media. Not sure we have much of that yet though.

src/openedx_content/applets/components/api.py
```diff
@@ -300,8 +286,76 @@ def create_component_and_version(  # pylint: disable=too-many-positional-argumen
...
+                    media_type.id,
+                    data=media_value,
+                    created=created,
+                )
+                media_id = media.id
+            case _:
+                raise ValueError(f"Invalid object for paths_to_media Media: {media_value!r}")
+
+        # Don't allow whitespace, absolute paths, or Windows-style paths
+        normalized_path = path.strip().replace('\\', '/').lstrip('/')
```
bradenmacdonald: Are we concerned that passing `media={'foo': A, '/foo': B}` silently keeps `B` because paths can be normalized to the same?

Also, Claude found this one:

> `previous_media` | `media_to_replace` is keyed on the raw, un-normalized path. Normalization happens later inside the helper. So a caller who passes `media_to_replace={'/foo.txt': None}` to delete a previously-stored `/foo.txt` will not delete it — the `None` entry is filtered out by the comprehension before normalization, leaving the original `foo.txt` intact. Either normalize keys before the merge, or document that media_to_replace keys must already be in normalized form.

(or throw an error instead of quietly normalizing)

bradenmacdonald (approved):
Nice! Just some suggestions but nothing blocking. I haven't tested this; let me know if you want me to.

src/openedx_content/applets/components/api.py
ormsbee: So the first item you mention did occur to me, but I didn't think it was likely. But I guess I should do the normalization up front. This is mostly a product of me thinking of it last, I think.

ormsbee:
Comments addressed in https://github.com/openedx/openedx-core/pull/565/changes/cdab85db38e38b114898513e02a89b3a9c74041e. I'm going to squash+merge this and finish the platform PR. I ended up making the path checking stricter.

PR #566, opened 2026-04-28
feat!: Make Timestamp, Author params consistent

kdmccormick:
## Description

This PR aims to unify handling of Timestamp and Author parameters across all API functions. It ensures that the default behavior is "figure out a reasonable Timestamp and Author" rather than "assume that they are `None`". Timestamps are forbidden for being `None`, and Authors can only be `None` if explicitly specified as such using the special `AUTHOR_NONE` sentinel.

BREAKING CHANGE: Various Timestamp params (named `<verb>ed`, `<verb>ed_at`) and Author params (named `<verb>ed_by`) have stricter rules. For example, a concrete `User` or explicit `AUTHOR_NONE` must always be specified outside of a `bulk_draft_changes_for`context. See `resolve_timestamp` and `resolve_author` functions for details. I will make this breaking change message more specific before merging.

## Background

See:
* https://github.com/openedx/openedx-core/issues/549

## Concerns

I haven't mapped it out exactly, but I believe this will add additional queries to several functions, where we need to get the active DraftChangeLogContext (for a given learning_package_id) in order to resolve TIMESTAMP_AUTO and AUTHOR_AUTO. Additionally, functions which only take an entity_id (or version_id, etc.) will need to perform an extra extra query in order to get the operation's learning_package_id.

## Status

I'm just looking for feedback on the approach currently.

Done:

* Implemented POC, just for src/openedx_content/applets/publishing.
* mypy passes in src/openedx_content/applets/publishing.

Todo:

* Unit tests
* Implement it in the rest of the applets.
* Manual tests
* Write up the BREAKING CHANGEs in more detail
* openedx-platform PR

## AI

This code is all written by hand, but I used Claude Opus 4.7 to help me plan it. In particular, it suggested that:

> **Log-only draft APIs raise inside a context.** `set_draft_version`, `soft_delete_draft`, and `reset_drafts_to_published` raise `ValueError` if called inside a `bulk_draft_changes_for` block with a non-`AUTO_*` author or timestamp, because the underlying `DraftChangeLog` records those fields at the log level (not per record). Outside a context the explicit values are honored.

which led me to add the `can_override_context` parameters.

### DIFF
--- a/src/openedx_content/applets/publishing/api.py
+++ b/src/openedx_content/applets/publishing/api.py
@@ -1,5 +1,5 @@
 """
-Publishing API (warning: UNSTABLE, in progress API)
+Publishing API
 
 Please look at the models.py file for more information about the kinds of data
 are stored in this app.
--- a/src/openedx_content/applets/publishing/api.py
+++ b/src/openedx_content/applets/publishing/api.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 from contextlib import nullcontext
 from datetime import datetime, timezone
-from typing import ContextManager, Optional, cast
+from typing import ContextManager, Optional, cast, TypeAlias, Literal
 
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
@@ -43,6 +43,12 @@ from .models.publish_log import Published
 # start with an underscore AND it is not in __all__, that function is considered
 # to be callable only by other applets in the openedx_content package.
 __all__ = [
+    "Timestamp",
+    "TIMESTAMP_AUTO",
+    "UserID",
+    "AUTHOR_AUTO",
+    "AUTHOR_NONE",
+    "Author",
     "get_learning_package",
     "get_learning_package_by_ref",
     "create_learning_package",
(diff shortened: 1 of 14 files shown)


kdmccormick:
@ormsbee @bradenmacdonald Let me know what you think about this approach, and whether you feel confident enough in it that I should rush it in for Verawood or wait until post-1.0.

src/openedx_content/applets/publishing/api.py
```diff
@@ -74,6 +80,65 @@
...
+"""
+
+
+UserID: TypeAlias = int
+"""
+A UserID is an integer.
+
+A UserID is meant to hold a primary key of a User object, but this is not
+enforced by the type checker. Future versions of openedx-core may enforce
+this more strictly by making UserID a NewType (subclass) of int.
```
bradenmacdonald, line 110: In the future, we could add a `UserID` type here, and/or move these types to openedx-core: https://github.com/openedx/openedx-platform/blob/master/openedx/core/types/user.py

I suspect that without modifying Django itself, aliasing `UserID = int` is the best we can do.

src/openedx_content/applets/publishing/api.py
```diff
@@ -74,6 +80,65 @@
...
+reasonable default for standard application API usage--consider AUTO_AUTHOR
+instead.
+"""
+
+
+AUTHOR_AUTO: Literal["AUTHOR_AUTO"] = "AUTHOR_AUTO"
+"""
+Attribute the operation to the same user as the enclosing `bulk_draft_changes_for` context.
+
+Outside of a context, using `AUTHOR_AUTO` is meaningless and will result in a `ValueError`.
```
bradenmacdonald, line 129: We _could_ use djngo-crum or something to try to auto-detect the user, but I think it's better to just explicitly pass it in.

src/openedx_content/applets/publishing/api.py
```diff
@@ -1634,25 +1806,33 @@ def bulk_draft_changes_for(
...
     is identical to this::
 
-        with bulk_draft_changes_for(component.learning_package.id):
+        with bulk_draft_changes_for(component.learning_package.id, changed_by=request.user.id):
             update_one_component(component.learning_package.id, component)
     """
+    if changed_at is TIMESTAMP_AUTO:
+        changed_at_dt = datetime.now(tz=timezone.utc)
+    else:
+        changed_at_dt = cast(datetime, changed_at)
```
bradenmacdonald, line 1827: Do we need this cast? If `changed_at` is a `Timestamp` and it's not `TIMESTAMP_AUTO` then it must be a `datetime` already right?

kdmccormick: Yeah, but I couldn't get mypy to believe it.

bradenmacdonald:
> I believe this will add additional queries to several functions, where we need to get the active DraftChangeLogContext (for a given learning_package_id) in order to resolve TIMESTAMP_AUTO and AUTHOR_AUTO

Getting the active `DraftChangeLogContext` doesn't use any queries, AFAIK.

> Additionally, functions which only take an entity_id (or version_id, etc.) will need to perform an extra extra query in order to get the operation's learning_package_id.

Yes, that would be a an extra query.

src/openedx_content/applets/publishing/api.py
```diff
@@ -162,13 +227,101 @@ def learning_package_exists(package_ref: str) -> bool:
     return LearningPackage.objects.filter(package_ref=package_ref).exists()
 
 
+def resolve_timestamp(
```
kdmccormick, line 230: ensure to add test data where `datetime is None`

src/openedx_content/applets/publishing/api.py
```diff
@@ -76,6 +80,65 @@
...
+
+Within a `bulk_draft_changes_for` context, `TIMESTAMP_AUTO` indicates to use the
+context's timestamp.  Outside of `bulk_draft_changes_for`, `TIMESTAMP_AUTO` indicates
+to the current time (`datetime.now`).
+
+`TIMESTAMP_AUTO` is a reasonable default for most situations.
+"""
+
+
+type Timestamp = datetime | Literal["TIMESTAMP_AUTO"]
```
ormsbee, line 95: Maybe `DATETIME_AUTO`? In Python, "timestamp" specifically means a Unix float timestamp, i.e. seconds since the epoch. There's even a `timestamp()` method on `datetime` objects.

ormsbee: Or Created/CREATED_AUTO, to parallel Author/AUTHOR_AUTO?

kdmccormick: I like `DATETIME_AUTO`, because the other params aren't always "created"; they're "changed", "set", etc.

Then we can have `type DatetimeOrAuto = datetime | DATETIME_AUTO`

kdmccormick:
Closed in favor of https://github.com/openedx/openedx-core/pull/573 (WIP)

PR #568, opened 2026-04-28
feat!: remove olx_importer app

ormsbee:
This was never published as part of the openedx-core package, and
existed as an early way to bring XBlock component data into our models
for testing purposes. Now that backup_restore exists and we have
alternate ways to import content via the UX as well (from v1 libraries
and courses), there's no point in keeping olx_importer around and up to
date.


kdmccormick (approved):
Thanks!

PR #571, opened 2026-04-29
Delete tag updates meilisearch

mgwozdz-unicon:
## Description
This PR in conjunction with the openedx-platform PR (https://github.com/openedx/openedx-platform/pull/38477) is the backend component for https://github.com/openedx/modular-learning/issues/260. This PR ensures that when a tag and its descendants are deleted that CONTENT_OBJECT_ASSOCIATIONS_CHANGED events take place for each associated object with no duplicates. The openedx-platform PR ensures that once those events happen, the entire search document is replaced for the relevant content.

*Goal*:
When a Course Author deletes a tag that is associated to Library content, they should be able to return to the Library page and not see those tags in the tag search filter.

## Supporting information
Github Issue: https://github.com/openedx/modular-learning/issues/260

## Testing instructions
Note: This cannot be tested without the code from the following PRs in place:
- https://github.com/openedx/frontend-app-authoring/pull/3035
- https://github.com/openedx/openedx-platform/pull/38477
1. Go to Taxonomies and select a taxonomy
2. Add new tags to this taxonomy that you plan to delete. (I have been adding a subtag (e.g. Wind1) with 2 sibbling children (W1.1 and W1.2) for my testing.)
3. Go to a Library and create 2 new pieces of content.
4. For one piece of content, associate the parent subtag (e.g. Wind1) and click Save.
5. For the other piece of content, associate a child (e.g. W1.2) and click Save.
6. Click on the tag search filter button and note that Wind1 and W1.2 appear in the tag search filter.
7. Return to the taxonomy and delete the subtag Wind1.
8. It should disappear from the taxonomy and a toast should display indicating what was deleted.
9. Go back to the Library and refresh the page in the browser.
10. The pieces of content you previously associated those tags to should now display that 0 tags are associated to them.
11. Click on the tag search filter button and note that Wind1 and W1.2 no longer appear in the tag search filter.

## Deadline

This is for our Verawood code cut stretch goal.

## Other information
While this code requires other PRs in place to be able to test, I believe it can be safely merged without them.

### DIFF
--- a/src/openedx_tagging/signal_handlers.py
+++ b/src/openedx_tagging/signal_handlers.py
@@ -34,7 +34,8 @@ def _is_explicit_tag_delete(
     if isinstance(origin, Tag):
         return origin.pk == instance.pk
 
-    # No-op if the origin isn't a queryset of tags, since that would be unexpected and we don't want to risk emitting too many events.
+    # No-op if the origin isn't a queryset of tags, since that would be
+    # unexpected and we don't want to risk emitting too many events.
     if not isinstance(origin, QuerySet) or origin.model is not Tag:
         return False
 
--- a/src/openedx_tagging/signal_handlers.py
+++ b/src/openedx_tagging/signal_handlers.py
@@ -3,11 +3,40 @@
 from functools import partial
 
 from django.db import transaction
-from django.db.models.signals import post_save
+from django.db.models import QuerySet
+from django.db.models.signals import post_save, pre_delete
 from django.dispatch import receiver
 
-from openedx_tagging.models.base import Tag
-from openedx_tagging.tasks import emit_content_object_associations_changed_for_tag_task
+from openedx_tagging.models.base import ObjectTag, Tag
+from openedx_tagging.tasks import (
+    emit_content_object_associations_changed_for_object_ids_task,
+    emit_content_object_associations_changed_for_tag_task,
+)
+
+
+def _is_explicit_tag_delete(instance: Tag, origin: object, using: str | None) -> bool:
+    """
+    Return True only for tags explicitly targeted by the delete operation.
+
+    Descendants deleted via CASCADE are skipped here because the explicit root
+    tag's handler emits updates for the whole subtree.
+    """
+    if isinstance(origin, Tag):
+        return origin.pk == instance.pk
+
+    if not isinstance(origin, QuerySet) or origin.model is not Tag:
+        return False
+
+    explicit_tags = origin.using(using)
+    if not explicit_tags.filter(pk=instance.pk).exists():
+        return False
+
+    lineage_parts = instance.lineage.rstrip("\t").split("\t")
+    ancestor_lineages = ["\t".join(lineage_parts[:index]) + "\t" for index in range(1, len(lineage_parts))]
+    if not ancestor_lineages:
+        return True
+
+    return not explicit_tags.filter(lineage__in=ancestor_lineages).exists()
 
 
 @receiver(post_save, sender=Tag)
@@ -28,5 +57,38 @@ def tag_post_save(sender, **kwargs):  # pylint: disable=unused-argument
         partial(
             emit_content_object_associations_changed_for_tag_task.delay,
             tag_id=tag_id
-        )
+        ),
+    )
+
+
+@receiver(pre_delete, sender=Tag)
+def tag_pre_delete(sender, **kwargs):  # pylint: disable=unused-argument
+    """
+    If a tag is deleted, enqueue async event emission for all associated objects.
+    """
+    instance = kwargs.get("instance", None)
+    origin = kwargs.get("origin", None)
+    using = kwargs.get("using", None)
+
+    if instance is None or instance.id is None:
+        return
+
+    if not _is_explicit_tag_delete(instance, origin, using):
+        return
+
+    object_ids = list(
+        ObjectTag.objects.using(using)
+        .filter(tag__lineage__startswith=instance.lineage)
+        .values_list("object_id", flat=True)
+        .distinct()
+    )
+    if not object_ids:
+        return
+
+    transaction.on_commit(
+        partial(
+            emit_content_object_associations_changed_for_object_ids_task.delay,
+            object_ids=object_ids,
+        ),
+        using=using,
     )
--- a/src/openedx_tagging/signal_handlers.py
+++ b/src/openedx_tagging/signal_handlers.py
@@ -14,24 +14,41 @@ from openedx_tagging.tasks import (
 )
 
 
-def _is_explicit_tag_delete(instance: Tag, origin: object, using: str | None) -> bool:
+def _is_explicit_tag_delete(
+    instance: Tag,
+    origin: Tag | QuerySet[Tag] | None,
+    using: str | None,
+) -> bool:
     """
     Return True only for tags explicitly targeted by the delete operation.
 
     Descendants deleted via CASCADE are skipped here because the explicit root
     tag's handler emits updates for the whole subtree.
+
+    Args:
+        instance: The Tag being deleted.
+        origin: The source of the delete operation - either a Tag instance (for instance.delete())
+                or a QuerySet[Tag] (for queryset.delete()), or None for other origins.
+        using: The database alias to use for queries, passed from the Django signal.
     """
     if isinstance(origin, Tag):
         return origin.pk == instance.pk
 
+    # No-op if the origin isn't a queryset of tags, since that would be unexpected and we don't want to risk emitting too many events.
     if not isinstance(origin, QuerySet) or origin.model is not Tag:
         return False
 
+    # Check if this instance is in the set of explicitly-targeted tags. If not, it's being deleted
+    # as a CASCADE side-effect, so it's not explicit.
     explicit_tags = origin.using(using)
     if not explicit_tags.filter(pk=instance.pk).exists():
         return False
 
-    lineage_parts = instance.lineage.rstrip("\t").split("\t")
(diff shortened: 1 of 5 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

src/openedx_tagging/signal_handlers.py
```diff
@@ -3,11 +3,40 @@
...
-from openedx_tagging.models.base import Tag
-from openedx_tagging.tasks import emit_content_object_associations_changed_for_tag_task
+from openedx_tagging.models.base import ObjectTag, Tag
+from openedx_tagging.tasks import (
+    emit_content_object_associations_changed_for_object_ids_task,
+    emit_content_object_associations_changed_for_tag_task,
+)
+
+
+def _is_explicit_tag_delete(instance: Tag, origin: object, using: str | None) -> bool:
```
ormsbee: If we expect `origin` to possibly be different classes, please list those classes explicitly, e.g. `origin: Tag | QuerySet[Tag]`, instead of annotating it as `object`.

src/openedx_tagging/signal_handlers.py
```diff
@@ -3,11 +3,40 @@
...
+    """
+    Return True only for tags explicitly targeted by the delete operation.
+
+    Descendants deleted via CASCADE are skipped here because the explicit root
+    tag's handler emits updates for the whole subtree.
+    """
+    if isinstance(origin, Tag):
+        return origin.pk == instance.pk
+
+    if not isinstance(origin, QuerySet) or origin.model is not Tag:
```
ormsbee: Is this a runtime error, i.e. should we be raising a TypeError here?

src/openedx_tagging/signal_handlers.py
ormsbee: Also, please document what these args represent in the docstring.

src/openedx_tagging/signal_handlers.py
```diff
@@ -3,11 +3,40 @@
...
+        return False
+
+    explicit_tags = origin.using(using)
+    if not explicit_tags.filter(pk=instance.pk).exists():
+        return False
+
+    lineage_parts = instance.lineage.rstrip("\t").split("\t")
```
ormsbee: Shouldn't this use `get_lineage()`?

src/openedx_tagging/signal_handlers.py
```diff
@@ -3,11 +3,40 @@
...
+    ancestor_lineages = ["\t".join(lineage_parts[:index]) + "\t" for index in range(1, len(lineage_parts))]
```
ormsbee, line 55: Please add a comment as to what this is doing.

src/openedx_tagging/signal_handlers.py
```diff
@@ -28,5 +57,38 @@ def tag_post_save(sender, **kwargs):  # pylint: disable=unused-argument
...
+def tag_pre_delete(sender, **kwargs):  # pylint: disable=unused-argument
+    """
+    If a tag is deleted, enqueue async event emission for all associated objects.
+    """
+    instance = kwargs.get("instance", None)
+    origin = kwargs.get("origin", None)
+    using = kwargs.get("using", None)
+
+    if instance is None or instance.id is None:
+        return
```
ormsbee, line 96: Please add a small comment here about how we would get here.

src/openedx_tagging/tasks.py
```diff
@@ -11,16 +12,18 @@
...
+    Emit CONTENT_OBJECT_ASSOCIATIONS_CHANGED once for each distinct object ID.
     """
-    object_ids = ObjectTag.objects.filter(tag=tag).values_list("object_id", flat=True)
     emitted_events = 0
+    seen_object_ids: set[str] = set()
+
+    for object_id in object_ids:
+        if object_id in seen_object_ids:
+            continue
+        seen_object_ids.add(object_id)
```
ormsbee: [Nit (optional)]: If the ordering is not important, you can also do this to emit each thing only once:


```suggestion
    for object_id in set(object_ids):
```

ormsbee (dismissed):
Handful of questions/requests. Thank you.

src/openedx_tagging/signal_handlers.py
mgwozdz-unicon: I don't think so. I think we want this to just be a no-op so that an unexpected type doesn't result in aborting or rolling back the delete operation. Let me know if you disagree.

tests/openedx_tagging/test_models.py
```diff
@@ -1114,6 +1117,70 @@ def test_rename_updates_search_index(self, mock_task_delay) -> None:
...
+        alice_object_id = "content-v1:org+course+run+type@unit+block@126"
+        delta_object_id = "content-v1:org+course+run+type@unit+block@127"
+        ObjectTag.objects.create(
+            object_id=alice_object_id,
+            taxonomy=self.alice.taxonomy,
+            tag=self.alice,
+        )
+        ObjectTag.objects.create(
+            object_id=delta_object_id,
+            taxonomy=self.delta.taxonomy,
```
bradenmacdonald, line 1175: It is confusing when reading this test in isolation, because the tags `self.alice` and `self.delta` appear to be coming from two difference taxonomies (`self.alice.taxonomy` vs. `self.delta.taxonomy`). I don't think this test belongs in `TestTagLineage` at all. The 4 or 5 tests related to event emitting should be in a separate `test_events.py` module.

I won't block the PR on this but I'd strongly prefer to see it at least flagged as a TODO if not moved before merge.

tests/openedx_tagging/test_models.py
```diff
@@ -1114,6 +1117,70 @@ def test_rename_updates_search_index(self, mock_task_delay) -> None:
...
+            object_id=object_id,
+            taxonomy=self.bob.taxonomy,
+            tag=self.bob,
+        )
+
+        with self.captureOnCommitCallbacks(execute=True):
+            self.bob.delete()
+
+        assert mock_task_delay.call_count == 1
+        assert mock_task_delay.call_args[1]["object_ids"] == [object_id]
```
bradenmacdonald, line 1159: Note: it seems odd that deleting a `Tag` emits an event here in this repo, but deleting an `ObjectTag` emits an event [here in openedx-platform](https://github.com/openedx/openedx-platform/blob/f20a6667f7c75daad691b77be3a12b7843167c1f/openedx/core/djangoapps/content_tagging/api.py#L413-L421). It would be good to note that here in the test case because that was the first question that came to my mind when seeing these tests - why are we testing deleting `Tag` but not `ObjectTag`, when both require a search index update?

I've updated https://github.com/openedx/openedx-core/issues/557 to mention this.

tests/openedx_tagging/test_models.py
```diff
@@ -1114,6 +1117,70 @@ def test_rename_updates_search_index(self, mock_task_delay) -> None:
...
+        """
+        Deleting a tag should enqueue an async task that emits
+        CONTENT_OBJECT_ASSOCIATIONS_CHANGED events for affected objects.
+        """
+        object_id = "content-v1:org+course+run+type@unit+block@125"
+        ObjectTag.objects.create(
+            object_id=object_id,
+            taxonomy=self.bob.taxonomy,
+            tag=self.bob,
+        )
```
bradenmacdonald: Nit: it's preferred to use `api.tag_object(...)` rather than directly calling `ObjectTag.objects.create(...)`

Unfortunately python doesn't give us any way to flag this in the code.

src/openedx_tagging/signal_handlers.py
```diff
@@ -3,11 +3,58 @@
...
+def _is_explicit_tag_delete(
+    instance: Tag,
+    origin: Tag | QuerySet[Tag] | None,
+    using: str | None,
+) -> bool:
+    """
+    Return True only for tags explicitly targeted by the delete operation.
+
+    Descendants deleted via CASCADE are skipped here because the explicit root
+    tag's handler emits updates for the whole subtree.
```
bradenmacdonald, line 26: Just curious, is there any reason this approach (using `_is_explicit_tag_delete`, manually emitting events for the descendants, and then suppressing events for the CASCADE) is preferable to just listening for each individual `Tag` deletion and emitting an event for each one, including the CASCADE deletes (but never manually emitting events for the descendants)?

The latter seems like it would have much simpler code since you wouldn't need this `_is_explicit_tag_delete` logic. But perhaps it's not as performant?

src/openedx_tagging/signal_handlers.py
ormsbee: If there is no known reason to have gotten into this state, please raise a TypeError here. As far as I can see, this is only being invoked from a place where you're specifically listening to Tag as the sender, so if it's anything other than these two, something is really broken, and it's better to break loudly so that the issue can be found more quickly.

src/openedx_tagging/signal_handlers.py
mgwozdz-unicon: It's to prevent duplicate events per object so that it emits one event per object regardless of how many tags in the subtree are applied to it. This does improve performance with N cascade deletes --> N tasks --> N sets of DB queries in the simple approach, versus 1 task with 1 query in the current approach. For large taxonomies this could be significant.

bradenmacdonald: That makes sense, and would be worth mentioning in the code.

Even with that optimization, the number of events emitted could be huge, but since they're in a celery task it should be OK.

bradenmacdonald (approved):

PR #575, opened 2026-04-30
fix: Handle Taxonomy CASCADE deletes in _is_explicit_tag_delete

kdmccormick:
## Merge info

For [Verawood](https://github.com/openedx/openedx-platform/pull/38435), we need to merge either this PR, or just the straightforward revert: https://github.com/openedx/openedx-core/pull/577. Both are ready to go.

EDIT: https://github.com/openedx/openedx-core/pull/576 is also a potential fix.

## Description

When a Taxonomy is deleted, Django fires pre_delete for each related Tag with origin set to the Taxonomy instance. The previous code raised a TypeError for any non-Tag, non-QuerySet origin, breaking any consumer that deletes a Taxonomy.

For a non-Tag-queryset origin (e.g. taxonomy.delete() or Taxonomy.objects.filter(...).delete()), we now emit only for root-level tags; their handler covers the whole subtree via lineage__startswith.

Bumps to v1.0.1.

## Supporting information

The original code was added in https://github.com/openedx/openedx-core/pull/571

while [upgrading to v1.0](https://github.com/openedx/openedx-platform/pull/38435), openedx-platform unit tests caught this:
https://gist.github.com/kdmccormick/912e71d4a57bf8aed046e9be3ca297e7

## AI

Generated by Claude Sonnet 4.6 with my prompt:

> (We're working within the src/openedx_tagging directory, no need to look at other subdirs of src)
> In 508be7d8a26b0ef6cb24b4e15a8b2a0244ec0f89, test_is_explicit_tag_delete_raises_for_unexpected_origin_type checks that a Taxonomy cannot be passed as the origin to _is_explicit_tag_delete. But this same behavior is causing failures in external consumes of this package, because deleting a Taxonomy causes a CASCADE which delete tags, thus passing Taxonomy as the origin to _is_explicit_tag_delete. This seems like a bug in openedx_tagging, no? Can you discern why the commit author made the choice to disallow this scenario?

which it got right on the first try. I just had a couple follow-ups to tweak its coding style.

I then had Opus 4.7 audit the change, and it suggested the additional test_taxonomy_delete_updates_search_index test.

I reviewed all code myself.

### DIFF
--- a/src/openedx_core/__init__.py
+++ b/src/openedx_core/__init__.py
@@ -6,4 +6,4 @@ The public APIs belong to the specific apps (openedx_content, openedx_tagging, e
 """
 
 # The version for the entire repository
-__version__ = "1.0.0"
+__version__ = "1.0.1"
--- a/src/openedx_tagging/signal_handlers.py
+++ b/src/openedx_tagging/signal_handlers.py
@@ -16,30 +16,33 @@ from openedx_tagging.tasks import (
 
 def _is_explicit_tag_delete(
     instance: Tag,
-    origin: Tag | QuerySet[Tag] | None,
+    origin: object,
     using: str | None,
 ) -> bool:
     """
     Return True only for tags explicitly targeted by the delete operation.
 
     Descendants deleted via CASCADE are skipped here because the explicit root
-    tag's handler emits updates for the whole subtree.
+    tag's handler emits updates for the whole subtree via lineage__startswith.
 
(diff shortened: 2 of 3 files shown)


ormsbee (dismissed):
This seems fine to me, and I would prefer to fix forward.

bradenmacdonald:
Nice fix - thanks @kdmccormick !

PR #577, opened 2026-04-30
revert: emit events so that deleting tag(s) updates studio search index

kdmccormick:
## Merge info

For [Verawood](https://github.com/openedx/openedx-platform/pull/38435), we need to merge either this PR, or the fix-forward PR: https://github.com/openedx/openedx-core/pull/575. Both are ready to go.

## Description

When a tag is deleted by CASCADE of another model (for example, deleting a Taxonomy), it causes a TypeError in _is_explicit_tag_delete, which expects the event origin to always be a Tag or QuerySet[Tag].

Bumps to 1.0.1

This reverts commit 508be7d8a26b0ef6cb24b4e15a8b2a0244ec0f89.

## Supporting

See https://github.com/openedx/openedx-core/pull/575 for more details and a proposed fix.

## AI

I used Claude Sonnet 4.6 in https://github.com/openedx/openedx-core/pull/575 to confirm that there was indeed a bug, but I did the revert by hand.

### DIFF
--- a/src/openedx_core/__init__.py
+++ b/src/openedx_core/__init__.py
@@ -6,4 +6,4 @@ The public APIs belong to the specific apps (openedx_content, openedx_tagging, e
 """
 
 # The version for the entire repository
-__version__ = "1.0.0"
+__version__ = "1.0.1"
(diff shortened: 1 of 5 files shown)


kdmccormick:
Closed in favor of https://github.com/openedx/openedx-core/pull/575

PR #582, opened 2026-05-04
docs: Note that 1.0.x is reserved for Verawood

kdmccormick:


bradenmacdonald (approved):
Sounds good. Though if we add any breaking changes and not just features, we're supposed to jump to 2.0 right?

src/openedx_core/__init__.py
```diff
@@ -6,4 +6,7 @@
 """
 
 # The version for the entire repository
+# DEVELOPERS PLEASE NOTE:
+# - 1.0.x is reserved for `verawood-backports`
+# - Use 1.1.0 as the next `main` release (and then delete this comment).
```
kdmccormick:
```suggestion
# - Use 1.1.0 (if non-breaking) or 2.0.0 (if breaking) as the next `main` release,
#   and then delete this comment.
```

kdmccormick:
@bradenmacdonald yep, good point, noted

PR #583, opened 2026-05-05
Update models docstrings to follow proper RST conventions

ormsbee:
This reformats a lot of our docstrings to follow RST practices properly, like real links to classes and models, fixing some Markdown-isms, reformatting broken definition lists, etc. The thing that makes me twitch the most is reformatting large block-comments over variable declarations to be docstring-like comments under them.

This doesn't fix everything, but it rehabilitates the docs to a place that is much better than where we started. This did not add any new content to the docs—it's just reformatting and some rephrasing on my part as part of some manual edits. I've deleted a few outdated TODO items, but I've otherwise refrained from trying to do any additional editing to the docstrings for now, in the hopes of landing this more easily.

To test locally, run `make docs` at the root. You may need to install the `requirements/doc.txt` requirements.

The Claude prompt:

> The openedx_content app has extensive docstrings in its models files, but those doscstrings often do not conform to good RST conventions. Sometimes Markdown formatting is used instead of RST. Class names and function names are not properly linked. Some annotations may be incorrect or inconsistently applied. Definition lists are sometimes spaced wrong, so they end up looking like blockquotes. Definition lists are sometimes defined with (illegal) multi-line terms, and this should be reformatted. Model fields are sometimes commented extensively in comments that go above the field declarations, instead of docstring comments below the fields (which would get picked up by Sphinx). State transitions are sometimes written as "->" instead of "→", leading to a poorer reading experience in rendered HTML documentation.
>
> Create a plan to fix these issues, and any other formatting errors you find. Reflow lines as necessary to keep within an 80 character line limit (if possible). Do not change the actual content of these comments.

### DIFF
--- a/src/openedx_content/applets/collections/models.py
+++ b/src/openedx_content/applets/collections/models.py
@@ -2,67 +2,73 @@
 Core models for Collections
 
 TLDR Guidelines:
+
 1. DO NOT modify these models to store full version snapshots.
 2. DO NOT use these models to try to reconstruct historical versions of
    Collections for fast querying.
 
-If you're trying to do either of these things, you probably want a new model or
-app. For more details, read on.
+If you're trying to do either of these things, you probably want a new model
+or app. For more details, read on.
 
 The goal of these models is to provide a lightweight method of organizing
-PublishableEntities. The first use case for this is modeling the structure of a
-v1 Content Library within a LearningPackage. This is what we'll use the
-Collection model for.
+:class:`PublishableEntity` objects. The first use case for this is modeling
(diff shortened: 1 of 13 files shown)


ormsbee:
@sarina, @kdmccormick: This PR is ready for review. I had to make a few minor edits after Claude went through it, and I modified some comments to remove outdated TODOs. Otherwise, it's all just reformatting as described in the prompt.

kdmccormick (approved):
Thanks!

sarina (approved):
❤️

PR #593, opened 2026-05-08
Events refinements: 'restored' flag on entity draft change, don't emit collection change when draft entity deleted.

bradenmacdonald:
This PR improves events in openedx-core.

It does these two things that I think we definitely want:

1. Updates the `ENTITIES_DRAFT_CHANGED` event, to add a `restored` boolean field. This allows event handlers to distinguish between "newly created" draft entities and "un-deleted" draft entities, which was previously not possible with only the information in the event itself. **This is a change to the event data payload**, but it's adding a field and the event is brand new and not yet published/used anywhere. I think we do want this, although it does add an extra query to the event generation code.
2. Adds extra validation to prohibit adding entities to a deleted collection. If you're trying to do that, it's likely a mistake/bug.

And it does this thing that I'm a bit on the fence about:

3. Changes the `COLLECTIONS_CHANGED` event to be less "smart", and _not_ emit a "collection changed" event when the draft version of an entity in the collection is deleted or un-deleted. **This is a change to the event semantics.** The reason behind this is below, but the short version is that collections aren't supposed to be specific to draft (or published) versions of entities, so changing (including deleting) a draft entity in a collection should not necessarily be considered changing the collection contents.

------------------------------

Background for the third change:

I tried to make the `COLLECTIONS_CHANGED` event "smart" so that it would reflect any user-visible changes to collections. Specifically, if you soft-deleted an entity that was in a collection, you'd get a `COLLECTIONS_CHANGED` event with `entities_removed=[...]` telling you that that entity is no longer in the collection. And if you then un-deleted that entity, you'd get a `COLLECTIONS_CHANGED` event with `entities_added=[...]` showing that the entity was "re-added" to the collection when it was un-deleted.

BUT this was fundamentally problematic because collections very deliberately are not part of the draft-publish workflow, and they technically shouldn't care about the "draft" state of an entity in particular. When testing the modulestore migrator in openedx-platform, I noticed that the related code was emitting redundant `COLLECTIONS_CHANGED` events, because of this mismatch.

The problem was with code like this:

```python
        # Should emit 2 events ("collection created", "entities added to collection") but emits 3 (created, added, added again):
        with api.bulk_draft_changes_for(lp1.id, changed_by=None, changed_at=now_time):
            col1 = api.create_collection(lp1.id, "col1", title="Collection 1", created_by=None)
            entity = _create_entity_with_version(lp1.id, "entity1")
            api.add_to_collection(lp1.id, "col1", PublishableEntity.objects.filter(id=entity.id))
```

When the bulk draft context ends, the `emit_collections_changed_for_entity_changes_task` would look for "un-deleted" entities that are part of a collection, and emit "added to collection" events for them, but if you created them in this way, there would be duplicate events - one queued by `add_to_collection`, and one queued when the bulk draft changes context ends.

Now, it was possible to work around this using 7903aec91d369e825988b7b53f6196eeb2754179 to distinguish between "un-deleted" and "created" entities, and this mostly solved the bug, but there was still at least one major edge case:

```python
        # Delete existing entity:
        api.soft_delete_draft(entity.id, deleted_by=None)

        # Should emit 2 events ("collection created", "entities added to collection") but emits 3 (created, added, added again):
        with api.bulk_draft_changes_for(lp1.id, changed_by=None, changed_at=now_time):
            api.set_draft_version(entity.id, v1.id)  # Un-delete the entity
            col1 = api.create_collection(lp1.id, "col1", title="Collection 1", created_by=None)
            api.add_to_collection(lp1.id, "col1", PublishableEntity.objects.filter(id=entity.id))
```
In this case, the handler that runs when the bulk context ends ("look for un-deleted entities and emit 'added to collection' events for any collections they're part of") has no way to "know" that the `add_to_collection` API already emitted events for adding those entities to the collection.

As far as I can tell, there is no clean way to update our code to avoid duplicate events in this case, while still emitting these "smart" events that reflect the draft state of the entities in the collection. Now, this isn't a _huge_ deal as the duplicate events are only in a very particular edge case involving "un-deleting" entities, but the fact that we cannot really avoid this without some ugly hacks tells me that the event design needs to be improved. Which is why I'm suggesting this particular improvement now, because it's technically a breaking change to the semantics of the brand-new `COLLECTION_CHANGED` event, and I want to "get it right" in our first release if possible, to avoid difficult changes later.

What this PR changes: `COLLECTION_CHANGED` is no longer emitted when entities are soft-deleted or un-deleted, potentially affecting the contents of a collection. Instead, code that cares about this needs to also listen for `ENTITIES_DRAFT_CHANGED` events indicating drafts are deleted or un-deleted.


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

bradenmacdonald:
@ormsbee @kdmccormick  Could I get your opinion here?

When we delete a [draft] component or container, do you think openedx-core should:

(A) send both the regular `ENTITIES_DRAFT_CHANGED` event and a `COLLECTION_UPDATED` event, indicating that the entity has been removed from the collection? (Because the draft entity has been deleted).

* This makes the platform-side implementation of the events much simpler and generally more correct, but
* This can result in duplicate "added to collection" events in a very specific edge case involving un-deleting entities and assigning them to a collection inside a bulk draft operation. (See above - I don't think this is a big deal though)
* This more or less assumes/formalizes that collections are collections of _draft_ entities, and we care specifically about the draft being deleted. This matches how we want the libraries authoring UI to work. I'm not sure if it was the original intention of "collections" or not - the code explicitly says collections themselves aren't versioned but it's unclear if they are meant to hold entities regardless of published/draft/deleted status, or generally hold draft entities.

or

(B) send only the regular `ENTITIES_DRAFT_CHANGED` event indicating the component/container has been deleted.

* This requires more work on the platform side, because everything that cares about "draft entities in collections", e.g. library and search code, has to listen for both the `ENTITIES_DRAFT_CHANGED` and `COLLECTION_UPDATED` events in almost all cases.
* This avoids the edge case with un-deleting inside a bulk change - no extra duplicate event gets emitted.
* This formalizes that collections are strictly sets of PublishableEntities which may or may not have draft or published versions.

The docs at https://github.com/openedx/openedx-core/blob/main/src/openedx_content/applets/collections/models.py don't say either way. Though collections _could_ have been implemented as foreignkeys to the `Draft` model instead of `PublishableEntity` if that's what we had wanted.

ormsbee:
I would say (B). My concern with (A) is precisely what you lay out:

> This more or less assumes/formalizes that collections are collections of draft entities, and we care specifically about the draft being deleted. This matches how we want the libraries authoring UI to work. I'm not sure if it was the original intention of "collections" or not - the code explicitly says collections themselves aren't versioned but it's unclear if they are meant to hold entities regardless of published/draft/deleted status, or generally hold draft entities.

Collections are also used as a way for people authoring Courses to find library content to add to them. When they are doing so, they need to be viewing the published state of those items. Something with a deleted draft (where that deletion has not yet been published) should still show up in that Collection when it's being browsed from the "add to course" UI.

bradenmacdonald:
> Something with a deleted draft (where that deletion has not yet been published) should still show up in that Collection when it's being browsed from the "add to course" UI.

I'm not sure this is working well at the moment. Some of the search index code conflates "draft doesn't exist" with "entity doesn't exist". But that's just a bug we can fix in the future, and I agree in principle, so I'll move forward with (B).

ormsbee (approved):

bradenmacdonald:
@ormsbee Are there any special considerations to how we merge this, given the Verawood branch etc.? I _think_ we haven't had any breaking changes to keep separate, so we can just merge this and tag 1.0.2 for use by both master + Verawood?

bradenmacdonald:
Rebased to fix some weird conflict with line endings; no code changes.

bradenmacdonald:
I see there have been some version bump PRs etc., so just to be safe I'm going to call this v1.1.0 and I'll backport it to `verawood-backports` as version 1.0.2 for Verawood.

PR #607, opened 2026-06-18
fix: prefer entity key in TOML file when restoring

ormsbee:
Prior to this commit, we would always assume the entity key would match the directory naming, but that is not a hard requirement. Furthermore, it looks like something in the backup code is adding hashes to the directories unnecessarily, so this was coming up quite often (e.g. where a Unit and HTMLBlock share the last part of their key).


src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -1060,7 +1061,17 @@ def _get_organized_file_list(self, file_paths: list[str]) -> dict[str, Any]:
...
+        # This is going to map static file directory roots to the appropriate
+        # entity keys.
+        comp_paths_to_keys = {}
+
+        # The ordering of the file processing is important because we need to
+        # ensure that TOML files for a given component are processed before the
+        # static files for that component. '.' sorts before '/', so "foo.toml"
+        # will sort before "foo/component_versions/v1/static/figure1.webp" or
+        # any other subdirectory of the "foo" component. Processing the TOML
+        # first allows us to map the directory to a entity key.
```
kdmccormick: [thinking out loud -- no action needed] I was wondering if this would break for entity refs that contain dots, but it seems like it should be OK since `entities/foo.bar.toml` will still sort before `entities/foo.bar/component_versions/etc`.

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -1060,7 +1061,17 @@ def _get_organized_file_list(self, file_paths: list[str]) -> dict[str, Any]:
             "collections": [],
         }
 
-        for path in file_paths:
+        # This is going to map static file directory roots to the appropriate
+        # entity keys.
+        comp_paths_to_keys = {}
```
kdmccormick: consider changing key->ref here and in the rest of the function body (entity_key->entity_ref, etc)
```suggestion
        # This is going to map static file directory roots to the appropriate
        # entity refs.
        comp_paths_to_refs: dict[str, str] = {}
```

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -1073,21 +1084,44 @@ def _get_organized_file_list(self, file_paths: list[str]) -> dict[str, Any]:
                 if path.endswith(".toml"):
                     # Component entity TOML files
                     organized["components"].append(path)
+                    component_toml_str = self._read_file_from_zip(path)
+                    component_toml = tomllib.loads(component_toml_str)
+                    entity_key = component_toml['entity']['key']
+                    comp_path = path[:-5]  # removes the ".toml" at the end
```
kdmccormick: [optional, nit]
```suggestion
                    comp_path = path.removesuffix(".toml")
```

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -1073,21 +1084,44 @@ def _get_organized_file_list(self, file_paths: list[str]) -> dict[str, Any]:
...
                     # Path structure: entities/<namespace>/<type>/<component_id>/component_versions/<version>/static/...
-                    # Example: entities/xblock.v1/html/my_component_123456/component_versions/v1/static/...
-                    component_key = Path(path).parts[1:4]  # e.g., ['xblock.v1', 'html', 'my_component_123456']
+                    # Example: entities/xblock.v1/html/my_component_a822bb/component_versions/v1/static/...
+
+                    # e.g. 'entities/xblock.v1/html/my_component_a822bb'
+                    component_root_path = '/'.join(Path(path).parts[0:4])
+
+                    try:
+                        component_identifier = comp_paths_to_keys[component_root_path]
```
kdmccormick:
```suggestion
                        component_ref = comp_paths_to_refs[component_root_path]
```

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -1073,21 +1084,44 @@ def _get_organized_file_list(self, file_paths: list[str]) -> dict[str, Any]:
...
+                    except KeyError:
+                        self.errors.append(
+                            {
+                                "file": path,
+                                "errors": "Could not find destination entity key for component static file."
```
kdmccormick: This will happen iff the component's TOML file was not seen before its first static asset, right? Would it make sense to say that more explicitly?
```suggestion
                                "errors": (
                                    "Component static file is missing a matching component definition file "
                                    f"(expected definition file at: {component_root_path}.toml)"
                                )
```

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -282,7 +283,7 @@ def create_zip(self, path: str) -> None:
             Exception: If the learning package cannot be found or if the zip creation fails.
         """
 
-        with zipfile.ZipFile(path, "w", compression=zipfile.ZIP_DEFLATED) as zipf:
+        with zipfile.ZipFile(path, "w", compression=zipfile.ZIP_DEFLATED, compresslevel=9) as zipf:
```
kdmccormick, line 286: what's the reasoning for adding `compresslevel=9`, and could you add it to the commit message?

kdmccormick (commented):
Just a couple small requests

src/openedx_content/applets/backup_restore/zipper.py
ormsbee: Oh right, I should separate this. I just noticed it while testing and slapped it on. The justification is that it brings the test file I was using from 614 MB to 568 MB, basically for free. Normally, it doesn't make that big a difference, but I suspect it might have to do with the fact that we're building the zip file so incrementally here. I'll separate it out.

src/openedx_content/applets/backup_restore/zipper.py
ormsbee: Yeah, as long as the export keeps the convention that the TOML file and the dir structure have matching names, then we can rely on the sort order. That matching name is the only thing actually correlating the two, so if those don't match, the archive is broken.

ormsbee:
@kdmccormick: Should be good for another review pass. Before merging, I'll squash down to two commits, one for the compressionlevel thing, and one for everything else.

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -1060,7 +1061,17 @@ def _get_organized_file_list(self, file_paths: list[str]) -> dict[str, Any]:
...
+        # This is going to map static file directory roots to the appropriate
+        # entity keys.
+        comp_paths_to_refs = {}
+
+        # The ordering of the file processing is important because we need to
+        # ensure that TOML files for a given component are processed before the
+        # static files for that component. '.' sorts before '/', so "foo.toml"
+        # will sort before "foo/component_versions/v1/static/figure1.webp" or
+        # any other subdirectory of the "foo" component. Processing the TOML
+        # first allows us to map the directory to a entity key.
```
kdmccormick:
```suggestion
        # entity refs.
        comp_paths_to_refs = {}

        # The ordering of the file processing is important because we need to
        # ensure that TOML files for a given component are processed before the
        # static files for that component. '.' sorts before '/', so "foo.toml"
        # will sort before "foo/component_versions/v1/static/figure1.webp" or
        # any other subdirectory of the "foo" component. Processing the TOML
        # first allows us to map the directory to a entity ref.
```

src/openedx_content/applets/backup_restore/zipper.py
```diff
@@ -1073,21 +1084,44 @@ def _get_organized_file_list(self, file_paths: list[str]) -> dict[str, Any]:
                 if path.endswith(".toml"):
                     # Component entity TOML files
                     organized["components"].append(path)
+                    component_toml_str = self._read_file_from_zip(path)
+                    component_toml = tomllib.loads(component_toml_str)
+                    entity_ref = component_toml['entity']['key']
+                    comp_path = path.removesuffix(".toml")
+
+                    # This maps the root path of a component, e.g."entities/xblock.v1/html/my_component_a822bb"
+                    # to the actual key, e.g. "xblock.v1:html:my_component". The last part of the key  will
```
kdmccormick:
```suggestion
                    # to the actual ref, e.g. "xblock.v1:html:my_component". The last part of the ref  will
```

kdmccormick (approved):
Two tiny nits, LGTM otherwise! No need for re-review.

I wasn't able to manually verify this myself (dev env issues).

PR #619, opened 2026-06-30
docs: Update CBE code location ADR

kdmccormick:
This ADR was written before the openedx-learning -> openedx-core repo rename and restructure happened. It also assumes that the Learning Pathways project would have set up a `src/openedx_learning` app already, which it has not yet. This commit fixes the ADR in both respects.


mgwozdz-unicon (approved):
Thank you for updating this for us!

PR #636, opened 2026-07-03
docs: cbe competency rule profile adr amendments

mgwozdz-unicon:
Claude assisted...

## Summary

Amends ADR 0002 (competency criteria model) and ADR 0003 (versioning) to close gaps in the `CompetencyRuleProfile` design in this Slack thread: https://openedx.slack.com/archives/C09A1BN1CSE/p1783010474048009

## Changes

**ADR 0002**

- `CompetencyRuleProfile` scope fields (`organization_id`, `course_id`, `competency_taxonomy_id`) are now consistently nullable, at most one populated per row; combining more than one on a single profile is explicitly out of scope for now (documented as a new Rejected Alternative, since a required-parent-scope hierarchy was considered and doesn't fit taxonomies that span multiple organizations).
- Added a unique constraint on the scope-column index, so at most one profile can exist per distinct scope combination.
- Added `CompetencyTaxonomy.taxonomy_overrides_org`, a boolean (default `false`) that breaks the tie when both an organization-scoped and a taxonomy-scoped profile could apply to the same criterion.
- Rewrote how `CompetencyCriterion.competency_rule_profile_id` gets assigned: it's recomputed at several explicit trigger points (creation, a more specific profile appearing later, a per-criterion override being set, an override being cleared back to the default) rather than resolved dynamically at evaluation time. Added the resolution-order tables (by which profile rows exist) and the org/taxonomy tiebreak table directly to the ADR.
- `CompetencyRuleProfile` is now never hard-deleted, even pre-use (a carve-out from the general delete-protection rule) — retirement is archive-only via a new `archived` column. Editing is restricted to `rule_type`/`rule_payload`; scope fields are immutable after creation.
- Corrected the `Grade` rule payload example to the platform's actual fractional grade scale (0.0–1.0, not 0–100), and specified the allowed `op` values (`gte`, `lte`, `eq`).

**ADR 0003**

- Decision 4 now defines "in use" for a `CompetencyRuleProfile` explicitly, and extends the existing edit-warning requirement to cover profile reassignment and override-state switching, not just direct edits.

### DIFF
--- a/docs/openedx_learning/decisions/0002-competency-criteria-model.rst
+++ b/docs/openedx_learning/decisions/0002-competency-criteria-model.rst
@@ -54,6 +54,7 @@ Decision
 
    This new database table will have the following columns:
    1. ``taxonomy_ptr_id``: Primary key and one-to-one foreign key to ``oel_tagging_taxonomy.id``.
+   2. ``taxonomy_overrides_org``: Boolean, defaults to ``false``. Used only while computing which single ``CompetencyRuleProfile`` to assign to a ``CompetencyCriterion`` (Decision 4). If, for a criterion's context, both an organization-scoped profile row and a taxonomy-scoped profile row exist as candidates, this field decides which one gets assigned: ``false`` (default) assigns the organization-scoped row; ``true`` assigns this taxonomy's own row instead, so it cannot be overridden by an organization. Once assigned, the criterion stores that one profile's id and this field plays no further part. Not exercised in this phase; see the MVP note in Decision 4.
 
    Lifecycle rules for this parent/child pair:
 
@@ -113,19 +114,27 @@ Decision
 
    Relationship to other concepts:
 
-   - Can be scoped by taxonomy, course, and/or organization.
+   - Each row is scoped by at most one of taxonomy, course, or organization (or by none, for the system default). See Decision 4 for how a criterion is assigned a profile when rows in more than one of these scopes could apply to it.
+   - At most one profile row may exist per distinct scope value (unique constraint, Decision 5).
+   - The system default is the single profile row where all three scope fields are null. It is seeded, not created/edited/deleted through the profile API.
    - Is referenced by ``CompetencyCriterion``, which may override its type/payload.
+   - Never hard-deleted; retirement is archive-only (Decision 7).
 
    This new database table will have the following columns:
 
    1. ``id``: unique primary key
    2. ``organization_id``: The ``organization_id`` of the organization that this competency rule profile is scoped to. Null if it is not scoped to a specific organization.
    3. ``course_id``: The ``course_id`` of the course that this competency rule profile is scoped to. Null if it is not scoped to a specific course.
-   4. ``competency_taxonomy_id``: The ``CompetencyTaxonomy.taxonomy_ptr_id`` of the competency taxonomy that this competency rule profile is scoped to.
+   4. ``competency_taxonomy_id``: The ``CompetencyTaxonomy.taxonomy_ptr_id`` of the competency taxonomy that this competency rule profile is scoped to. Null if it is not scoped to a specific taxonomy.
    5. ``rule_type``: “View”, “Grade”, “MasteryLevel” (Only “Grade” will be supported for now)
    6. ``rule_payload``: JSON payload keyed by ``rule_type`` to avoid freeform strings. It is structured JSON (not arbitrary freeform data): each ``rule_type`` defines the allowed payload shape and required keys, and validation enforces this contract. JSON is used instead of fixed columns like ``op``, ``value``, and ``scale`` so that future rule types (for example, ``MasteryLevel`` thresholds or plugin-defined evaluators such as CEL-based rules) can add their own fields without repeated schema migrations or many nullable columns. Examples:
 
-      1. ``Grade``: ``{"op": "gte", "value": 75, "scale": "percent"}``
+      1. ``Grade``: ``{"op": "gte", "value": 0.75, "scale": "percent"}``. Allowed ``op`` values: ``gte``, ``lte``, ``eq``. ``value`` must be a fraction between 0.0 and 1.0 inclusive, matching the platform's existing fractional grade representation, not a 0-100 scale.
+   7. ``archived``: Boolean, defaults to false. Set instead of deleting a profile that is no longer wanted. Archived profiles are hidden from authoring and new associations but remain queryable, so existing ``CompetencyCriterion`` rows and learner status history stay resolvable.
+
+   Editing a profile may change ``rule_type``/``rule_payload`` only; scope fields are immutable after creation, to avoid silently re-scoping criteria that already resolved to this profile.
+
+   MVP note: only taxonomy-scoped and system-default profiles are created; every profile row created in this phase has ``course_id`` and ``organization_id`` null. Enabling course- or organization-scoped profiles later needs no schema change here -- the columns, the uniqueness constraint, and the re-assignment behavior (Decision 4) already support them.
 
 4. ``CompetencyCriterion`` concept (``CompetencyCriteria`` database table)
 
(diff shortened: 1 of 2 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -113,19 +114,27 @@ Decision
 
    Relationship to other concepts:
 
-   - Can be scoped by taxonomy, course, and/or organization.
+   - Each row is scoped by at most one of taxonomy, course, or organization (or by none, for the system default). See Decision 4 for how a criterion is assigned a profile when rows in more than one of these scopes could apply to it.
```
jesperhodge: Question: if taxonomy + course + organization are all none, this is system default. Does uniqueness apply to the system default as well, i.e. there can only be one system default rule profile?

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -54,6 +54,7 @@ Decision
 
    This new database table will have the following columns:
    1. ``taxonomy_ptr_id``: Primary key and one-to-one foreign key to ``oel_tagging_taxonomy.id``.
+   2. ``taxonomy_overrides_org``: Boolean, defaults to ``false``. Used only while computing which single ``CompetencyRuleProfile`` to assign to a ``CompetencyCriterion`` (Decision 4). If, for a criterion's context, both an organization-scoped profile row and a taxonomy-scoped profile row exist as candidates, this field decides which one gets assigned: ``false`` (default) assigns the organization-scoped row; ``true`` assigns this taxonomy's own row instead, so it cannot be overridden by an organization. Once assigned, the criterion stores that one profile's id and this field plays no further part. Not exercised in this phase; see the MVP note in Decision 4.
```
jesperhodge: Questions:
- the taxonomy here decides if its own competency rule profile should be overridden by the org course profile if present or not. Why is this decided by the taxonomy?
- "Not exercised in this phase" - what does that mean for MVP implementation? Create but not use this field?

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
jesperhodge: Question:
Trying to understand how this works in practice. Let's say for example we have a course, and this course has various assignments. Some of these assignments demonstrate the "Python" skill, some of these demonstrate the "JS" skill. So we make sure the correct competency taxonomy that contains different skills is linked (is this linking done by the course author?)
Then the author can use AND/OR statements to apply the competency rule profile to different assignments.
Since there is a course-wide or taxonomy-wide rule profile, initially the taxonomy decides that all assignments need to have gte 75% of success (why is this specified for the taxonomy, as it applies to assignments in a course?) or the author can set a taxonomy rule profile for the course stating that all assignments need to meet 75% of success, and he can override this per assignment if desired.
But unless the author manually edits assignment rules, they can't set e.g. 75% for Python assignments, 50% for JS assignments, and 65% for take-home assignments in Python?
So... how is having one rule-profile per course or even per taxonomy helpful? Why don't they differentiate by skill or assignment type or something else?

jesperhodge: Alternatively: why have this complicated construct if it can create hurdles (like not distinguishing between skill and assignment types)?
Why not like this?
- `rule_type` and `rule_payload` can be set per assignment if desired, like before, overriding any course-wide or org-wide values or values that are set up higher in the criteria tree (similar to how problem grading works now)
- The author can define a `CompetencyRule` that he assigns to one or more assignments/tag applications (criterion), or to a skill tag via `CompetencyCriteriaGroup` / to a tree node, or to the whole course or multiple courses. That relationship can be reflected via another table
- Even CompetencyRules that are applied to a whole course can be marked to apply only to specific skills or assignment types
- No more fixed inflexible rule profiles - the author just creates a rule and assigns it to whatever he wants, and being able to apply it to the whole course or even org allows him to be efficient.

jesperhodge (changes requested):
I left some questions.
On a high level, I wonder instead of rule profiles if it's not better to allow course authors to define rules and assign them to whatever they want, including but not limited to the whole course - see my comments for more detail.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
mgwozdz-unicon: Yes, there should only be 1 system default rule profile.

jesperhodge (approved):
The ADR looks good to me, from my side it's ready.

The points from my previous review have been answered, and they are not relevant:

CBE Courses are already designed in a way where it is intended for you to have one grade threshold across the board for a course, or sometimes different rules for different taxonomies.
But they generally don't differentiate per skill, per assignment type, and such.

Nit: as discussed, adding some additional explanation for this will be helpful, just for background. But don't let that block this PR.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
jesperhodge: My point is mute (see https://github.com/openedx/openedx-core/pull/636#pullrequestreview-4647681664).

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -158,7 +236,7 @@ Decision
    6. ``StudentCompetencyCriteriaStatus(user_id, competency_criteria_id)``
    7. ``StudentCompetencyCriteriaGroupStatus(user_id, competency_criteria_group_id)``
    8. ``StudentCompetencyStatus(user_id, oel_tagging_tag_id)``
-   9. ``CompetencyRuleProfile(competency_taxonomy_id, course_id, organization_id)``
+   9. ``CompetencyRuleProfile(competency_taxonomy_id, course_id, organization_id)`` (unique -- at most one profile per distinct scope combination)
```
ormsbee: How are we ensuring unique combinations if there are nullable fields?

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
mgwozdz-unicon: I'm drafting some clarifications to this in the ADR now, but to answer your questions:
1. Why the taxonomy decides the override: a single taxonomy can compete against many organization-scoped profiles, so the tiebreak needs one consistent answer that applies to all of them. Putting it on the taxonomy avoids setting (and keeping in sync) the same flag on every org-scoped profile row. Added as a new bullet under Decision 1's "Relationship to other concepts."

2. "Not exercised in this phase" means: the column is created and defaults to `false`, but no code reads it yet, since organization-scoped profiles aren't created in this phase (see the MVP note in Decision 4), so the conflict it resolves can't occur yet. Reworded that sentence in the column description to say this directly.

mgwozdz-unicon:
My latest commit addresses the open review threads, plus one design gap two of them turned out to share.

**Jesper, `taxonomy_overrides_org` field (Decision 1):** added a bullet explaining why the flag lives on the taxonomy rather than the org-scoped profile: a single taxonomy can compete against many org-scoped profiles, so the tiebreak needs one answer that applies to all of them, not one duplicated per org.

**Jesper, "Not exercised in this phase" (Decision 1):** reworded that sentence. The field is created and defaults to `false`, but nothing reads it yet since org-scoped profiles aren't created in this phase, so the conflict it resolves can't occur.

**Jesper, system default uniqueness (Decision 3):** confirmed only one system default row should ever exist, but this surfaced a real gap: a plain unique constraint across nullable scope columns doesn't enforce that in SQL, `NULL` is never equal to `NULL`, and our MySQL backend doesn't support the usual partial/conditional constraint workaround either. Fixed by adding a generated `scope_key` column that collapses the three scope columns into one comparable value, with an ordinary unique constraint on it (Decision 3, Decision 5). Also added a check constraint so at most one of the three raw scope columns can be set per row.

**Dave, same gap on the index list (Decision 5):** same fix as above. Also documented the rejected partial-constraint approach in Rejected Alternatives so it doesn't get re-proposed later.

**Jesper, background on why this feature exists (Context):** per our conversation, the Context section now states plainly that in the common case, the mastery rule is the same across the whole instance (the system-wide default), and that taxonomy, organization, course, and criterion scoping/overrides exist to handle progressively narrower exceptions to that default, not the typical case.

**Also:** made the system default profile editable (`rule_type`/`rule_payload`) through the profile API. It's still only ever created or deleted via the migration seed, not through the API.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -113,19 +118,30 @@ Decision
 
    Relationship to other concepts:
 
-   - Can be scoped by taxonomy, course, and/or organization.
+   - Each row is scoped by at most one of taxonomy, course, or organization (or by none, for the system default). A check constraint enforces that at most one of ``organization_id``, ``course_id``, and ``competency_taxonomy_id`` is non-null per row. See Decision 4 for how a criterion is assigned a profile when rows in more than one of these scopes could apply to it.
+   - At most one profile row may exist per distinct scope value. This is enforced by a unique constraint on the generated ``scope_key`` column (Decision 5), not a plain unique constraint on the three raw scope columns; see the ``scope_key`` column definition below for why.
+   - The system default is the single profile row where all three scope fields are null, which gives it ``scope_key = null``. Null values are never unique-checked, so this row's singularity is guaranteed procedurally instead: it is seeded once via migration and never created or deleted through the profile API. Its ``rule_type``/``rule_payload`` can be edited through the profile API like any other profile.
```
ormsbee: Let's not make this field nullable. We can either keep the scope format with blank values for the specific parts, e.g. `org:,course:,taxonomy:`, or even make it blank, since a blank string will be unique and prevent duplicates.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -113,19 +118,30 @@ Decision
...
+   4. ``competency_taxonomy_id``: The ``CompetencyTaxonomy.taxonomy_ptr_id`` of the competency taxonomy that this competency rule profile is scoped to. Null if it is not scoped to a specific taxonomy.
+   5. ``scope_key``: Database-generated column, not set directly, that collapses the three scope columns above into one comparable value: for example ``"org:5"``, ``"course:12"``, or ``"taxonomy:7"``, and null when all three scope columns are null (the system default). This exists because SQL never treats two ``NULL`` values as equal for uniqueness purposes, so a plain unique constraint across the three nullable scope columns would not stop two rows from sharing the same scope (for example two rows both with ``organization_id=5`` and the other two columns null). Collapsing the scope into one generated column that is non-null whenever a row is actually scoped sidesteps that, and does so identically on every database backend this project supports, including MySQL, which does not support the conditional/partial unique indexes that would otherwise be the usual fix.
```
ormsbee: 1. Just to clarify, this is being set by app logic based on the database values, right?
2. Whatever this format is, please make it consistent and trivial to parse, e.g. if it's always "org:X,course:Y,taxonomy:Z" regardless of blank values.
3. I just want to flag that these are all internal ID references, and that's fine, so long as we never expect to directly export this anywhere.
4. @kdmccormick: May I get your thoughts on this naming? This is dynamically generated to make sure we can enforce unique constraints with nullable inputs, but I realize it's not an OpaqueKey as the `_key` suffix implies under [OEP-68](https://docs.openedx.org/projects/openedx-proposals/en/latest/best-practices/oep-0068-bp-content-identifiers.html#summary). Does this make sense as `scope_code`?

ormsbee (changes requested):
Minor requests/questions on the `scope_key`.

FWIW, if we want to handwave the details of how the scope key is derived and what it's named, and pick that up again in the implementation PR, I'm totally fine with that. I generally want to avoid full schema specification in ADRs, since this may drift from the model over time, and I tend to prefer really detailed documentation in the models files instead.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
mgwozdz-unicon: I just committed with changes in accordance with these answers:
1. I think it can be database generated based on the database values. Or does it need to reference the MongoDB values?
2. Went with "org:X,course:Y,taxonomy:Z" 
3. Yes, all internal ID references with no expectation to export.
4. Renamed to `scope_code`.

Happy to revise further if needed.

ormsbee: 1. MongoDB doesn't need to be involved in this at all. I just wanted to make sure we weren't actually using SQL to generate it or something. What you have is fine, thank you.

ormsbee (approved):

jesperhodge (approved):
Amazing work!

Issue #634, opened 2026-07-02
Remove Taxonomy subclasses functionality + system-level taxonomies


bradenmacdonald:
## Summary of Actions

* Remove the "system-defined" taxonomy feature entirely, replacing it with a `read_only` flag on the base Taxonomy class
* Remove all taxonomy subclasses and the `cast()` functionality
* Remove the built-in "Language Taxonomy" and "User Taxonomy" which weren't useful nor usable.

# Previous Context And Discussion

Very relevant to the [CBE Taxonomy Type](https://github.com/openedx/openedx-core/issues/618) use case.

## Background

Our taxonomy models currently support the idea of taxonomy types, where each taxonomy can be a "normal" taxonomy but could also be a "typed" taxonomy implemented using a subclass of `Taxonomy` that adds extra functionality.

## How it works:

`Taxonomy` instances have [a `_taxonomy_class` column](https://github.com/openedx/openedx-core/blob/6d50ec174258d365e9fe9939d361f33340e500d5/src/openedx_tagging/models/base.py#L298-L306) which can be accessed on the model as [`taxonomy.taxonomy_class`](https://github.com/openedx/openedx-core/blob/6d50ec174258d365e9fe9939d361f33340e500d5/src/openedx_tagging/models/base.py#L331-L360), returning the actual `Taxonomy` subclass used for that instance. When loading taxonomies, the [`Taxonomy.cast()`](https://github.com/openedx/openedx-core/blob/6d50ec174258d365e9fe9939d361f33340e500d5/src/openedx_tagging/models/base.py#L378-L394) method will "reload" the taxonomy using the correct subclass.

## Why we have it

We originally intended to have more built-in taxonomy types, only fragments of which still actually exist:

* [System defined taxonomy](https://github.com/openedx/openedx-core/blob/6d50ec174258d365e9fe9939d361f33340e500d5/src/openedx_tagging/models/system_defined.py#L21-L35) - the most basic subclass; all it does is override `system_defined=True` to mark these taxonomies as read-only (users shouldn't manually add/remove tags).
[...]

Issue #625, opened 2026-07-01
[Arch] Determine Competency ID field strategy for taxonomy competency data


mgwozdz-unicon:
### Use Case

As a Platform Administrator, I want a stable, human-readable Competency ID visible for each competency in the Competency Management UI and on the assignment workflow, so that I can reference and communicate specific competencies using a meaningful identifier that my institution recognizes.

### Description

**Current behavior**

The existing tag/taxonomy data model exposes several identifier fields per tag: an internal PK (`id`, never user-visible), a stable external reference (`uuid`), and a client-meaningful key (`_key` / `key`). None of these is designed for use as a human-readable Competency ID in a management UI or assignment workflow. No such field currently exists.

**Requested change**

Determine the ID field strategy for competency taxonomies: which field or fields represent the Competency ID, how the ID is assigned, and how it behaves across import, re-import, API retrieval, and UI display. This spike produces a documented recommendation for the data model implementation ticket — not code.

Key constraints the recommendation must respect:

- This is an internal identifier. If CASE IDs or CTDL IDs are added in the future, they are separate fields with their own semantics — this ID is not a CASE ID.  
- Changes to the data model are additive; existing identifier fields are not replaced.  
- The recommendation must be approved by the engineering team and the Open edX community.

### Acceptance Criteria

This is an architecture spike. The deliverable is a documented technical recommendation, not working code. There is no Gherkin and no QA path.

- [ ] Recommendation document published (see Open Questions for format).  
- [ ] Recommendation specifies whether a new field is added or an existing field is repurposed, and justifies the choice.  
- [ ] ID assignment strategy decided and documented (see Open Questions).  
- [ ] Uniqueness scope decided: within a taxonomy, globally, or advisory only.  
- [ ] Import and re-import behavior specified: whether the ID is preserved, overwritten from the import file, or conflict-flagged.  
- [ ] API serialization specified: which field is exposed and under what name.  
- [ ] Backward compatibility and migration implications documented.  
- [ ] Recommendation reviewed and approved by the engineering team and Open edX community.  
- [ ] All Open Questions resolved before this ticket is closed.

### Open Questions

1. **ID assignment strategy (architect, blocking):** Two options:  
     
   - Option A (auto-generate with override): derive a default from the competency name — initials of each word plus a hierarchical position suffix (e.g., `PK-1`, `PK-1.2`, `PK-1.2.3` for root / second level / third level). User can override. The exact format of the suffix is part of what this spike defines.  
[...]

Issue #613, opened 2026-06-30
[BE] Implement CBE core data models (CompetencyTaxonomy, criteria, learner status)


mgwozdz-unicon:
## Use Case

Competency-Based Education (CBE) programs require learners to *demonstrate* mastery of specific
competencies, not just complete coursework. Open edX currently has no way to define or evaluate the
rules that determine whether a learner has demonstrated a competency, for example "a learner must
score 80% or higher on Assignment 1 OR Assignment 2 to demonstrate the Writing Poetry competency."

This issue implements the foundational database layer that makes the following possible:

- Course Authors and Platform Administrators can define competency achievement criteria in Studio,
  specifying which course content (assignments, exams, etc.) a learner must complete, at what
  threshold, and with what AND/OR logic.
- The platform can track each learner's progress toward demonstrating each competency as they
  receive grades and completions.
- A history of criteria changes is preserved for audit and traceability.

Without this data model, no competency progress dashboards, automated competency evaluations, or
CBE-specific authoring tools can be built.

---

## Description

Implement the Django database models defined in
[ADR-0002 (Competency Criteria Model)](https://github.com/openedx/openedx-core/blob/main/docs/openedx_learning/decisions/0002-competency-criteria-model.rst),
[ADR-0003 (Competency Criteria Versioning)](https://github.com/openedx/openedx-core/blob/main/docs/openedx_learning/decisions/0003-competency-criteria-versioning.rst)
and
[ADR-0004 (Competency Mastery Concurrency)](https://github.com/openedx/openedx-core/blob/main/docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst).
This is a data-layer issue: no REST endpoints, no UI, and no business logic beyond schema
constraints and two seed migrations. Delete protection was originally moved out to #799 pending #655.
#655 closed on 2026-09-02 with an approved design, and #799 is now closed as superseded. The
`on_delete` values that design implies for this issue's foreign keys are set out in the Deletions
section below, since revised by an amendment to ADR-0002 Decision 7.

The code goes in `src/openedx_learning/applets/cbe/`, the app created by PR #712 per
[ADR-0001](https://github.com/openedx/openedx-core/blob/main/docs/openedx_learning/decisions/0001-competency-criteria-location.rst).

**Already delivered.** PR #712 added the `openedx_learning` app and the `CompetencyTaxonomy` model.
This issue adds the one column that PR left out, `taxonomy_overrides_org`.

**Authoring and definition models**

[...]

PR #637, opened 2026-07-03
docs: add ADR for institution-defined tag code field

mgwozdz-unicon:
Claude assisted...

## Summary
- Proposes `Tag.code`, a generic, optional field on `openedx_tagging`'s `Tag` model, to hold an institution's own competency identifier — distinct from CASE/CTDL and from the existing `external_id` field.
- Defines a completeness rule requiring `code` for every tag in a Competency Taxonomy, enforced through a generic property on `Taxonomy` that only a Competency Taxonomy subclass turns on, so `openedx_tagging` never has to know what a Competency Taxonomy is.
- Specifies assignment behavior: user-entered when given; a placeholder is generated only when a *new* tag is created with no code (via import or the existing Taxonomy Editing UI, which has no field for it yet); re-import can correct a placeholder but never clears or regenerates an existing value just because a file's column is blank.
- Answers `openedx/openedx-core#625` ("Determine what and how many ID fields should be added to the Taxonomy data model").

## Why
Competency Management UX designs show a "Competency ID" that institutions click to build competency-criteria associations. Institutions need their own internal identifier for this, separate from any future CASE/CTDL identifiers. The existing `external_id` field can't be reused: it's de facto immutable and is the load-bearing lookup key for import/export matching, so giving it a second, editable meaning would break that contract.

## Status
Proposed. This is a recommendation for engineering and Open edX community review, not an implementation — no code changes are included in this PR.

## Dependencies and follow-ups (tracked separately, not in this PR)
- `#613` (implements the `CompetencyTaxonomy` model) needs to override the `tag_code_required` property this ADR defines. This ADR's property should land before or alongside #613, not after it — #613 depends on this ADR, not the reverse.
- `#634` ("Taxonomy subclasses: embrace or deprecate?") is an open architecture discussion that could affect the subclass-override mechanism this ADR relies on.
- A pre-existing bug in the import validation path (a taxonomy-type cast gets discarded) needs its own fix for this ADR's completeness rule to work correctly.
- Implementation tickets for the data model change, GET endpoint exposure, import/export support, Competency Management UI display, and docs/release notes aren't filed yet.

These are called out in a comment on #625 rather than in the ADR itself.


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

docs/openedx_tagging/decisions/0010-tag-code-field.rst
```diff
@@ -0,0 +1,181 @@
...
+Proposed
+
+Context
+-------
+
+`Issue #625 <https://github.com/openedx/openedx-core/issues/625>`_ asks the project to
+determine an ID field strategy for Competency-Based Education (CBE) competency data.
+Competency Management UX designs show a "Competency ID" on each competency tag: a
+short, human-readable identifier an institution can reference in its own systems
+(spreadsheets, internal catalogs, etc). Clicking it drives the workflow that
```
ormsbee: Is it always one identifier, or do institutions sometimes have multiple identifiers for the same tag (e.g. from different systems)?

docs/openedx_tagging/decisions/0010-tag-code-field.rst
```diff
@@ -0,0 +1,181 @@
...
+creates a ``CompetencyCriteria`` association between the competency and content
+criteria rules.
+
+This identifier is explicitly **not** a CASE or CTDL identifier. Those are external
+standards bodies' IDs that may need their own storage in the future; conflating them
+with an institution's internal code would make each harder to reason about
+independently. This ADR is scoped to the internal, institution-defined code only.
```
ormsbee: I don't know if I agree with that. If we can have a robust system that maps namespaced identifiers to tags, I could imagine us using a system where internal namespaces are peers to standardized ones.

That being said, if and when that time comes, we can always migrate the existing data that we have to some new, more general schema. So I do support the idea that we're focusing strictly on internal codes for now, with the simpler model that implies.

docs/openedx_tagging/decisions/0010-tag-code-field.rst
```diff
@@ -0,0 +1,181 @@
...
+``Tag`` already has three identifiers, none of which fit:
+
+- ``id``: internal primary key, never user-visible.
+- ``uuid``: stable external reference, not human-readable or institution-assigned.
+- ``external_id``: intended to link a Tag to a record in an externally-defined
+  taxonomy. It is de facto immutable: no API path mutates it after creation, tag
+  re-sync code documents it as "must be an immutable ID", and import/export uses it
+  purely as the lookup key to find and update existing tags on re-import. Repurposing
+  it as an admin-editable Competency ID would conflict with that existing, load-bearing
+  contract.
```
ormsbee: Can you expand on why this is currently immutable, and why it must remain so? My understanding was that `external_id` was created for a similar use case to what's being proposed here.

ormsbee (commented):
My overall question is around `external_id`, whether it's really such a bad thing for it to become mutable, and why that is. And what would `external_id` be set to vs. `code` should this ADR pass?

docs/openedx_tagging/decisions/0010-tag-code-field.rst
mgwozdz-unicon: It should be only one identifier. Today Competency Frameworks/Taxonomies are sourced through hand crafted spreadsheets rather than any sort of source system probably 99% of the time. I'm pretty sure none of the other LMSs support multiple identifiers for competencies. If there were other identifiers, they would most likely be the CASE GUID or CTDL identifier, which should be explicitly stored as such.

docs/openedx_tagging/decisions/0010-tag-code-field.rst
mgwozdz-unicon: When support is added for the CASE and CTDL identifiers, I can definitely envision wanting something akin to a tag_identifiers table  where the columns of that table are effectively just tag_id, other_id_type, and other_id to support other_id_types of CASE and CTDL. The institution id (code) and/or external_id could potentially be stored in this table too if desired. I wasn't trying to preclude that kind of an idea with this.

mgwozdz-unicon: Although, it might make sense to continue keeping the third party identifiers separate from the institution/openedx identifiers. I'm not sure I'd want to migrate those (external_id and code) into this tags_identifiers table since they'd already be functioning fine where they are. Really the only reason to add a new table at all would be out of a fear that the number of additional identifiers might get unweildy, and really even just the 2 additional (CASE and CTDL) may not constitute "unweildy".

docs/openedx_tagging/decisions/0010-tag-code-field.rst
mgwozdz-unicon: Trying to jog my memory on this... I think it was because we needed an identifier that wasn't internal to Open edX (since sharing identifiers internal to the platform was deemed a security concern) that could be stable on import/export. Because otherwise, if you edit the external_id and re-import, then the platform has no way of knowing what you were editing. Whereas institutions need a mutable identifier that they can keep as human readable and in alignment with the current tag name that may be edited as well.

mgwozdz-unicon:
@ormsbee Great questions - I'm happy to update my draft for clarity around them, but I want to make sure any follow-up you may have is resolved first, so just let me know if you have follow-up questions/thoughts or if you're ready for me to update it.

kdmccormick:
I'm struggling to understand the need for `code` given the identifiers that Tags already have:

* `id`, an immutable instance-local identifier
* `uuid`, an immutable cross-platform identifier
* `external_id`, another immutable cross platform identifier
* `value`, a human-readable, editable identifier

Could you link to the UX designs and/or provide a concrete example of what a given Tag might have for its value, code, and external_id?

mgwozdz-unicon:
@kdmccormick 

> I'm struggling to understand the need for `code` given the identifiers that Tags already have:
> 
> * `id`, an immutable instance-local identifier
> * `uuid`, an immutable cross-platform identifier
> * `external_id`, another immutable cross platform identifier
> * `value`, a human-readable, editable identifier
> 
> Could you link to the UX designs and/or provide a concrete example of what a given Tag might have for its value, code, and external_id?

Yes, here is a screenshot from the [UX designs](https://www.figma.com/design/VhAdaucNQS49x0Afp9xPZw/Competency-Application?node-id=1-3973&t=2uhUGr5twGSNJH6H-4):
image

So in this example, "CCRS-1.3" would be the code value for the tag that has the value "Inference". Here CCRS stands for "Career and College Readiness Skills" which is the name of the Taxonomy. If the institution changed their code (institution competency id) to "CCRS-1.4" and re-uploaded and it had instead been the external_id value, then the platform would have no way of knowing what tag is supposed to get updated to "CCRS-1.4". So in this example, when the institution first does their upload, code and external_id may have the same value, but code can change while external_id must always be the same. The id would be internal to the instance and uuid can't be used for this purpose because it's not human readable.

docs/openedx_tagging/decisions/0010-tag-code-field.rst
ormsbee: So is it accurate to say that if we either exported the UUID, or provided a "previous_id" column in the import CSV format so that we could remap `external_id` entries, we wouldn't need this new `code` field?

kdmccormick:
Thanks, this helps a lot.

>  If the institution changed their code (institution competency id) to "CCRS-1.4" and re-uploaded and it had instead been the external_id value, then the platform would have no way of knowing what tag is supposed to get updated to "CCRS-1.4". 

Is changing the institutional competency id a known use case? They appear to me, visually, like the sort of code that an institution would define once and then make a point to keep stable.

ormsbee:
@mgwozdz-unicon: 

> ... I want to make sure any follow-up you may have is resolved first, so just let me know if you have follow-up questions/thoughts or if you're ready for me to update it.

My main line of questioning is just to see if we can somehow amend the import format so that we can still use the `external_id` field for this data. Otherwise, I worry that we're going to end up with the case that `external_id` will be the same as `code` the vast majority of the time and look like duplication to end users, and that it won't be clear how or why they differ at other times.

jesperhodge:
@mgwozdz-unicon @ormsbee:
Am I right that some of the pushback is because we are trying to add a field to taxonomy that is not only irrelevant to non-competency taxonomies but also is a synonym for external_id yet referring to something else?
I know we planned to not have a `Competency` table but just have one tag in a competency taxonomy be a competency. But I'm wondering if we shouldn't add a `CompetencyTagData` or `CompetencyMetadata` table or something like that. That would allow us to add additional data to each competency tag right in the CBE applet, leaving the tag table as-is. It would be more scalable, because possibly we could in the future have more cases of customizing a taxonomy tag vs a tagging tag, and we would keep a very clean separation of concerns.
The new field `code` or `competency_code` or something like that would go in the new table and not pollute the taxonomy model. I feel we would have a lot more flexibility in adding any fields we want without needing to worry about the taxonomy model itself.
(Similar to how we have a `CompetencyTaxonomy` table now that just adds additional info to a taxonomy, marking it as a competency hierarchy without needing to change the taxonomy model).

kdmccormick:
> But I'm wondering if we shouldn't add a CompetencyTagData or CompetencyMetadata table or something like that. That would allow us to add additional data to each competency tag right in the CBE applet, leaving the tag table as-is. It would be more scalable, because possibly we could in the future have more cases of customizing a taxonomy tag vs a tagging tag, and we would keep a very clean separation of concerns.

I think this is a path forward if there ends up being further competency-specific tag metadata. But at the moment, I think we're both reluctant about proliferation of identifiers (whether CBE-specific or not) without clarity that the existing identifiers don't suffice.

bradenmacdonald:
This use case is exactly what the `external_id` field is already for. "External" means "External to Open edX", not "external to the institution".

Making it mutable is fine.

> If the institution changed their code (institution competency id) to "CCRS-1.4" and re-uploaded and it had instead been the external_id value, then the platform would have no way of knowing what tag is supposed to get updated to "CCRS-1.4". 

This is a problem whether you use `code` or `external_id`. But I would say generally institutions shouldn't be changing the main ID ("code" in this example) of a tag at all. If they do, there are workarounds though, if we make the field mutable.

jesperhodge:
@bradenmacdonald I just updated the old PR for `external_id`s right now so that it is up to date and can be looked at and perhaps merged: https://github.com/openedx/openedx-core/pull/491
(I tagged you as reviewer)
-> but the problem described in the ADR is real, as changing external_id breaks the import-export workflow.
Making `external_id` mutable conflicts with this.
@ormsbee 's idea to store previous external_ids is a good solution to this, making external_id mutability okay, though it takes some finessing to make this work well with exports. (I think it would need a good amount of effort.)
@mgwozdz-unicon it is true that external_id is there for this purpose of syncing to an external system. What I don't understand: why does it need to be mutable? Is there an external_id for the imported CSV that points to something different than the tag code you want to add? Otherwise, why isn't the tag code added as external_id before importing the CSV? Is there no workaround to keep it immutable?

bradenmacdonald:
> @ormsbee 's idea to store previous external_ids is a good solution to this, making external_id mutability okay, though it takes some finessing to make this work well with exports.

Do we need to include it on exports? I would think that `previous_external_ids` could be a field that's only used on imports, in the rare case of changing tag IDs, and not something we need to store or export. The goal of that field is to trigger a [`RenameTag`](https://github.com/openedx/openedx-core/blob/e6375c44ab3e38e86f37d51551f7b6ec01b067f8/src/openedx_tagging/import_export/actions.py#L319) in the import plan rather than a `DeleteTag` + `AddTag`.

I don't think our taxonomy exports need to include history or "previous IDs"; they don't even include any taxonomy metadata at the moment. In any case, there is an official taxonomy published somewhere else, which likely includes a proper changelog ([example](https://docs.lightcast.dev/updates/skills-taxonomy-changelog)), so we don't need to make a poor-man's version of that. Just let people use the official upstream dataset if they need the full picture and history etc.

jesperhodge:
> I would think that previous_external_ids could be a field that's only used on imports, in the rare case of changing tag IDs, and not something we need to store or export. The goal of that field is to trigger a [RenameTag](https://github.com/openedx/openedx-core/blob/e6375c44ab3e38e86f37d51551f7b6ec01b067f8/src/openedx_tagging/import_export/actions.py#L319) in the import plan rather than a DeleteTag + AddTag.

That sounds right to me, I should have referred to the import and not the export. And the solution you put here does seem much simpler than I thought.

ormsbee:
I agree that we don't need to persist the `previous_id` data in the database, or the export, and that it would only be read on import.

We might want to eventually persist a changelog, since we already break it down into ImportAction primitives anyway (e.g. one row per import, with a JSON representation of the actions that were taken).

ormsbee:
Misc. consideration for testing (does not need to be part of any ADR): One edge case would be an `external_id` swap, e.g. "A" -> "B" and "B" -> "A" in the same import. Renames might require a two step process, where all tags-to-be-renamed are first renamed to something random before being reset to their final state `external_id` values.

mgwozdz-unicon:
I've revised the ADR based on this discussion plus a follow-up team conversation. Summary: dropped the new `Tag.code` field entirely. `Tag.external_id` becomes the institution's editable competency identifier instead, with mutability scoped to a new `previous_id` import column for now (Taxonomy Editing UI and the generic tag API stay unchanged, that's future-phase work).

To the open questions:

**@ormsbee** — yes, exactly what you proposed. A `previous_id` column in the import file, matched against an existing tag's `external_id`, remaps that tag in place (updates its `external_id` to the new `id`) instead of deleting and recreating it. That removes the need for `code` entirely, since `external_id` is now allowed to change through that path.

**@kdmccormick** — renaming the identifier is a confirmed use case, not hypothetical: institutions do this when they revise a competency taxonomy (e.g. after a curriculum or accreditation-standard update). They renumber the existing tags in place rather than standing up a new taxonomy for the same competencies, so we need the rename to survive without losing each tag's associations (like its `CompetencyCriteria` links) or its import/export identity.

**@jesperhodge** — on why it needs to be mutable at all: the alternative is treating every identifier change as delete-and-recreate, which is exactly what loses those associations and the tag's identity across import/export. That's the "keep external_id strictly immutable" alternative in the revised ADR, and it's rejected for that reason. Also, to your and Braden's resolved point: `previous_id` is import-only, not persisted, and doesn't need to show up in exports.

**@jesperhodge** on `CompetencyMetadata`: not adding it at this time. There's no competency-specific tag metadata to justify a separate table right now, since this ADR's identifier need is fully handled by `external_id`. If a descriptive field (tag description, competency statement) is needed later, I'd want it directly on `Tag` as a generic field rather than gated behind a competency-specific table, since non-competency taxonomies could reasonably want descriptions too. Happy to revisit if a real competency-specific field shows up.

ormsbee (approved):

kdmccormick (approved):
Thanks!

bradenmacdonald (approved):

PR #656, opened 2026-07-10
docs: add ADR for cross-instance taxonomy identity

mgwozdz-unicon:
- Adds ADR 0011, proposing a stable, immutable `uuid` field on `Taxonomy` so that competency
  criteria copied along with course content (new course run, course export/import, course/library
  copy) can be bound by reference to the correct taxonomy on the target instance, rather than an
  export_id match that isn't reliable across instances.
- Defines the reconciliation policy for repeat imports: if the incoming tag diff is additions
  only, apply it; if it includes any rename, reparent, or delete, refuse the import rather than
  risk silently mutating taxonomy state that unrelated content on the target depends on.
- Scopes this decision to the legacy XBlock/modulestore course path only. Cross-instance library
  copy, the newer `openedx_content` Component/Container model, and full taxonomy/tag versioning
  are explicitly out of scope, noted as excluded or deferred with reasoning.

### DIFF
--- a/docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
+++ b/docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
@@ -11,13 +11,23 @@ Proposed
 Context
 -------
 
-Competency Based Education (CBE) will bring about the use case where, when course content is
-copied by any of the three supported mechanisms, a new course run, course export/import, or
-course/library copy, any competency criteria attached to that content should be copied along with
-it. Competency criteria (``CompetencyCriteria``, ``CompetencyCriteriaGroup``) are versioned via
-``django-simple-history`` per :ref:`openedx-learning-adr-0003`, but the taxonomies and tags they
-reference are deliberately non-evaluative, unversioned display metadata. Copying a criterion
-therefore also means resolving the taxonomy and tags it points to on the target.
+Competency Based Education (CBE) introduces ``CompetencyCriteria``: see
+:ref:`openedx-learning-adr-0002` for the full model, but in brief, a competency criterion asserts
+that a piece of course content demonstrates a specific competency, which it identifies by pointing
+at a Tag in a Taxonomy. Criteria are grouped into a ``CompetencyCriteriaGroup`` tree, also defined
+in ADR 0002, that combines them with AND/OR logic and may be scoped to a specific course run
+(``CourseRun`` in ``openedx_catalog``, identified by the same ``course_key`` that identifies the
+corresponding legacy modulestore course; "course" throughout this decision means that same
+course-run identity, the one ADR 0002's ``course_id`` references).
+
+When course content is copied by any of the three supported mechanisms, a new course run, course
+export/import, or course/library copy, any competency criteria attached to that content should be
+copied along with it. Criteria themselves are versioned via ``django-simple-history`` per
+:ref:`openedx-learning-adr-0003`, but the taxonomy and tags they reference are deliberately not:
+they're non-evaluative display metadata that can change independently without creating a new
+version of the criteria pointing at them. Copying a criterion therefore also means resolving the
+taxonomy and tags it points to on the target, since the copy has to have a corresponding tag for it
+to reference.
 
 No stable, cross-instance identity exists for a ``Tag`` today: ``Tag.external_id`` is an editable,
 instance-scoped identifier designed for import-file bookkeeping (see :ref:`openedx-tagging-adr-0006`).
@@ -30,8 +40,10 @@ The three copy mechanisms differ significantly in current maturity:
 
 - **Course export/import** (legacy XBlock/modulestore course): already works, via a ``tags.csv``
   sibling file resolved by ``export_id`` at import time.
-- **New course run**: always same-instance/same-database. ``copy_tags()`` exists in
-  ``openedx_tagging.api`` but is not wired into the course-rerun flow.
+- **New course run**: always same-instance/same-database, so no taxonomy/tag resolution is ever
+  needed; what's missing is recreating the ``ObjectTag``, ``CompetencyCriteria``, and
+  ``CompetencyCriteriaGroup`` rows for the new run (see Copy semantics, below). ``copy_tags()``
+  exists in ``openedx_tagging.api`` but is not wired into the course-rerun flow yet.
 - **Library copy**: always same-instance today; no cross-instance transport exists in
   ``content_libraries``, and none is planned.
 
@@ -74,13 +86,29 @@ no reconciliation applies to them.
 Copy semantics
 ~~~~~~~~~~~~~~
 
-Competency criteria are copied **by reference**: the target's criteria are bound to a taxonomy
-sharing the source's ``export_id``, not to an independent duplicate. This is a larger commitment than a
-by-value copy, but a by-value copy would leave the target's competency evaluation permanently
-disconnected from the taxonomy it depends on, undermining the goal of this use case.
+Competency criteria are recreated on the target, not moved. For any of the three mechanisms, the
+``ObjectTag``, ``CompetencyCriteria``, and ``CompetencyCriteriaGroup`` rows attached to the copied
+content are duplicated and re-keyed to the new object and course ids, using an old-id-to-new-id
+mapping built during the copy. A recreated ``CompetencyCriteriaGroup`` shares a parent group with
+the original, so the two are combined by ``OR`` by default.
+
+The ``Taxonomy``/``Tag`` rows a recreated ``ObjectTag`` points to are handled differently: they are
+never duplicated, only referenced, and it's this taxonomy/tag relationship the rest of this
+decision means by **by reference**. For a same-instance mechanism (new course run, library copy),
+source and target already share the identical taxonomy row, so no resolution is needed at all. For
+a cross-instance mechanism (course export/import), the reference is resolved to whichever taxonomy
+on the target shares the source's ``export_id``, per Resolution on import, below. This is a larger
+commitment than a by-value taxonomy copy, but a by-value copy would leave the target's competency
+evaluation permanently disconnected from the taxonomy it depends on, undermining the goal of this
+use case.
 
 Resolution on import
 ~~~~~~~~~~~~~~~~~~~~~
+
+This section, and Reconciliation on repeat import, below, apply specifically to course
+export/import: it is the only mechanism where a matching taxonomy might not already exist on the
+target. New course run and library copy need no taxonomy-side resolution at all, per Copy
+semantics, above.
 
 On import, whether the source and target are the same deployment or two different organizations'
 instances, the behavior is uniform:
--- a/docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
+++ b/docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
@@ -19,10 +19,12 @@ it. Competency criteria (``CompetencyCriteria``, ``CompetencyCriteriaGroup``) ar
 reference are deliberately non-evaluative, unversioned display metadata. Copying a criterion
 therefore also means resolving the taxonomy and tags it points to on the target.
 
-No stable, cross-instance identity exists for a ``Taxonomy`` or ``Tag`` today. ``Taxonomy.export_id``
-and ``Tag.external_id`` are both editable, instance-scoped identifiers designed for import-file
-bookkeeping (see :ref:`openedx-tagging-adr-0006`), not for answering "does the target already have
-this taxonomy."
+No stable, cross-instance identity exists for a ``Tag`` today: ``Tag.external_id`` is an editable,
+instance-scoped identifier designed for import-file bookkeeping (see :ref:`openedx-tagging-adr-0006`).
+``Taxonomy.export_id`` is different: per `modular-learning#183
+<https://github.com/openedx/modular-learning/issues/183>`_, it was always intended to answer "does
+the target already have this taxonomy," which is exactly what this decision needs; it just hasn't
+been documented as such, or exercised for identity-matching purposes, until now.
 
 The three copy mechanisms differ significantly in current maturity:
 
@@ -47,16 +49,18 @@ Decision
 Stable identity
 ~~~~~~~~~~~~~~~~
 
-Add an immutable ``uuid`` field to ``Taxonomy``, generated at creation and preserved through
-export/import. ``export_id`` cannot serve this purpose on its own: it is free text, chosen and
-editable by whoever administers a taxonomy, so nothing guarantees that the same ``export_id`` on
-two different instances refers to the same taxonomy, or that two different taxonomies on two
-instances never happen to share one. A ``uuid`` is generated once, never touched by a person, and
-so cannot collide or drift the way a human-chosen identifier can. ``export_id`` keeps its existing
-role as the human-meaningful identifier; ``uuid`` adds the machine identity it was never designed
-to provide. This follows the existing convention in this codebase of using ``uuid`` for a stable
-external reference (see ``PublishableEntity.uuid``, :ref:`openedx-content-adr-0003`), applied here
-to a model that currently lacks it.
+Use the existing ``Taxonomy.export_id`` field as the cross-instance identity, rather than adding a
+new field. ``export_id`` is already required, unique, and format-validated
+(``^[\w\-.]+$``) at the model level, and the REST API already accepts a caller-supplied value on
+taxonomy creation. Two institutions that each set up the same third-party taxonomy (for example,
+Lightcast Open Skills) can establish that they're the same taxonomy simply by using the same
+``export_id`` (for example, a reverse-DNS-style value like ``io.lightcast.open-skills``), something
+an immutable, randomly-generated identifier could never let them do, since two independent imports
+would always get two different random values with no way to reconcile them afterward.
+
+``export_id``'s existing mutability (it can be edited after creation) is also a feature here rather
+than a gap: it is how the deferred manual-merge case below would actually be performed, by editing
+one instance's ``export_id`` to match the other's.
 
 ``Tag`` does not need a new identifier: ``Tag.external_id`` (already used by the tag import/export
 plan-building logic, see :ref:`openedx-tagging-adr-0006`) is sufficient for within-taxonomy
@@ -67,7 +71,7 @@ Copy semantics
 ~~~~~~~~~~~~~~
 
 Competency criteria are copied **by reference**: the target's criteria are bound to a taxonomy
-sharing the source's ``uuid``, not to an independent duplicate. This is a larger commitment than a
+sharing the source's ``export_id``, not to an independent duplicate. This is a larger commitment than a
 by-value copy, but a by-value copy would leave the target's competency evaluation permanently
 disconnected from the taxonomy it depends on, undermining the goal of this use case.
 
@@ -77,9 +81,9 @@ Resolution on import
 On import, whether the source and target are the same deployment or two different organizations'
 instances, the behavior is uniform:
 
-- If no taxonomy with a matching ``uuid`` exists on the target, auto-create one, seeded from the
+- If no taxonomy with a matching ``export_id`` exists on the target, auto-create one, seeded from the
   tags that traveled with the export.
-- If a taxonomy with a matching ``uuid`` already exists, reconcile it (see below) rather than
+- If a taxonomy with a matching ``export_id`` already exists, reconcile it (see below) rather than
   creating a duplicate.
 
 A single uniform rule was chosen over branching by deployment relationship because the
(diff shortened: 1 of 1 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,191 @@
...
+copied by any of the three supported mechanisms, a new course run, course export/import, or
+course/library copy, any competency criteria attached to that content should be copied along with
+it. Competency criteria (``CompetencyCriteria``, ``CompetencyCriteriaGroup``) are versioned via
+``django-simple-history`` per :ref:`openedx-learning-adr-0003`, but the taxonomies and tags they
+reference are deliberately non-evaluative, unversioned display metadata. Copying a criterion
+therefore also means resolving the taxonomy and tags it points to on the target.
+
+No stable, cross-instance identity exists for a ``Taxonomy`` or ``Tag`` today. ``Taxonomy.export_id``
+and ``Tag.external_id`` are both editable, instance-scoped identifiers designed for import-file
+bookkeeping (see :ref:`openedx-tagging-adr-0006`), not for answering "does the target already have
```
bradenmacdonald: Hmm, I guess it's not documented clearly but the goal of `export_id` is actually to identify taxonomies across different instances, exactly what you are requesting with this ADR. It is meant for answering "does the target already have this taxonomy."

See https://github.com/openedx/modular-learning/issues/183

> "As a content author, I want a consistent way to identify taxonomies, _especially across instances_".

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,191 @@
...
+Add an immutable ``uuid`` field to ``Taxonomy``, generated at creation and preserved through
+export/import. ``export_id`` cannot serve this purpose on its own: it is free text, chosen and
+editable by whoever administers a taxonomy, so nothing guarantees that the same ``export_id`` on
+two different instances refers to the same taxonomy, or that two different taxonomies on two
+instances never happen to share one. A ``uuid`` is generated once, never touched by a person, and
+so cannot collide or drift the way a human-chosen identifier can. ``export_id`` keeps its existing
+role as the human-meaningful identifier; ``uuid`` adds the machine identity it was never designed
+to provide. This follows the existing convention in this codebase of using ``uuid`` for a stable
+external reference (see ``PublishableEntity.uuid``, :ref:`openedx-content-adr-0003`), applied here
+to a model that currently lacks it.
```
bradenmacdonald: I don't think a UUID is what we want.

If I set up the Lightcast Open Skills taxonomy on instance A, and someone else sets up the Lightcast Open Skills taxononomy on instance B, there should be some way to indicate that they're the same, and future exports/imports across the instances will handle that. But if they each get an immutable UUID at the time of original import, then there is no way to achieve that. Whereas with export ID, it's easy, you just make sure they both have the same export ID like `io.lightcast.open-skills`.

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,191 @@
...
+is enough to satisfy this use case safely; worth revisiting if refusals turn out to be common in
+practice.
+
+Real versioning for taxonomies and tags
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``Taxonomy``/``Tag`` could gain version history (e.g. via ``django-simple-history``, matching
+``CompetencyCriteria``), which would let drift between source and target be tracked precisely
+rather than detected only as additive or not. Likely needed eventually, but not required for this
+use case, since the additions-only reconciliation policy already satisfies it without history.
```
bradenmacdonald: Please link to https://github.com/openedx/openedx-core/issues/455 which is our tentative plan for making taxonomies versioned.

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,191 @@
...
+instances, without ever having copied content between each other. Since a ``uuid`` establishes
+identity only once two taxonomies have actually shared a copy operation, importing between them
+today would create a second, unrelated taxonomy on the receiving side, not recognize them as the
+same one. Automatic matching cannot safely resolve this: falling back to matching by name or
+content similarity would reintroduce the false-positive risk ``uuid`` was introduced to avoid, two
+instances with genuinely different taxonomies that happen to look similar would be silently
+merged. The eventual resolution is a deliberate, human-initiated action: an administrator manually
+reassigns one instance's taxonomy to adopt the other's ``uuid``, retroactively establishing shared
+identity going forward. Not designed here, since it is a distinct, rare operation, orthogonal to
+the copy-time behavior this decision covers.
```
bradenmacdonald: This is what I was saying above, but it can be done more easily via `export_id`. It's awkward to compare and edit UUIDs.

bradenmacdonald (commented):
I'm not totally opposed to using UUIDs here, but I don't think it's necessary and I do think you can achieve everything that you want by making an effort to make the default value of `export_id` "more unique" than it currently is.

mgwozdz-unicon:
Thanks @bradenmacdonald , this makes sense, went with `export_id` in the latest commit rather than adding a new `uuid` field.

You're right that `export_id` already has what's needed here: it's required, unique, and format-validated at the model level, and the REST API already accepts a caller-supplied value on create. Linked modular-learning#183 in the Context section so the intended purpose is documented where the next person will look for it.

Additionally:

- Added "Add a new immutable `uuid` field" to Rejected Alternatives, so the reasoning for not going that route is on record.
- The "Manual merge of independently-created taxonomies" deferred item now notes that today this requires a direct API call: Studio's Taxonomy Editing UI has no field to set or edit `export_id`.
- Linked openedx-core#455 in the versioning section per your comment there.

Let me know if I've missed anything.

ormsbee (commented):
I agree with @bradenmacdonald's review, and using `export_id`. It's similar to what we already do for things like courses and libraries, where we rely on meaningful keys.

One thing I want flag is that courses and libraries are implicitly namespaced to an org, which is something that we're not currently doing with taxonomies. We don't really have system vs. org-level libraries or courses. If there is ever a situation in which multiple orgs want to independently tweak or update the same logical taxonomy, we may want to consider some kind of namespacing conventions.

mgwozdz-unicon:
> One thing I want flag is that courses and libraries are implicitly namespaced to an org, which is something that we're not currently doing with taxonomies. We don't really have system vs. org-level libraries or courses. If there is ever a situation in which multiple orgs want to independently tweak or update the same logical taxonomy, we may want to consider some kind of namespacing conventions.

I added a note about this to the "Excluded from this decision" section to flag it for the future.

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,219 @@
...
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+Competency Based Education (CBE) will bring about the use case where, when course content is
```
jesperhodge: Before this ADR uses "competency criteria," can we get one plain sentence defining it? Something like: "A competency criterion asserts that a piece of course content demonstrates a specific competency, which it identifies by pointing at a tag in a taxonomy." Right now CompetencyCriteria/CompetencyCriteriaGroup show up in parentheses on line 17 with zero setup, and this ADR lives in openedx_tagging, which explicitly isn't supposed to know about competencies. A reader coming from the tagging side needs to know why a criterion touches a taxonomy at all before the rest of the Context makes sense.

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,219 @@
...
+copied by any of the three supported mechanisms, a new course run, course export/import, or
+course/library copy, any competency criteria attached to that content should be copied along with
+it. Competency criteria (``CompetencyCriteria``, ``CompetencyCriteriaGroup``) are versioned via
+``django-simple-history`` per :ref:`openedx-learning-adr-0003`, but the taxonomies and tags they
+reference are deliberately non-evaluative, unversioned display metadata. Copying a criterion
+therefore also means resolving the taxonomy and tags it points to on the target.
```
jesperhodge:
```suggestion
it. Each competency criterion (CompetencyCriteria) identifies the competency it evaluates by referencing a specific Tag in a Taxonomy. Criteria themselves are versioned via django-simple-history (see :ref:openedx-learning-adr-0003), but the taxonomy and tags they reference are deliberately not: they're non-evaluative display metadata that can change independently without creating a new version of the criteria pointing at them. Copying a criterion therefore also means resolving the taxonomy and tags it points to on the target, since the copy has to land somewhere with a matching tag for it to reference.
```

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,219 @@
...
+
+``Tag`` does not need a new identifier: ``Tag.external_id`` (already used by the tag import/export
+plan-building logic, see :ref:`openedx-tagging-adr-0006`) is sufficient for within-taxonomy
+matching. Free-text taxonomies have no ``Tag`` rows and travel as literal strings unconditionally;
+no reconciliation applies to them.
+
+Copy semantics
+~~~~~~~~~~~~~~
+
+Competency criteria are copied **by reference**: the target's criteria are bound to a taxonomy
```
jesperhodge: Does this actually work?
- A `CompetencyCriteria` is indeed bound to a tag application that is bound to an object and a taxonomy, where the object is an object in e.g. the course that's being copied. Does a course copy copy the taxonomy, and do the tags resolve so that the object tag ids actually apply to the new taxonomy_id that has the same export_id?
- A `CompetencyCriteria` is also bound to a `CompetencyCriteriaGroup`. How is this `CompetencyCriteriaGroup` copied to the new course or content?
- Do I understand **by reference** to mean a reference of the criteria to the export_id of a taxonomy? Or by reference to the competency criteria? In that case I'm confused who references them and how the reference could be resolved if we have a different environment with a different database.
- A `CompetencyCriteriaGroup` is optionally bound to a `course_id`. I assume that is a course run. I assume it actually has to have the course_id to apply to a course. Since a `CompetencyCriteria` belongs to a `CompetencyCriteriaGroup` that belongs to a course, wouldn't copying it by reference mean that it does not apply to the new course now?

jesperhodge (changes requested):
Some clarifications needed :)

jesperhodge:
Claude flagged this:
"One inconsistency worth flagging for the team: ADR 0002 says course_id FKs to openedx_catalog_courserun, while ADR 0011 frames the near-term copy target as the legacy modulestore course. The two ADRs describe different course-model layers without reconciling them."

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
jesperhodge: To clarify my point 1:
An object_tag encodes a course run in the object_id. So if we do a course re-run, for example, the object_tags reference the previous course run, which means that the new course has no tags. When tags are copied to the new course, the copy of the tags somehow needs to be resolved to the new object_ids of the course content, not the same as previously. I assume this is handled via this `tags` file in course imports - and we will now apply the same logic to re-runs and such - but can you confirm this assumption is true?

bradenmacdonald: I am not sure if this is handled at all in course re-runs today. (But it should be.) Some testing will be required.

mgwozdz-unicon:
@jesperhodge, addressed your comments:

1. Added a plain-sentence definition of a competency criterion to the Context section, and pointed to ADR 0002 for the full model. This ADR still only owns taxonomy/tag identity, not the competency model itself.
2. Took your suggested wording for that paragraph, with minor tweaks.
3. This was a fair question, "by reference" was doing double duty for two different things. Now split apart in Copy semantics:
   - `Taxonomy`/`Tag` are never duplicated, for any of the three mechanisms. That's the only thing "by reference" describes now: same-instance mechanisms (new course run, library copy) already share the identical taxonomy row, so there's nothing to resolve; cross-instance (course export/import) resolves the reference via `export_id`, per the existing Resolution on import section.
   - `ObjectTag`, `CompetencyCriteria`, and `CompetencyCriteriaGroup` are copied by value: duplicated and re-keyed to the new object/course ids, using an old-id-to-new-id mapping built during the copy. A new `CompetencyCriteriaGroup` shares a parent group with the original (combined by OR by default).
   - So to your bullet 4: yes, a `CompetencyCriteriaGroup` copied this way does end up with the new course_id, since it's a new row, not the old one repointed.
4. Confirmed by checking the actual rerun code: `ObjectTag` copying is not implemented for course reruns today, only for unrelated flows (block duplicate/paste, library-to-course upstream sync). The Context section now says this plainly rather than implying `copy_tags()` just needs wiring in. Building this is new platform work; agreed it's a scope conversation for the team rather than something to spell out in the ADR.

On the course_id inconsistency you flagged separately: checked both ADRs against the actual `CourseRun` model. It's not actually two competing identifier systems, `CourseRun.course_key` (openedx_catalog) is the same modulestore course key ADR 0011 already talks about; the model's docstring says it's meant to be the canonical FK target for exactly this kind of reference. ADR 0011 just never said so. Added a sentence in Context tying "course" explicitly to `CourseRun`, so it's clear this is the same course-run identity ADR 0002's `course_id` references, not a different one.

@bradenmacdonald your suspicion on the rerun/ObjectTag question was right, confirmed above.

bradenmacdonald:
@mgwozdz-unicon @jesperhodge Note that there _is_ code [to duplicate tags when an XBlock is duplicated](https://github.com/openedx/openedx-platform/blob/0f57cf2bbf13358af5343f0fc2ddd952b656d570/openedx/core/djangoapps/content_tagging/handlers.py#L115-L128) but reruns don't involve the `XBLOCK_DUPLICATED` event.


We probably need to add a similar handler for the [`COURSE_RERUN_COMPLETED` event](https://github.com/openedx/openedx-events/blob/52d385c771893aa761a3dde15d23b25302770450/openedx_events/content_authoring/signals.py#L361-L372).

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+~~~~~~~~~~~~~~~~
+
+Use the existing ``Taxonomy.export_id`` field as the cross-instance identity, rather than adding a
+new field. ``export_id`` is already required, unique, and format-validated
+(``^[\w\-.]+$``) at the model level, and the REST API already accepts a caller-supplied value on
+taxonomy creation. Two institutions that each set up the same third-party taxonomy (for example,
+Lightcast Open Skills) can establish that they're the same taxonomy simply by using the same
+``export_id`` (for example, a reverse-DNS-style value like ``io.lightcast.open-skills``), something
+an immutable, randomly-generated identifier could never let them do, since two independent imports
+would always get two different random values with no way to reconcile them afterward.
```
jesperhodge:
```suggestion
Use the existing ``Taxonomy.export_id`` field as the cross-instance identity across multiple institutions and external systems, rather than adding a
new field.
```

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+
+``export_id``'s existing mutability (it can be edited after creation) is also a feature here rather
+than a gap: it is how the deferred manual-merge case below would actually be performed, by editing
+one instance's ``export_id`` to match the other's.
+
+``Tag`` does not need a new identifier: ``Tag.external_id`` (already used by the tag import/export
+plan-building logic, see :ref:`openedx-tagging-adr-0006`) is sufficient for within-taxonomy
+matching. Free-text taxonomies have no ``Tag`` rows and travel as literal strings unconditionally;
+no reconciliation applies to them.
```
jesperhodge:
```suggestion
The existing ``Tag.external_id`` is sufficient for within-taxonomy matching.
```

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+(``^[\w\-.]+$``) at the model level, and the REST API already accepts a caller-supplied value on
+taxonomy creation. Two institutions that each set up the same third-party taxonomy (for example,
+Lightcast Open Skills) can establish that they're the same taxonomy simply by using the same
+``export_id`` (for example, a reverse-DNS-style value like ``io.lightcast.open-skills``), something
+an immutable, randomly-generated identifier could never let them do, since two independent imports
+would always get two different random values with no way to reconcile them afterward.
+
+``export_id``'s existing mutability (it can be edited after creation) is also a feature here rather
+than a gap: it is how the deferred manual-merge case below would actually be performed, by editing
+one instance's ``export_id`` to match the other's.
```
jesperhodge:
```suggestion
``export_id``'s remains mutable and supports the manual-merge case below by editing
one instance's ``export_id`` to match the other's.
```

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+
+The ``Taxonomy``/``Tag`` rows a recreated ``ObjectTag`` points to are handled differently: they are
+never duplicated, only referenced, and it's this taxonomy/tag relationship the rest of this
+decision means by **by reference**. For a same-instance mechanism (new course run, library copy),
+source and target already share the identical taxonomy row, so no resolution is needed at all. For
+a cross-instance mechanism (course export/import), the reference is resolved to whichever taxonomy
+on the target shares the source's ``export_id``, per Resolution on import, below. This is a larger
+commitment than a by-value taxonomy copy, but a by-value copy would leave the target's competency
+evaluation permanently disconnected from the taxonomy it depends on, undermining the goal of this
+use case.
```
jesperhodge:
```suggestion
The ``Taxonomy``/``Tag`` rows a recreated ``ObjectTag`` points to are handled differently: they are
never duplicated, only referenced, and it's this taxonomy/tag relationship the rest of this
decision means by **by reference**.
```
Since resolution on import is already outlined below, we don't need it here again.

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+
+Resolution on import
```
jesperhodge:
```suggestion
Resolution on course export/import
```

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+
+This section, and Reconciliation on repeat import, below, apply specifically to course
+export/import: it is the only mechanism where a matching taxonomy might not already exist on the
+target. New course run and library copy need no taxonomy-side resolution at all, per Copy
+semantics, above.
+
```
jesperhodge:
```suggestion
```
No need to list everything that is not affected.

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+On import, whether the source and target are the same deployment or two different organizations'
```
jesperhodge:
```suggestion
On course import, whether the source and target are the same deployment or two different organizations'
```

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+instances, the behavior is uniform:
+
+- If no taxonomy with a matching ``export_id`` exists on the target, auto-create one, seeded from the
+  tags that traveled with the export.
+- If a taxonomy with a matching ``export_id`` already exists, reconcile it (see below) rather than
+  creating a duplicate.
+
+A single uniform rule was chosen over branching by deployment relationship because the
+reconciliation policy below already guards against silent corruption in both cases; adding a
+second behavior for the cross-organization case would add complexity without removing risk.
```
jesperhodge: If necessary, this can explained in the "rejected alternatives" section, or it can just be dropped.
```suggestion
```

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+  to that taxonomy.
+
+The actor triggering a course copy has no standing to mutate a taxonomy shared with other,
+unrelated content on the target instance. In a manual tag re-import (the existing use of this
+plan-building logic), the file represents the deliberate intent of whoever owns that taxonomy. A
+copied course's snapshot only reflects what the source looked like at export time: a tag missing
+from it doesn't mean the source deleted it, it may be something the target added independently.
+Auto-applying deletes or renames on that basis risks silently corrupting taxonomy state that
+unrelated courses on the target depend on. Additions carry no such risk: they only ever add new
+identities, never touch existing ones.
```
jesperhodge:
```suggestion
```
Probably not needed.

docs/openedx_tagging/decisions/0011-cross-instance-taxonomy-identity.rst
```diff
@@ -0,0 +1,247 @@
...
+
+Failure surfacing
+~~~~~~~~~~~~~~~~~~
+
+A refusal is modeled as an ordinary import task failure, using the existing
+``UserTaskStatus.fail(message)`` / ``Error`` artifact / ``import_status_handler`` mechanism in
+``contentstore``. The Studio Authoring MFE's import flow already reads this same ``Message`` field
+(``CourseImportContext.tsx``), so this requires no new frontend or backend surface, and no
+distinction needs to be drawn between a Platform Administrator and a Course Author watching the
+same import: whoever is watching sees the existing failure message.
```
jesperhodge:
```suggestion
A refusal is modeled as an ordinary import task failure.
```
We can probably drop the rest as implementation details, especially since those mechanics already exist on edx-platform.

jesperhodge (approved):
Looks very good! No reservations from my side, this looks ready.

Feel free to reject all the suggestions! They are completely optional.

As we have been looking at making ADRs relatively concise and skip implementation detail, I've just gone through this and looked for places that could be shortened or where we can skip implementation details as such.

This is not meant as "please improve this", but instead just as inspiration for you on how we can make this ADR more concise, if desired. Shortening technical documents is hard, so I thought I just see what suggestions may be helpful to you.

ormsbee (approved):

bradenmacdonald (approved):

PR #657, opened 2026-07-10
Competency Status Storage ADR

jesperhodge:
UPDATE: ADRs in this PR have been completely rewritten as of July 17, 2026.

Strong involvement from Claude (AI).

I haven't fully reviewed changes to diagrams and ADRs 2 and 3 yet, made with Claude, need my own review still.

## Summary

These are some significant changes to the data model and approach. They will warrant a good amount of discussion.
A main concern is that the leaf node table and leaf node history table could possibly contain billions of rows, at the scale of the `PersistentSubsectionGrade` table of edx-platform. The current solution accepts this, but is this acceptable for OpenEdx? If not, there is also a rejected alternative to not store leaf nodes and just compute them at read time, but that would make it harder to keep subsection statuses frozen so that later competency criteria changes don't change the status.

ADR 4: How learner competency mastery is recorded
correctly under concurrent, out-of-order grade-change events without
paying a per-event serialization cost at scale.

ADR 5: How competency criteria statuses for learners are stored at scale, given the tables
can be massive (billions of rows).


openedx-webhooks:
Thanks for the pull request, @jesperhodge!


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

mgwozdz-unicon (changes requested):
Following up on feedback I'd shared earlier outside this PR favoring the batch-lock approach over keyed partitioning, given the Kafka/Redis dependency and the accuracy-over-latency tradeoff for this use case: this revision reflects that direction, and the Kafka/Redis coupling concern is resolved.

I'm wondering if we should also consider a per-learner lock instead of the single deployment-wide lock. A DB advisory lock keyed on a hash of `user_id` would give the same same-learner serialization the chosen approach relies on, but would let different learners' batches run in parallel, which the current design gives up entirely. It also sidesteps everything Alternative 1 was rejected for: no event-transport dependency, no partition-key contract with openedx-platform, no reconciliation backstop.

Rejected Alternative 2's argument doesn't quite cover this: it treats "per-learner isolation" as equivalent to "keyed partitioning" ("Making parallel evaluation correct requires... per-learner isolation, i.e. the keyed-partitioning alternative above"), but a per-learner DB lock gets you per-learner isolation without touching the transport at all. If there's a reason this doesn't hold up here (lock-management overhead across many concurrent per-learner locks, contention patterns, something else), that reasoning is worth adding to the doc.

Requesting changes to get this addressed, either by adding the analysis to Rejected Alternatives or by folding it into the Decision.

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
```diff
@@ -0,0 +1,157 @@
...
+
+Two forces shape how this recording should happen:
+
+- **Same-learner correctness.** Grade-change events arrive asynchronously and can be delivered out
+  of order and processed on more than one worker. Because writes are append-only, two evaluations
+  for the same learner that overlap can each read a stale snapshot of the sibling leaf statuses and
+  each append a derived roll-up computed from an incomplete picture (a write-skew). Leaf rows are
+  always correct, since each leaf is a pure function of its own grade; only the derived
+  group/competency rows can be left wrong. Nothing crashes and no constraint is violated, but a
+  learner's stored competency status can be silently incorrect.
```
ormsbee: Maybe this is a silly question, but can't we mitigate this by making sure to commit the transaction of the leaf statuses, and then re-read the latest commited state of the leaves before doing the roll-up?

ormsbee: Another thing we can do (if it's necessary) is to arrange it so that the roll-ups are a log that have an incrementing version number, and create a composite unique index that would cause a conflict that would reject concurrent writes. The losing write would have to catch the error, increment the version number again, and re-read the database state.

jesperhodge:
> Following up on feedback I'd shared earlier outside this PR favoring the batch-lock approach over keyed partitioning, given the Kafka/Redis dependency and the accuracy-over-latency tradeoff for this use case: this revision reflects that direction, and the Kafka/Redis coupling concern is resolved.
> 
> I'm wondering if we should also consider a per-learner lock instead of the single deployment-wide lock. A DB advisory lock keyed on a hash of `user_id` would give the same same-learner serialization the chosen approach relies on, but would let different learners' batches run in parallel, which the current design gives up entirely. It also sidesteps everything Alternative 1 was rejected for: no event-transport dependency, no partition-key contract with openedx-platform, no reconciliation backstop.
> 
> Rejected Alternative 2's argument doesn't quite cover this: it treats "per-learner isolation" as equivalent to "keyed partitioning" ("Making parallel evaluation correct requires... per-learner isolation, i.e. the keyed-partitioning alternative above"), but a per-learner DB lock gets you per-learner isolation without touching the transport at all. If there's a reason this doesn't hold up here (lock-management overhead across many concurrent per-learner locks, contention patterns, something else), that reasoning is worth adding to the doc.
> 
> Requesting changes to get this addressed, either by adding the analysis to Rejected Alternatives or by folding it into the Decision.

@mgwozdz-unicon I'll improve the description in the ADR.

Here's why this solution is not good:

It fights the batching that the performance depends on. The chosen design's throughput comes from batching across learners: one bulk read of current statuses, evaluate the whole batch in memory, one bulk write — a few round-trips regardless of how many learners/events are in the batch. A per-learner lock is naturally per-learner (in the earlier design, per-event): acquire lock → read that learner → evaluate → write → release, one learner at a time. That reintroduces per-learner (or per-event) transaction/commit overhead plus lock acquire/release churn, which is exactly the overhead batching was collapsing. Under bursty grading across many learners, that per-unit overhead dominates.

Lock-lifecycle machinery would be multiplied across millions of keys. One deployment-wide lock has exactly one lifecycle to run: acquisition, timeout, stale-lock recovery on a crashed worker. A per-learner lock needs a per-learner lock table (or keyed advisory locks) and that same lifecycle replicated across potentially millions of learner keys, held and waited on concurrently, plus a connection+worker tied up per contended lock.

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
jesperhodge: I think this question is not relevant anymore with the current version, right? Since now we only process on one worker at a time.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -188,34 +217,51 @@ Decision
...
-      1. ``id``: unique primary key
-      2. ``competency_criteria_id``: Foreign key to ``CompetencyCriterion.id``
-      3. ``user_id``: Foreign key pointing to user_id (presumably the learner's id, although it appears that it is possible for staff to get grades as well) in ``auth_user`` table
-      4. ``status_id``: Foreign key to ``CompetencyMasteryStatuses.id``
-      5. ``created``: The timestamp at which the student's criterion status was set.
+      1. ``id``: 64-bit ``BigAutoField`` primary key
+      2. ``competency_criteria_id``: logical foreign key to ``CompetencyCriterion.id`` (no database-level constraint)
+      3. ``user_id``: logical foreign key to the learner's user id (no database-level constraint)
+      4. ``status_id``: foreign key to ``CompetencyMasteryStatuses.id``
+      5. ``effective_source_timestamp``: the timestamp of the grade change this status was computed from, used for the out-of-order defense in :ref:`openedx-learning-adr-0004`
```
ormsbee: How is effective_source_timestamp different from created?

ormsbee (commented):
It seems like this entire ADR's premise is that per-event processing is prohibitively expensive, and we must therefore batch event processing under various locking mechanisms. It's not clear to me why that is. The subsection grade update is already a somewhat expensive, async task. We could plausibly wrap any updates we're doing to competency mastery data in the same transaction that writes the subsection grade update, so that those things are always kept in sync.

Approximately much work are we expecting to happen in order to do the mastery recalculation?

jesperhodge:
> It seems like this entire ADR's premise is that per-event processing is prohibitively expensive, and we must therefore batch event processing under various locking mechanisms. It's not clear to me why that is. The subsection grade update is already a somewhat expensive, async task. We could plausibly wrap any updates we're doing to competency mastery data in the same transaction that writes the subsection grade update, so that those things are always kept in sync.
> 
> Approximately much work are we expecting to happen in order to do the mastery recalculation?

@ormsbee You're right, and it changed my thinking.

I changed the ADR according to this:

Per-event mastery work is actually small, a handful of reads and writes for one learner; the big numbers in the ADR (~200 leaves/course, billions of rows) are platform totals, not the cost of one recalculation. Throughput was never really the issue.

The thing we were actually protecting against is write-skew: two concurrent updates for the same learner can each roll up a status from a stale view of sibling data. But mastery already only moves forward from non-completion to completion, and we can lean on that harder: if every write is "take the higher of stored and computed," concurrent updates converge on their own, no coordination needed. That's the piece that lets us drop the lock, and the throughput ceiling it created goes with it, because now updates can run in parallel.

Which brings us back to your suggestion. It works now, and I don't have a strong preference yet between two shapes:

- A) Recording mastery right inside the grade transaction, as you proposed. Simple and real-time, though it ties mastery's fate to the grade write and assumes they share a database. A tradeoff to point out: mastery work then sits on the synchronous grading path, so a slow mastery query or a bug adds latency to, or rolls back, the grade write (grades are the more critical data).
- B) A signal into openedx-core after the grade write. Keeps mastery decoupled, at the cost of a brief window where they're not yet in sync (which seems fine given mastery is already eventually-consistent by nature).
Batching drops out as a nice-to-have for high-volume bursts rather than something correctness depends on.

Curious which of the two shapes feels right to you, or if there's a third way we're not seeing.

mgwozdz-unicon:
@jesperhodge 
> Which brings us back to your suggestion. It works now, and I don't have a strong preference yet between two shapes:
> 
> * A) Recording mastery right inside the grade transaction, as you proposed. Simple and real-time, though it ties mastery's fate to the grade write and assumes they share a database. A tradeoff to point out: mastery work then sits on the synchronous grading path, so a slow mastery query or a bug adds latency to, or rolls back, the grade write (grades are the more critical data).
> * B) A signal into openedx-core after the grade write. Keeps mastery decoupled, at the cost of a brief window where they're not yet in sync (which seems fine given mastery is already eventually-consistent by nature).
>   Batching drops out as a nice-to-have for high-volume bursts rather than something correctness depends on.
> 
> Curious which of the two shapes feels right to you, or if there's a third way we're not seeing.

I don't think we should go with Option A because we do not intend for mastery to always be tied to a subsection grade. We've actually proposed tying mastery to several other things (course completion, unit grade (a concept that doesn't exist in the platform currently), problem grade, and rubric criterion grade (also a concept that doesn't exist in the platform currently)). So I would go with option B out of these two.

docs/openedx_learning/decisions/0002-competency-criteria-model-diagram.md
```diff
@@ -0,0 +1,150 @@
+# Competency data model (updated)
```
mgwozdz-unicon: Unless others disagree, I think I'd recommend updating the existing PNG or trading it out for the mermaid diagram in the existing ADR 0002, as well as making it clearer that we already supported the active vs history dynamic and how the mastery levels fit in and anything else needed, rather than adding this extra file. We want to try to keep the ADRs light and concise and only change what really needs to be changed. We've also been asked to try to keep the implementation detail in ADRs to a minimum, which may also be relevant here.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -188,34 +217,51 @@ Decision
...
+      6. ``created``: on ACTIVE, when the row was first written; on HISTORY, when this advance was appended
 
-   3. Add a new database table for ``StudentCompetencyCriteriaGroupStatus`` with these columns:
+      The ACTIVE table additionally has a ``modified`` timestamp (last in-place update) and is unique
```
mgwozdz-unicon: I feel like these `modified` fields should get enumerated like the rest of the fields.

ormsbee:
Quick thoughts:

1. We want to keep state in the same transaction as much as it's feasible to do so. Where we diverge from that, it's because operational considerations have _forced_ us to compute it asynchronously in a separate process (e.g. subsection grade calculation).
2. However, in the case where we _are_ doing things asynchronously like grades, if that grade is the record of truth that mastery will be based on, we should trigger mastery recalculation in the same transaction if possible. The same goes for passing a course, getting a Unit grade, etc. -- whenever the record of truth changes for that, we should try to keep things in the same transaction unless we absolutely can't. Correcting grades is a pain, but a well understood one. Adding another place where we can have drift like this is going to cause trouble for operators eventually.
3. My understanding is that moving forward from non-completion to completion can still cause race condition problems, if two separate transactions each would bring a student to right below some cutoff, but the combination would bring the student over. Or if the student is given some kind of score/level of mastery. Given this case, I think it's safest to force race conditions to fail using unique constraints.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
jesperhodge: This is premised on how ADR 0004 is now - it very well might change when we have incorporated your latest feedback; for example, if we integrate everything in a grade update transaction, then we should remove this new field.
I updated the description - it tracks when the grade was changed, not the competency, for async defensiveness per ADR 0004 as it is right now, but this is subject to change.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
jesperhodge: Is it better now?

mgwozdz-unicon:
@ormsbee @jesperhodge 
> Quick thoughts:
> 
> 1. We want to keep state in the same transaction as much as it's feasible to do so. Where we diverge from that, it's because operational considerations have _forced_ us to compute it asynchronously in a separate process (e.g. subsection grade calculation).
> 2. However, in the case where we _are_ doing things asynchronously like grades, if that grade is the record of truth that mastery will be based on, we should trigger mastery recalculation in the same transaction if possible. The same goes for passing a course, getting a Unit grade, etc. -- whenever the record of truth changes for that, we should try to keep things in the same transaction unless we absolutely can't. Correcting grades is a pain, but a well understood one. Adding another place where we can have drift like this is going to cause trouble for operators eventually.
> 3. My understanding is that moving forward from non-completion to completion can still cause race condition problems, if two separate transactions each would bring a student to right below some cutoff, but the combination would bring the student over. Or if the student is given some kind of score/level of mastery. Given this case, I think it's safest to force race conditions to fail using unique constraints.

I could have sworn that there was a reason that we chose to use openedx-events for this process when we last discussed it in January, but I'm having a hard time remembering what it was. My first thought was dependency direction, but that doesn't seem right. I wonder if it was just that we were thinking that CBE would be taking more of a plugin shape so we wanted to keep it decoupled from openedx-platform. Outside of the other cons that Jesper raised for the single transaction (increased latency for the transaction and shared database), I don't have an issue with the single transaction approach if we want to accept those tradeoffs.

Also as a side note: I'm pretty sure openedx-platform and openedx-core share a database already, so the "shared database" point in ADR 4 (line 220) probably needs some clarification.

jesperhodge:
> Also as a side note: I'm pretty sure openedx-platform and openedx-core share a database already, so the "shared database" point in ADR 4 (line 220) probably needs some clarification.

I'll remove it, it's a Claude remnant not needed.

jesperhodge:
@ormsbee @mgwozdz-unicon sounds like we're in alignment, so I'll update the ADR to follow Dave's recommendation.

ormsbee:
Yeah, I was thinking that there's still value in using openedx-events to decouple from platform, without the message bus aspect of it.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,352 @@
...
+      deployment runs everything on one database and gains no new mandatory infrastructure, keeping
+      faith with :ref:`openedx-learning-adr-0004`'s no-new-mandatory-infrastructure value; a large
+      deployment points the alias at a separate physical database purely through settings, with no
+      schema change and no data migration. Building the router now, rather than deferring it, is what
+      makes that later split a configuration change instead of a migration on a billion-row table.
+      Foreign keys that would cross the alias boundary (a learner-status row referencing a
+      criteria-definition row, and the reference to the user) are declared without database-level
+      constraints, as edx-platform does for ``PersistentSubsectionGrade.user_id``, so the tables can
+      live in different databases; the delete protection of :ref:`openedx-learning-adr-0002` is
+      therefore enforced in application code, not by a database constraint.
```
bradenmacdonald: As mentioned in our discussion on https://github.com/openedx/openedx-core/pull/661 , not having foreign keys is definitely a downside of this approach, and making the deletion protection more complicated and fragile.

> as edx-platform does for ``PersistentSubsectionGrade.user_id``, so the tables can live in different databases

Is there an ADR or documentation around `PersistentSubsectionGrade.user_id` ? I can see that it indeed [is not a foreign key](https://github.com/openedx/openedx-platform/blob/daa2d28e1759d6dd810b4e3cf7ebc802a2310962/lms/djangoapps/grades/models.py#L331), but I can't find any explanation of that in the code nor ADRs, so I'm wondering if it is actually meant to be deployed on a separate database or not. As far as I can tell, the only table that has ever been designed to run on a separate database is the `StudentModuleHistoryExtended` table ([it's the only one with a router](https://github.com/openedx/openedx-platform/blob/daa2d28e1759d6dd810b4e3cf7ebc802a2310962/openedx/envs/common.py#L440-L442)).

jesperhodge: @bradenmacdonald @mgwozdz-unicon
I wonder what you think:
- I was trying to model the `StudentModuleHistoryExtended` closely since it already solves the table size issue.
- `StudentCompetencyCriteriaStatus` (the Leaf tables) are the big one. Both ACTIVE and HISTORY tables should be of similar size, which my naive calculations put at worst-case billions of rows for our largest platforms. But could we get away with using the db alias and router for only the history table, if at all? If this is not necessary for the ACTIVE table, we could put a foreign key only on the ACTIVE table, which would be already helpful. I just wonder if there is any similar-sized table on edx-platform that is an example for not using a db router so we know we do not need that optimization?
- I think I was wrong for including `group` and `mastery` here. Those tables (`StudentCompetencyCriteriaGroupStatus`, `CompetencyMasteryStatuses`), as well as their HISTORY tables, are significantly smaller, so I would just use normal foreign keys on them.
- I would love to NOT prematurely optimize here where not necessary. I'm in favor of the most robust and simple solution that is scalable enough. If we can simplify anything here and get rid of anything not needed - all the better.

jesperhodge: - `PersistentSubsectionGrade` -> it sounds like I just made a mistake there, I think I should remove that reference.
- Expect me to go through this ADR line by line tomorrow and straightening out misleading references like this and shorten it significantly.

jesperhodge: It looks like `StudentModule` is one of the large tables, on a similar magnitude as what we are doing, and is not subject to a db alias and router. I will investigate more to confirm this. Assuming this is correct, I recommend to change this:
- only the `StudentCompetencyCriteriaStatusHistory` - the HISTORY Leaf table - gets the router and db alias, and no foreign keys. All other tables will be normal tables with foreign keys, including the ACTIVE Criteria Status (Leaf table). So we should be mostly free to add foreign keys, with this one exception.

jesperhodge:
@mgwozdz-unicon remind me why we even need the history tables and not only keep the active tables?

mgwozdz-unicon:
> @mgwozdz-unicon remind me why we even need the history tables and not only keep the active tables?

We need people to be able to audit student mastery statuses to understand why a learner did or didn't achieve mastery of a particular competency or any of the associated "measurement instruments" (gradeable subsections).

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
```diff
@@ -0,0 +1,90 @@
...
+4. How should learner competency mastery be recorded concurrently and at scale?
+================================================================================
+
+Status
+------
+Proposed.
+
+Context
+-------
+When a learner is graded on a subsection, the platform must evaluate whether that grade
```
mgwozdz-unicon: The subsection trigger is specific to the POC but won't be the only trigger long term. Maybe say something like:
"When a learner is graded on a subsection (or any other learning instrument associated to a competency with a competency criteria, like a course or rubric criterion)"

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
```diff
@@ -0,0 +1,90 @@
...
+demonstrates any attached competencies and record the learner's mastery. Mastery is recorded at
+three levels: the criterion (leaf), the criteria group, and the competency. Per
+:ref:`openedx-learning-adr-0002` and :ref:`openedx-learning-adr-0005`, all three levels are
+*materialized* (stored), not recomputed on read, so that dashboards and other read surfaces stay
+fast. A single grade change therefore writes the changed leaf's status and then re-evaluates and
```
mgwozdz-unicon: Maybe this is nitpicky, but the re-evaluation has to happen regardless because ultimately (beyond this MVP), this should drive notifications to the learner of their accomplishment and/or automatic badge issuing or automatic certificate generation. The way this currently reads, it sounds like we only have to do this evaluation at this time because of the table structure, which isn't quite true.

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
```diff
@@ -0,0 +1,92 @@
...
+these reads feed the roll-up write and take the row locks above, so a replica's lag would compute a
+roll-up from stale siblings. The read-replica offload in :ref:`openedx-learning-adr-0005` is
+reserved for the read-only dashboard and reporting paths.
+
+**3. Entry point: edx-platform subsection grade change.** edx-platform
+computes subsection grades in an async celery task (`recalculate_subsection_grade_v3`) triggered by a score-change signal, not on the
+request thread. After that task writes the subsection grade, it calls a public openedx-core function
+within the same transaction; this function does the monotone merge and the upward roll-up. This should be generalized as needed to other places that trigger a competency status update.
+
+**4. ACTIVE writes and roll-ups commit atomically with the grade, or the whole task rolls back and retries.**
```
mgwozdz-unicon: Either it's not clear to me what this is trying to say or it contradicts point 3 above it. I think my confusion might stem from the phrase "with the grade" because that is implying to me that writes to our competency status tables roll up atomically with the subsection grade table. But I think this is trying to say that writes for the competency criteria hierarchy status tables commit atomically with each other. Am I confused or does this need to be reworded for clarity, or both?

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
```diff
@@ -0,0 +1,92 @@
...
+**5. The leaf HISTORY append runs outside the transaction, best effort.** Because the leaf HISTORY table
+(``StudentCompetencyCriteriaStatusHistory``) may be routed to a
+separate physical database (:ref:`openedx-learning-adr-0005`), its append runs after that transaction commits, as a best-effort,
+non-blocking write.
+
+
+Rejected Alternatives
+---------------------
+
+1. Prevent concurrent writes with a deployment-wide lock.
```
mgwozdz-unicon: I think it would be nice if the Rejected Alternatives each had brief Pros and Cons.

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
```diff
@@ -0,0 +1,92 @@
...
+
+    Once every write is a monotone merge and each
+    parent recompute is serialized only against concurrent writers of that same node by a brief row lock, correctness holds without needing a larger lock.
+
+4. Recompute derived levels on read instead of materializing them.
```
mgwozdz-unicon: This numbering is off

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
mgwozdz-unicon: Nevermind, I reread point 3, and I see that they're not contradictory now. It's possible that this could still be made clearer though since I did have to reread to understand.

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
mgwozdz-unicon: Or a rationale sentence or two like in ADR 5 below.

mgwozdz-unicon:
@jesperhodge The inline comments are from my completed review. I had Claude review as well. These are the things that came up from that which I found to be relevant:

> **Questions**
> 
> 1. `docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst` (mechanism 2, "When a child advances, its parent is recomputed..."): the parent-recompute lock (`SELECT ... FOR UPDATE`) only guarantees a fresh read of sibling children if it's the *first* read in that DB transaction. MySQL's default isolation, REPEATABLE READ, fixes a transaction's read snapshot at its first plain (non-locking) `SELECT`. A locking read like `FOR UPDATE` always sees latest-committed data and doesn't itself fix the snapshot, but any earlier plain read in the same transaction (e.g. re-reading the grade or the leaf row before the parent lock) would. If the recorder does any plain read before taking the parent lock, the later child reads could still be served from that earlier snapshot rather than post-lock data, silently reintroducing the write-skew this mechanism is meant to prevent. Nothing in the ADR states the required read order or assumed isolation level. Worth pinning down explicitly, e.g. "the parent lock must be the transaction's first read" or "isolation is set to READ COMMITTED for this path."


> 2. `docs/openedx_learning/decisions/0005-competency-mastery-storage.rst` ("Enable read-replica offload for heavy reads for the leaf tables"): read-replica offload is scoped only to the leaf-level tables (`StudentCompetencyCriteriaStatus` and `StudentCompetencyCriteriaStatusHistory`). Dashboards presumably also read group- and top-level status. Is leaving those two levels off the replica path intentional (they're orders of magnitude smaller, so primary-DB reads are cheap enough there), or an oversight? A one-line justification either way would help.

I'm pretty sure that this was because those tables are smaller, but I agree that this should be made clearer.

> **Suggestion**
> 
> `docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst` (mechanism 5, "The leaf HISTORY append runs outside the transaction, best effort"): the leaf HISTORY append is explicitly best-effort and non-blocking after the transaction commits, with no stated retry or reconciliation path. Since HISTORY exists specifically to serve as an audit trail (per ADR 0005's context section), a silently dropped append is a permanent audit gap. Worth at least a sentence on the acceptable failure mode: logged and ignored, or retried.

I would prefer at least one retry before logging and ignoring. If there's an accepted pattern in openedx for this though, then I defer to that.

> **Nits**
> 
> - `docs/openedx_learning/decisions/0003-competency-criteria-versioning.rst`: typo, "acvance" should be "advance."
> - `docs/openedx_learning/decisions/0002-competency-criteria-model.rst`: list items 7, 9, and 11 (the new History-table index entries) have two spaces after the number, while the rest of the list uses one space. Doesn't break the Sphinx build, just inconsistent.
> - `docs/openedx_learning/decisions/0002-competency-criteria-model.rst`: double blank line before the "Intended update flow" section.

ormsbee:
Django's design assumes `READ-COMMITTED`, and now [defaults to that on MySQL](https://github.com/django/django/blob/3e389b7ddaf08109900da5415ddaac5a355a170f/django/db/backends/mysql/base.py#L238) unless it's explicitly overridden in the connection options.

```python
In [3]: from django.db import connection
   ...: 
   ...: with connection.cursor() as cursor:
   ...:     cursor.execute("SELECT @@transaction_isolation;")
   ...:     print(cursor.fetchone()[0])
   ...: 
READ-COMMITTED
```

Running in higher isolation levels is not supported by our platform.

jesperhodge:
@mgwozdz-unicon @ormsbee I updated the ADRs based on your comments.
Main change: moved the leaf HISTORY append out of the grade transaction into its own transaction.on_commit-dispatched, retrying Celery task, and added a unique constraint on the advance (learner, node, status) as its idempotency key.

mgwozdz-unicon (approved):
Thanks for getting those changes in! Approving, though Claude flagged one more thing: the `db_constraint=False` decision itself has no stated rationale anywhere, it's the one bullet in ADR 0005 without any, while everything else in that doc (64-bit PKs, the dedicated HISTORY alias, read-replica scoping) explains why. This is the same field Braden raised a deletion-protection concern about on #661: that's resolved since his approach keys off the `competency_criteria_id` FK, not `user_id`, but that resolution isn't written down yet either. What's the actual reasoning for dropping the constraint on `user_id` here, and is it established for this table specifically, or assumed by analogy to `StudentModule`/`PersistentSubsectionGrade`? Not blocking on this, just think it might be good to add.

ormsbee:
The concurrency ADR looks good to me. I have just a couple of concerns with the storage one. Will write up in the morning.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+and a rule.
+
+The student mastery statuses tied to these tree nodes are stored in:
+- `StudentCompetencyCriteriaStatus` (leaf nodes)
+- `StudentCompetencyCriteriaGroupStatus` (middle nodes)
+- `StudentCompetencyStatus` (top-level)
+
+For each of these, we also need to persist history, because we need an audit trail to understand
+why a learner did or didn't achieve mastery of a particular competency or any of the associated "measurement instruments"
+(gradeable subsections).
```
ormsbee: It does not block this ADR, but let's talk with Product/UX folks and explore just how much of an audit history we really need. History storage is the most dangerously unbounded thing in this system. Moving it to a separate system and potentially pushing it to another database entirely doesn't really solve it.

For instance, say we determine that we only need history showing when they cross the aggregate threshold for the first time. We could potentially save billions of rows of storage for a large site that would otherwise be storing a new row for every time someone clicked "submit" on a problem somewhere.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+Storing every leaf multiplies out at scale. A course can carry on the order of 200 leaf criteria,
+so the leaf level is where the row count concentrates: the leaf table (learners x attempted
+leaves) potentially reaches the low billions for an Open edX instance with millions of learners. The dominant
+multiplier is this per-leaf breadth (roughly 200x per course), not time. Mastery is monotonic (see
+"Advance-only banking" below): a node can only advance through the small status lattice, at most a handful of
+forward steps ever.
+
+That scale is not, on its own, what makes a relational database struggle. A point lookup against a
+billion-row table backed by the right composite index is a logarithmic-time index seek regardless
+of the table's size; the dashboard reads this feature performs are exactly such point lookups. What
```
ormsbee:
```suggestion
of the table's size; the dashboard that reads this performs exactly such point lookups. What
```

Is this what was meant?

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+
+**Alleviate performance concerns by following established edx-platform practices.**
+edx-platform has already large tables of similar magnitude, such as `StudentModule`,
+and they have proven themselves.
+This means that as long as we follow their structure, we can have confidence that we are
+not introducing any huge new problems.
+They do not use physical partitioning, but work by using 64-bit primary keys,
+enabling optional read-replica offloads, and splitting out a history table that
+uses a dedicated database router. The decisions below are designed to mirror these
+existing patterns.
```
ormsbee: Okay, the database router thing was a giant, ugly hack, because I threw together the original `StudentModuleHistory` model in a half-day hackathon project to help out one of the first CS courses on the platform, and I neglected to override the primary key defaults to make it 64-bit. The result was that we were running out of primary keys, and some devops folks (Kevin I think?) retrofitted a new table in there

Also, the `courseware_studentmodulehistory` table stays somewhat manageable for large instances because they truncate it from time to time, because this history is used for course staff debugging purposes, not auditing. So this week's history is really important when a student is asking, "I submitted the right answer, why is the grader marking me wrong?", but that's not a thing we need to look up six months from now.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+and stays about the same order of size as ACTIVE. Keeping ACTIVE and HISTORY separate still pays:
+ACTIVE is a single in-place current row optimized for the dashboard point lookup and is the row
+per-learner concurrency is anchored on (:ref:`openedx-learning-adr-0004`), while HISTORY is
+append-only.
+
+**64-bit primary keys from the start.** The leaf ACTIVE and HISTORY tables use a 64-bit
+``BigAutoField`` primary key, chosen up front, mirroring edx-platform's
+``UnsignedBigIntAutoField`` on ``PersistentSubsectionGrade`` ("primary key will need to be
+large for this table"). Changing a primary-key type on a billion-row table later is
+prohibitively expensive.
```
ormsbee: This shouldn't need to be justified, as `BigAutoField` should be the default practice per [OEP-38](https://docs.openedx.org/projects/openedx-proposals/en/latest/best-practices/oep-0038-Data-Modeling.html#use-a-numeric-primary-key).

Also, please don't put up anything to imply that `UnsignedBigIntAutoField` is a good idea. It was an overkill reaction to having 32-bit primary keys run out on key tables, but it's more headache than it's worth. It theoretically gives us 2X the space of `BigAutoField`, but (a) the existing space of `BigAutoField` is so big that we'll never realistically reach it; (b) reaching it would break the database in a bunch of other ways anyway; and (c) declaring a custom field type like that introduces maintenance headaches like... (d) painful compatibility issues with PostgreSQL, which does not support unsigned ints.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+
+**A dedicated database alias and router for the leaf HISTORY table, baked in from the start.**
+This only applies to the `StudentCompetencyCriteriaStatusHistory` table.
+The leaf HISTORY table (``StudentCompetencyCriteriaStatusHistory``), the dominant table in this
+model, is the one learner-status table assigned to a dedicated Django database alias through a
+database router, mirroring edx-platform's courseware-history router
+(``StudentModuleHistoryExtended``), which is likewise a history table. The alias defaults to the
+main database.
```
ormsbee: Please don't do this. See my comment above about why this routing exists. Let's keep it simple and in the same database so that transactions just work.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+
+**No database-level foreign keys to `user_id` on ACTIVE or HISTORY table.**
+Foreign keys to ``user_id`` must have ``db_constraint=False`` set, mirroring edx-platform's own
```
ormsbee: That was done because we were forced to in order to do an in-place upgrade courseware_studentmodule to have 64-bit primary keys without downtime on edx.org: openedx/openedx-platform#20971

It is not intended as a general pattern. Please leave a real foreign key.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+``StudentModule``. This follows the app-boundary decision in :ref:`openedx-learning-adr-0001`, which keeps
+learner-status models decoupled from the concrete user model. Furthermore,
```
ormsbee: I'm not clear as to what part of the referenced ADR this is drawing from, but our convention in this repo has been to have real foreign keys to the User model—it's just that we respect Django's ability for projects to define custom user models by grabbing it from `settings.AUTH_USER_MODEL`, instead of assuming that it's always `django.contrib.auth.models.User`.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+a real constraint has an ongoing cost at this table's write volume:
+a hot user row would see extra lock contention under concurrent writes. This
+follows the app-boundary decision in :ref:`openedx-learning-adr-0001`, which keeps
+learner-status models decoupled from the concrete user model.
```
ormsbee: So in theory, very little should write to the User model, but it's true that we have seen reports of concurrency issues that we don't understand very well at this point:

* https://github.com/openedx/completion/pull/443
* https://github.com/openedx/openedx-platform/pull/38815

On that note, if there is competency information that makes sense to attach in a 1:1 way per student, e.g. a `CompetencyStudent` model that would go 1:1 with `User`, it might make sense to make various competency models foreign key to that instead of directly to `User`.

In either case though, I think we can remove this decision from the ADR.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+(Decision 7): that boundary keys off ``competency_criteria_id`` and its ancestor tables, not
+``user_id``, so dropping the database-level constraint here does not weaken it.
+
+**Enable read-replica offload for heavy reads for the leaf tables.**
+This only applies to the ACTIVE `StudentCompetencyCriteriaStatus` and HISTORY
+`StudentCompetencyCriteriaStatusHistory` tables. Offload is scoped to the leaf level because that is
+where row count and read volume concentrate (the ~200x per-course leaf multiplier above); the group
+and top-level status tables are orders of magnitude smaller, so their dashboard reads are cheap point
+lookups that do not need to leave the primary and are not worth the replica-lag complexity.
+Prior art: ``edx_django_utils``'s ``read_replica_or_default()``.
```
ormsbee: This is always a possibility for the future, but one we'd want to punt until we had verified that it was a problem worth tackling and testing for. StudentModule load is really high because it's used in many, many places, the payload can be enormous, and the lookup pattern happens via a very expensive/large composite index. My hope would be that even on sites that make heavy use of CBE, the load from it should be much lower.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+
+**Advance-only banking, monotonic.** Once a node reaches ``Demonstrated`` its ACTIVE row is retained
+("banked"): the recorder never automatically regresses it, not on a later downward grade correction
+and not on a criteria change. This applies at every level, including the leaf. A genuine downward
```
ormsbee: I strongly suspect that this will end up changing in the future, but I'm okay with it as a starting position.

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
```diff
@@ -0,0 +1,163 @@
...
+grade correction does not advance the status, so it writes no HISTORY row and leaves the banked
+ACTIVE status unchanged; because HISTORY records only advances, it never carries suppressed
+regressions. Reversing a banked status is a separate administrative action, out of scope here.
+This monotonicity is what makes out-of-order and duplicate delivery safe, since a late or replayed
+event can never lower a status, and :ref:`openedx-learning-adr-0004` relies on it.
+
+**Retroactive criteria changes are monotonic for the learner.** A retroactive edit can newly grant
+or preserve mastery, but it never silently revokes it, and it never rewrites a learner's recorded
+leaf mastery downward.
```
ormsbee: Again, I take the point that we shouldn't silently do this by default, but I expect people will have errors in their content settings, and this will need to be adjustable at some point. But also again, I'm not going to block this as a starting position.

Also, we've had mention of use cases in the past where mastery has a shelf-life, e.g. Continuing Medical Education (CME) credits.

ormsbee (changes requested):
These concerns are over the data storage portion. Please forgive me if I've made these comments earlier. (I thought I had already commented on the 64-bit ints before, but maybe I never properly submitted that review. ☹️)

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
jesperhodge: @mgwozdz-unicon what are your thoughts?

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
jesperhodge: Maybe my misunderstanding. https://github.com/openedx/openedx-core/blob/845dc1fc64f93023d7c5672ddc3e03c44effc2aa/docs/openedx_learning/decisions/0001-competency-criteria-location.rst?plain=1#L48

docs/openedx_learning/decisions/0005-competency-mastery-storage.rst
jesperhodge: Yes, I think that is already the intention here, let me know what you think:
status advances are only monotonic, e.g. from partially demonstrated to demonstrated. So whenever there is no such status advance, we will not write anything into the history table.
I even gave it a hard db constraint in ADR 0002:
"   7. ``StudentCompetencyCriteriaStatusHistory(user_id, competency_criteria_id, status_id)`` (unique -- at most one HISTORY row per learner, leaf, and status level, which also serves as the idempotency key for the append in :ref:`openedx-learning-adr-0004`)"
I will make this a separate point so that it is more clear.
Think like this: The active table stores group and leaf nodes - for each learner for each subsection they attempt, for example. Sure, that is a lot.
Now when the learner tries a subsection multiple times and never crosses the grade threshold, it will remain "attempted, not demonstrated" and only one history entry will be written - because there is only a one-time status advance.
Since every node just has 3 or 4 statuses - for example "not demonstrated" -> "partially demonstrated" -> "demonstrated" for group nodes (I'm paraphrasing here, it might be a little inaccurate), the history table will only ever receive 3 or 4 writes for the group node.
So the history table will only ever be 3-4x as large as the active table.

What are your thoughts so far? Does that alleviate your concern, or do you see further opportunity to cut down the history table?

Also, one note: if we want to allow course authors or django admins to manually edit the status later downwards in case there was a mistake, and we want to keep that in the audit trail as well, we need to drop the unique constraint.

mgwozdz-unicon: I'm pretty sure some people will definitely want to see the history at that granular of a level, and in a few different use cases.

One of the things I remember from our interviews with institutions was that many folks were talking about the importance of learners being able to see their historical progress of competency attainment across multiple attempts in a single assignment as well as across the entire course, and surely at a program level as well. 

But there will also be analytics/reporting use cases where instructors and instructional designers will want to understand some kind of aggregation of how learners were progressing through each material as it rolls up to the competency so that they know which learning materials need improvements and can try to get insights about which learning materials did well so that they can emulate that success.

But yes, the product folks and our CBE SME Jason could probably all provide more insights here.  I'm sure that there are certain metrics that would take priority over others, but I feel pretty confident that auditing historical records will be desired at the leaf/subsection level, which is the one that is the biggest storage challenge. It's actually some of the middle roll-ups that some people may not care about as much, but I also think that some people will definitely want all the detail they can get. 

If someone has a reasonably complicated logical hierarchy for how to demonstrate mastery of a particular competency and a learner inquires about why they haven't received their badge for it, then someone will need to do an audit to figure out where in that logical hierarchy the student didn't achieve so that they can explain this to them. 

To Dave's point later on, there probably is a limit to the amount of time that a learner would ask the question "I thought I demonstrated mastery of this competency. Do an audit to find out why I was told I haven't." Which could probably make it safe to truncate the table at some point in history, though I know there are also laws and retention periods and such that need to be adhered to.

Out of these use cases, the one that would likely want historical data furthest in the past is actually probably the analytics/reporting case in the event of a course that runs over several months and they want to compare data for a specific subsection from the last 5-10 times this course took place, for example.

jesperhodge:
@ormsbee @mgwozdz-unicon @thelmick-unicon two open questions:

- Do we need the history table to keep records forever, or can we auto-truncate it after a certain period?
- If we only store advancing states in the history table, does that conflict with any edge cases? For example, if a learner grade gets corrected by an instructor, or a django admin corrects mistaken entries - does the correction need to be traceable in the history table?

mgwozdz-unicon:
> * Do we need the history table to keep records forever, or can we auto-truncate it after a certain period?

I think it would be safe to auto-truncate as long as the instance owner is able to control what the auto-truncation period is.

> * If we only store advancing states in the history table, does that conflict with any edge cases? For example, if a learner grade gets corrected by an instructor, or a django admin corrects mistaken entries - does the correction need to be traceable in the history table?

I think we would want corrections and edits traceable in the history table as well.

jesperhodge:
> I think it would be safe to auto-truncate as long as the instance owner is able to control what the auto-truncation period is.

@mgwozdz-unicon I'll defer to you on this, since I don't really know what the pieces are that need to be read off of the history table. -> So I'll add auto-truncation, but what should the mechanism for instance owners be - an environment variable? And what should the default setting for this be?

> I think we would want corrections and edits traceable in the history table as well.

@mgwozdz-unicon how about we make it so there are two routes for this?
1. we record only advances, no repeat attempts or evaluations, and we don't allow downwards corrections that result from criteria tree changes. This keeps the history table small
2. Only in the case where the assignment is directly changed by a course instructor or admin we record this change in the django history (even if it's downward).

jesperhodge:
@ormsbee I addressed your feedback; only the two open questions discussed with Mary above are not handled yet.

thelmick-unicon:
@jesperhodge 
> Do we need the history table to keep records forever, or can we auto-truncate it after a certain period?

I agree with Mary, this is going to be variable by institution, depending on when they will want to truncate the history. As for a default, my first thought is 2 years, but I think that's open to discussion. This might be a good question for Jason.

> If we only store advancing states in the history table, does that conflict with any edge cases? For example, if a learner grade gets corrected by an instructor, or a django admin corrects mistaken entries - does the correction need to be traceable in the history table?

Again, I agree with Mary that manual changes should be captured and traceable in the history.

> how about we make it so there are two routes for this?
> we record only advances, no repeat attempts or evaluations, and we don't allow downwards corrections that result from criteria tree changes. This keeps the history table small
> Only in the case where the assignment is directly changed by a course instructor or admin we record this change in the django history (even if it's downward).

Feedback we received from institutions was that they explicitly wanted to capture all repeat attempts and evaluations so both the instructor and student can see all attempts to evaluate if progress is being made or to review feedback from previous attempts.

mgwozdz-unicon:
> > I think it would be safe to auto-truncate as long as the instance owner is able to control what the auto-truncation period is.
> 
> @mgwozdz-unicon I'll defer to you on this, since I don't really know what the pieces are that need to be read off of the history table. -> So I'll add auto-truncation, but what should the mechanism for instance owners be - an environment variable? And what should the default setting for this be?

Yes, I think an environment variable sounds good. Claude is recommending a Django setting as a precedent in the codebase. The 2 years Tammie suggested sounds good to me, though that's probably an implementation detail that can get worked out later. I put a question out to Jason about it.

> > I think we would want corrections and edits traceable in the history table as well.
> 
> @mgwozdz-unicon how about we make it so there are two routes for this?
> 
> 1. we record only advances, no repeat attempts or evaluations, and we don't allow downwards corrections that result from criteria tree changes. This keeps the history table small
> 2. Only in the case where the assignment is directly changed by a course instructor or admin we record this change in the django history (even if it's downward).

I agree with Tammie that we definitely want to keep data for repeat attempts/evaluations. In general, we probably do want to allow recording downwards corrections whether the correction was authored by a person or by software since I'm pretty sure I recall that D2L Brightspace supports displaying a log of all grade changes to an assignment over time whether it was changed by a system or a specific person. That way we can support parity with what institutions already expect.

I think the main consolation here is that ADR 3 says "If a user edits competency criteria definitions or competency object/tag associations after related learner status exists, Studio must display an explicit warning that student statuses have already been set, and these changes will be applied going forward, so existing learner statuses will not be retroactively updated." So at least they won't get retroactively updated by the software in that case.

jesperhodge:
@mgwozdz-unicon @thelmick-unicon The platform stores attempts and their scoring already, right? Why can't we just pull the scores and history from that table without saving the skill status updates for attempts? I would like to avoid adding any of this to the history table.
Unless there's anything that the platform does not store already?

Followup: if you can specify exactly what data we want from the audit trail, maybe we don't need to store history at all? Instead just pull everything from the existing data on attempted subsections and such.

mgwozdz-unicon:
> @mgwozdz-unicon @thelmick-unicon The platform stores attempts and their scoring already, right? Why can't we just pull the scores and history from that table without saving the skill status updates for attempts? I would like to avoid adding any of this to the history table. Unless there's anything that the platform does not store already?
> 
> Followup: if you can specify exactly what data we want from the audit trail, maybe we don't need to store history at all? Instead just pull everything from the existing data on attempted subsections and such.

@jesperhodge The platform's attempt/submission data doesn't carry any notion of competency or criteria, that mapping only exists here. So pulling from it doesn't remove the storage need, it just relocates it: we'd still need to record which CompetencyCriteria (and which version of it) applied when a status was evaluated. Without that, this collapses into rejected alternative 1 in this same ADR (transient computation), which we already ruled out because retroactive criteria edits would make recomputed history wrong, there's no frozen record of what was true at the time. Separately, the ADR already frames this as covering "any of the associated measurement instruments," not just gradeable subsections, so keeping competency status/history in one place avoids re-deriving this pattern per instrument as new measurement types get added.

jesperhodge:
I've worked with Claude to come up with some sort of suggestion on how to handle this. Here is a summary:
"
Two different things are being kept in one table, and they want opposite things. One is a record of mastery: this learner reached Demonstrated on this criterion at this time. It stays small, and it should be kept forever, because it is effectively a transcript. The other is a record of activity: every attempt, every score, the feedback an instructor wrote. It grows without limit, and it is the kind of data that ought to expire. That mismatch, not the size, is the real problem. The request for full attempt history is also not scope creep. ADR 0005 says the history tables exist to understand "why a learner did or didn't achieve mastery," then records only advances, so a learner who tried five times and never passed leaves no trace, and the design cannot answer its own question. The recommendation is to keep the mastery record in ADR 0005, and to add a few counters to the row it already writes: attempts, and best score. Those reveal whether a learner is making progress, and they cost nothing, because the row is written anyway. The attempt log itself belongs in a separate ADR, where it can have its own retention policy, and where the harder question about how to store the record of activity. Note that written instructor feedback lives in ORA anyway, which is not currently owned by openedx-core.
"

jesperhodge:
Okay, some new perspectives.
I investigated two directions with Claude:
A) ways to compact, aggregate, and offload the history to reduce size / migration complications;
B) my doubts on whether we need the history at all - Claude's evaluation is we do not need it. (More promising avenue.)

More details:
A) in short, there are a couple alternatives here. 1. We can use analytics tools like aspects / clickhouse to completely avoid MySQL; 2. we can compress rows older than 1 month into (user_id, course_id, month) rows that use compression algorithms, which saves a huge amount of space; 3. we can do the same thing as in 2 but offload everything older than a month to S3 as well.

B)
I have my doubts we need the history table for leaf criteria statuses at all, because it mainly reflects data that is already stored elsewhere. I had Claude analyze how we could avoid the history table, and it's opinion is also that the history table is not needed. Here is a summary of its analysis:

"We do not need the huge attempt-history table, and dropping it removes the whole scaling problem. The audit requirement bundled together three different things: every problem attempt a learner makes, every recalculation of a subsection score, and the much smaller record of when a learner's competency status actually changed. edx-platform already stores the first one and always has, so duplicating it into our competency tables would be paying for tens of billions of rows twice. The genuine concern behind the request is not volume but traceability: when we say a learner earned a competency, we need to know what the criteria looked like at that moment, since criteria can be edited later. That is a question about authored definitions, not learner activity, and definitions are tiny by comparison. So the fix is to keep our small advance-only history and have each recorded advance point at the criteria configuration that produced it. Importantly that is a pointer, not a copy: the learner row gains a single column, while the configuration it names is stored once and shared by every learner evaluated against it, so that side grows with authoring edits rather than with learners. At the leaf level we add almost nothing, because ADR 0003 already versions criteria definitions, so the pointer is simply the version number of something we are keeping anyway. This is the same trick edx-platform already uses to freeze what content made up a graded subsection. Instructor reviews get their own separate table, because their volume is bounded by how much staff actually write, not by learner activity. The cost is that some audit questions become slow, awkward, cross-database queries, which we have agreed is acceptable. In exchange we avoid partitioning, sharding, archiving to S3, and multi-terabyte schema migrations entirely."

@mgwozdz-unicon @ormsbee thoughts?

docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst
```diff
@@ -0,0 +1,137 @@
...
+  path must keep up under peak load.
+
+Decision
+--------
+
+**1. Every write is a monotone merge, never a blind overwrite.** A node's status is written as
+``status := max(stored status, newly computed status)`` (a single ``GREATEST``-style ``UPDATE``,
+atomic at the row for the duration of that one statement, with no application-level lock). Because
+the merge takes the higher of the two values, it is commutative, idempotent, and insensitive to
+order. This is why out-of-order delivery and re-delivery are harmless without sequence tracking.
```
kdmccormick: try to write in plainer english and cut right to the point.

here is my understanding of points 1-3. let me know if you disagree.

> 1. **Writes to a node's status will only ever advance its value.** Roughly, `status := max(stored status, newly computed status)`. As a result, out-of-order delivery and re-delivery are harmless.
> 2. **When a child's value advances, its parent will be recomputed in the same transaction**. Parent nodes may have conjunctive mastery rules like "demonstrated only when all children are demonstrated". Imagine that two child nodes share a parent, and each child advanced in mastery at the same time. If each child tried to recompute the parent's mastery at the same time, then they might use the other child's outdated mastery value. To protect against this, a node that is advanced in mastery should take a row-level lock and update just its immediate parent in the same transaction. Two updates that touch the same ancestors will essentially take turns turns up the path to the root, in a consistent order, so concurrent updates cannot deadlock.
> 3. **Subsection grade changes will trigger the mastery computation.** openedx-platform
computes subsection grades in an async celery task (`recalculate_subsection_grade_v3`) triggered by a score-change signal. After that task writes the subsection grade, it will call a public openedx-core function within the same transaction to recompute mastery as described in 1 and 2. As other kinds competency criteria are defined (e.g. completion), additional triggers will be added.

i know it can be very challenging to distill text like this--it took me 20 minutes write this comment :P but the practice of doing it will force you to refine your ideas and it will help dave and I review more quickly.

jesperhodge: Thanks so much for doing that! Sounds good, I'll take another pass reviewing / improving the language. This indeed looks much better!

It's great to have an example of how this text can be improved. Now that I've looked at it, if you ever see something like this where text is not clear enough or needs to be more understandable / more in plain English - in the future, you can also just leave a comment for me to rewrite it. No need to dig around for you to decipher what I mean or find suggestions yourself - if it's not clear enough, I can rewrite so it becomes clear, even before you review the content.

kdmccormick: in this case, the exercise of writing it helped me come up to speed on the content of the ADR. but in the future i'll just ask, thanks!

jesperhodge:
Pulled ADR 4 out into https://github.com/openedx/openedx-core/pull/713/ .

mgwozdz-unicon (changes requested):
@jesperhodge Thank you for updating this PR based off of the consensus I'm seeing in [Slack](https://openedx.slack.com/archives/C09A1BN1CSE/p1786568835908359) that we don't need to store learner competency status history for MVP.

After contemplating this a bit, I'm getting the impression that we may not need ADR 5 at all anymore given that the decision was to not store history for student competency statuses. I think it would be cleaner to just update Decision 5 of ADR 3 to say that history won't be stored for the learner status tables for MVP and just add some Rejected Alternatives covering everything that has happened in this PR.

Following are some suggestions for what updates to make, but you don't need to consider them set in stone if you feel there are some other points that are important to include that I have missed or if you'd rather state something differently:

## Suggested Decision 5 replacement (`0003-competency-criteria-versioning.rst`)

```rst
5. Do not store history for learner competency status tables, and update rows in place.

   For ``StudentCompetencyCriteriaStatus``, ``StudentCompetencyCriteriaGroupStatus``, and
   ``StudentCompetencyStatus``, each row is updated in place when a learner's status
   changes. There is no history of prior status values beyond the ``modified`` timestamp,
   and no separate history table.

   Open edX has no concept of gradeable subsection attempts. This means that an attempt will actually be defined at the level of an individual problem, which can result in tens of billions of rows for a large Open edX instance. That scale creates real operational burden (schema migrations, backups, truncation and retention policy). Therefore, we did additional market research and found that storing history for learner competency status will not be required by the initial pilot partners for the MVP of the CBE implementation, and it would be safe to add later if needed.
```

## Suggested Rejected Alternatives additions (append after existing item 4)

```rst
5. Keep the learner status tables append-only, storing every status change as a new row
   (the design this replaces).

    - Pros:
        - Every write is an insert rather than a read-modify-write, so there is no current
          row to keep consistent.
        - Preserves a full audit trail of every status change.
    - Cons:
        - No MVP requirement calls for this history.
        - Grows the leaf table by a further multiplier of problem attempts per learner per
          leaf, reaching tens of billions of rows for a large instance.
        - A read must resolve the latest row for a learner and node rather than reading one
          in-place row, which is more expensive and more complex.

6. Compute leaves transiently, never store them.

    - Pros:
        - Eliminates the largest table, since leaf demonstration would be computed on
          demand from the leaf's rule and the learner's grade.
    - Cons:
        - A recomputed leaf reflects the rule as it stands now, not the rule in force when
          the learner was graded, which contradicts Decision 4 above and can silently
          lower a status.

7. Store child evaluations on the parent group row instead of a leaf table.

    - Pros:
        - Avoids the largest table entirely.
    - Cons:
        - A leaf write becomes a read-modify-write of a column shared with every sibling.
        - No unique index or foreign key behind a status packed into a per-group array.
        - Couples a leaf's frozen mastery to the current shape of the criteria tree.
        - Removes the ability to individually track competency status progress by learning object (e.g. subsection)

8. Put the leaf table behind its own database alias/router, a separate physical database,
   or native partitioning/sharding, from the start.

    - Pros:
        - Physically isolates or splits the largest table from the start.
    - Cons:
        - A second alias gives up the atomicity these writes need with the grade write.
        - Imposes real operational cost on every deployment with nothing measured to
          justify it, and remains available later if a specific need is proven.

9. Serve heavy leaf-table reads from a read replica.

    - Pros:
        - Keeps dashboard and reporting reads off the primary.
    - Cons:
        - Premature: no measurement shows the primary struggling, and these are point
          lookups on a composite index, not the wide, expensive reads that drive
          ``StudentModule`` load in edx-platform.
        - Adding it later is a per-query choice, not a schema decision.

10. Give the leaf table a custom unsigned 64-bit primary key.

    - Cons:
        - ``BigAutoField``'s range is already far out of reach for this table.
        - Unsigned integers do not exist in PostgreSQL, and the custom field type carries
          ongoing maintenance cost for no real benefit at this scale.

11. Drop the database-level foreign key constraint on the learner column.

    - Cons:
        - This repo's convention is a real foreign key to ``settings.AUTH_USER_MODEL``.
          Reports of contention on the user row exist elsewhere in the ecosystem, but are
          not understood well enough here to design around.
```

mgwozdz-unicon (approved):
Thank you for getting those changes in!

ormsbee (approved):

ormsbee:
Approved, but it looks like it will need manual rebase before it can be merged.

jesperhodge:
Hi @ormsbee I rebased this. Assuming the pipeline passes, could you merge?

PR #658, opened 2026-07-10
docs: add ADR for non-nullable tag external_id

mgwozdz-unicon:
## Why
CBE needs every tag to have an `external_id` so it can be associated with objects as part of competency criteria. Non-CBE taxonomies are expected to want a mandatory, institution-assigned identifier too, particularly once there's a need to convert an existing regular taxonomy into a Competency Taxonomy.

## Summary
- Proposes making `Tag.external_id` `NOT NULL` at the database level, for every taxonomy.
- Builds on `openedx/openedx-core`'s ADR 0010 (mutable `external_id`): backfills every existing NULL row with a value derived from the tag's `value` (first ~7 characters, with a deterministic numeric-counter fallback on collision), then adds the constraint in the same migration.
- New tags created without an `external_id` (REST API, import, UI) get one auto-generated the same way if not provided, rather than the request being rejected; existing integrations that omit `external_id` today don't need to change.
- Institutions can replace an auto-generated value with their own through ADR 0010's rename pathway.
- Not treated as a breaking API change or routed through the DEPR process: the request side stays optional, and the response side only narrows from nullable to always-present.

### DIFF
--- a/docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
+++ b/docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
@@ -29,6 +29,9 @@ mechanism that lets multiple tags in the same taxonomy have "no external_id" at 
 Making the field mandatory means every existing NULL row in every current Open edX
 instance needs a real value before the constraint can be added, since this is a
 published library whose migrations run unmodified against arbitrary downstream data.
+``Tag.value`` already carries the same per-taxonomy, case-insensitive uniqueness
+constraint (``unique_together(taxonomy, value)``), so backfilling ``external_id`` from
+``value`` cannot introduce a new collision under the constraint being added here.
 
 Decision
 --------
@@ -41,18 +44,17 @@ migration.
   backfill-then-constrain shape as the ``depth``/``lineage`` migration
   (``0020_tag_depth_and_lineage.py``): populate real values first, then add the
   constraint in the same migration once every row satisfies it.
-- **Backfill algorithm.** Take the first ~7 characters of the tag's ``value`` (matching
-  ``external_id``'s existing case-insensitive collation). If that prefix collides with
-  another tag's ``external_id`` already in the same taxonomy, append a numeric counter
-  to the prefix and keep incrementing it until it's unique. This is deterministic and
-  bounded: it never needs an unbounded retry loop, since appending an incrementing
-  counter to an already-short prefix is guaranteed to terminate within the taxonomy's
-  tag count.
+- **Backfill algorithm.** Set ``external_id`` to the tag's existing ``value``. Since
+  ``(taxonomy, value)`` is already a unique, case-insensitive constraint matching
+  ``external_id``'s own, this is guaranteed unique with no collision handling needed.
+  Implementations may transform the copied value for UI legibility, for example
+  uppercasing it and replacing spaces with underscores, but no particular format is
+  required by this decision.
 - **New tags going forward.** The REST API and import format keep treating
   ``external_id`` as optional for the caller, unchanged from today. When a tag is
-  created without one, the same prefix-plus-collision-fallback algorithm generates one
-  automatically, rather than rejecting the request. No existing integration that omits
-  ``external_id`` today has to change.
+  created without one, the same backfill algorithm generates one automatically, rather
+  than rejecting the request. No existing integration that omits ``external_id`` today
+  has to change.
 - **Institutions can replace an auto-generated value.** If an institution doesn't want
   the auto-generated ``external_id``, they can change it to their own value through
   ADR 0010's rename pathway.
--- a/docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
+++ b/docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
@@ -1,0 +1,101 @@
+.. _openedx-tagging-adr-0012:
+
+12. Non-nullable tag ``external_id``
+=====================================
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+`ADR 0010 <0010-mutable-tag-external-id.rst>`_ made ``Tag.external_id`` mutable through an
+import-only rename pathway, to serve as the institution-editable "Competency ID" for
+Competency-Based Education (CBE) tags. That ADR left ``external_id`` nullable, scoped to
+competency taxonomies.
+
+Building the CBE competency workflow surfaced a stronger requirement: every tag needs a
+value for ``external_id`` in order to be able to associate it with objects as part of 
+competency criteria. It's also anticipated that non-CBE taxonomies will want a mandatory,
+institution-assigned identifier in the future, particularly when there becomes a desire to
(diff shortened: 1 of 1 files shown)


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

bradenmacdonald (approved):
Generally this makes sense to me.

I'm just wondering about the algorithm for deriving the default `external_id`. I think I would prefer just using `value` directly or using the uppercase version of `value` (perhaps with spaces replaced with underscores) as the default for backfilling (with a number added if needed, but it almost never will be since `value` is also unique per taxonomy and most taxonomies either have external IDs for all values or none).

The reason is that I expect we'll be displaying these values in the UI, and in the taxonomies I've looked at there are often a lot of tags that start with similar prefixes. So I think things like `ADMINISTRATIVE_FUNCTIONS`, `ADMINISTRATION`, `ADMINISTRATIVE_SOFTWARE` are better external_ids than `Adminis`, `Adminis2`, `Adminis3`, and will be less confusing to users when they're seen.

docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
```diff
@@ -0,0 +1,101 @@
...
+Building the CBE competency workflow surfaced a stronger requirement: every tag needs a
+value for ``external_id`` in order to be able to associate it with objects as part of 
+competency criteria. It's also anticipated that non-CBE taxonomies will want a mandatory,
+institution-assigned identifier in the future, particularly when there becomes a desire to
+convert regular taxonomies into Competency Taxonomies.
+
+Today, ``external_id`` is nullable specifically to work around its own
+``unique_together(taxonomy, external_id)`` constraint: NULL is the only value Postgres,
+MySQL, and SQLite all treat as exempt from a uniqueness constraint, so it's the
+mechanism that lets multiple tags in the same taxonomy have "no external_id" at once.
```
jesperhodge, line 28: Nit. For clarification: this just means that if we don't have nullable `external_id`, then every `external_id` in the same taxonomy must be unique and we can't use a workaround like a string to avoid external_id if it's not available, correct?
And this is does not pose a problem for us because the "value" field already is unique per taxonomy, so just backfilling to external_id will have external_id be unique.

docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
```diff
@@ -0,0 +1,101 @@
...
+instance needs a real value before the constraint can be added, since this is a
+published library whose migrations run unmodified against arbitrary downstream data.
+
+Decision
+--------
+
+Make ``Tag.external_id`` ``NOT NULL`` at the database level, for every taxonomy, in one
+migration.
+
+- **Backfill for existing rows.** A ``RunPython`` step generates a value for every tag
```
jesperhodge, line 44: Nit. Sounds fine, but curious why not two migrations - a data migration for backfill followed by a migration with this constraint?

docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
```diff
@@ -0,0 +1,101 @@
...
+  whose ``external_id`` is currently NULL, following the same
+  backfill-then-constrain shape as the ``depth``/``lineage`` migration
+  (``0020_tag_depth_and_lineage.py``): populate real values first, then add the
+  constraint in the same migration once every row satisfies it.
+- **Backfill algorithm.** Take the first ~7 characters of the tag's ``value`` (matching
```
jesperhodge: Why the first 7 characters, not all of the value?
Taking the first 7 characters seems to create unnecessary collisions because we have skills like
"data" and "data science" and such that share the same initial characters. And having external ids like "data1" and "data2" for "data" and "data science" could be misleading.
Values are already used as unique labels / identifiers. Wouldn't copying the whole value string (if necessary adding some affixes) be cleaner? Is there any drawback to it?

jesperhodge (changes requested):
Looks great to me, just have a look at my comment about the 7-character cutoff.

mgwozdz-unicon:
@bradenmacdonald Agreed, and it turns out to simplify things really nicely. Because `value` already has its own `unique_together(taxonomy, value)` constraint (case-insensitive, matching `external_id`'s), backfilling `external_id` directly from `value` is guaranteed collision-free, so no counter or fallback logic is needed at all. I kept your uppercase/underscore suggestion in as an example of a UI-legibility transform an implementation could apply, but didn't mandate a specific format in the ADR itself.

@jesperhodge, addressed your three comments:

1. Confirmed, and added a sentence to the Context section: `value` already carries the same per-taxonomy, case-insensitive uniqueness constraint as the one being added to `external_id`, so backfilling from it can't introduce a new collision. No workaround needed.
2. Going to leave this as one migration rather than splitting it. The existing `0020_tag_depth_and_lineage.py` migration already established backfill-then-constrain-in-one-migration as the pattern here, so this follows precedent. Splitting it would also leave a backfilled-but-not-yet-constrained state that a downstream instance could get stuck in if its migration run is interrupted or the two land out of order across a release, one migration avoids that.
3. Resolved by the same change as Braden's comment: backfill now uses the full `value` rather than a fixed-length prefix, so there's no truncation to create the collisions you flagged (e.g. "data" vs. "data science").

docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
```diff
@@ -41,18 +44,17 @@ migration.
...
-- **Backfill algorithm.** Take the first ~7 characters of the tag's ``value`` (matching
-  ``external_id``'s existing case-insensitive collation). If that prefix collides with
-  another tag's ``external_id`` already in the same taxonomy, append a numeric counter
-  to the prefix and keep incrementing it until it's unique. This is deterministic and
-  bounded: it never needs an unbounded retry loop, since appending an incrementing
-  counter to an already-short prefix is guaranteed to terminate within the taxonomy's
-  tag count.
+- **Backfill algorithm.** Set ``external_id`` to the tag's existing ``value``. Since
+  ``(taxonomy, value)`` is already a unique, case-insensitive constraint matching
+  ``external_id``'s own, this is guaranteed unique with no collision handling needed.
```
bradenmacdonald: It _is_ possible I think for some weird taxonomy to exist where only some tags have external IDs specified and one of those existing external IDs is the same as the `value` of a different tag that doesn't yet have an external ID. This should be an extremely rare case, maybe never encountered at all, but it is possible, and if it happened, then some sort of conflict handling (append a number etc.) would be required.

bradenmacdonald (approved):

ormsbee (approved):

docs/openedx_tagging/decisions/0012-non-nullable-tag-external-id.rst
mgwozdz-unicon: You're right. I committed an update to address this.

jesperhodge (approved):

PR #661, opened 2026-07-15
docs: add ADR for persisted competency deletability signal

mgwozdz-unicon:
## Why
[openedx-core#655](https://github.com/openedx/openedx-core/issues/655) proposes how to enforce the
delete/edit guardrails ADR-0002 and ADR-0003 already decided for competency taxonomy data, and
raised surfacing a `deletable` signal on the relevant GET endpoints so Studio can disable/warn on
the delete action instead of only discovering the block from a failed request. How to compute that
signal turned into a real design question (read-time cost vs. write-path complexity), so it gets
its own ADR rather than staying an implementation detail of that issue.

## Summary
- Adds ADR 0005: persist a `deletable` boolean on the five delete-protected models (`Taxonomy`,
  `CompetencyTaxonomy`, `Tag`, `ObjectTag`, `CompetencyCriteriaGroup`, `CompetencyCriteria`) rather
  than computing it per-request or caching it.
- The flip is one-directional (`True` to `False` only, never back), set synchronously in the same
  transaction as the learner-status write that triggers it. No cache-invalidation logic needed as a
  result.
- No backfill migration needed: the referencing learner-status tables are new, so no existing row
  can already be non-deletable.
- Enforcement stays on the write path; the persisted field is a UX convenience, not the guardrail
  itself.
- Rejects two alternatives considered in #655: a per-request `Exists()`/`OuterRef()` annotation, and
  caching that computed value with a TTL.

Related to #655.

### DIFF
--- a/docs/openedx_learning/decisions/0005-persisted-deletability-signal.rst
+++ b/docs/openedx_learning/decisions/0005-persisted-deletability-signal.rst
@@ -1,0 +1,106 @@
+.. _openedx-learning-adr-0005:
+
+5. How should delete-eligibility be signaled for competency guardrail records?
+================================================================================
+
+Status
+------
+
+Proposed
+
+Context
+-------
+
+:ref:`openedx-learning-adr-0002` Decision 7 already decides *which* records are delete-protected:
+once any related learner status exists in ``StudentCompetencyStatus``,
+``StudentCompetencyCriteriaGroupStatus``, or ``StudentCompetencyCriteriaStatus``, hard delete is
+blocked for ``Taxonomy``/``CompetencyTaxonomy``, ``Tag``, ``ObjectTag``, ``CompetencyCriteriaGroup``,
+and ``CompetencyCriteria``. ``CompetencyRuleProfile`` has no delete path at all, with or without
+learner status; archive is its only removal path.
+
+What that decision doesn't cover is how a caller finds out a record is currently protected *before*
+attempting to delete it, so Studio can disable or warn on the delete action instead of only
+discovering the block from a failed request.
+
+Decision
+--------
+
+Persist the signal instead of computing it at read time
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Add a boolean ``deletable`` field, defaulting to ``True``, to each of the five delete-protected
+models: ``Taxonomy``/``CompetencyTaxonomy``, ``Tag``, ``ObjectTag``, ``CompetencyCriteriaGroup``,
+and ``CompetencyCriteria``. GET endpoints for these models expose the column directly; no
+annotation or subquery runs on the read path at all.
+
+``CompetencyRuleProfile`` does not get this field: it has no DELETE endpoint to gate, per ADR-0002
+Decision 7.
+
+No backfill computation is needed. All five models' rows only ever become non-deletable by way of
+a ``StudentCompetencyCriteriaStatus``/``StudentCompetencyCriteriaGroupStatus``/
+``StudentCompetencyStatus`` row referencing them, and those three tables are created by this same
+body of work; no such row can exist yet at migration time, for a pre-existing ``Tag``/``Taxonomy``
+row or otherwise. The field's plain ``True`` default is therefore already correct for every
+existing row; the migration adding the column needs no accompanying data migration.
+
+The flip is monotonic and can be set synchronously
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``deletable`` only ever transitions ``True`` to ``False``, never back. Learner status rows are
+append-only and never deleted (:ref:`openedx-learning-adr-0003` Decision 5), so once a record is
+referenced by a learner status row, it can never become deletable again. This is what makes a
+persisted column safe here without a general cache-invalidation scheme: there is no code path that
+would ever need to flip it back to ``True``.
+
+Because the transition is one-directional, it can be set synchronously, in the same transaction as
+the write that creates the referencing status row: whichever ``Taxonomy``/``CompetencyTaxonomy``,
+``Tag``, ``ObjectTag``, ``CompetencyCriteriaGroup``, or ``CompetencyCriteria`` row that status row
+references gets ``deletable`` set to ``False`` as part of the same write. There is no
+eventual-consistency gap where a GET could show ``deletable=True`` after the guardrail already
+applies.
+
+Enforcement stays on the write path
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The persisted field is a UX convenience, not the enforcement mechanism. DELETE endpoints re-check
+``deletable`` server-side before deleting; a client is never trusted to have an up-to-date copy.
+If a delete is blocked, the handler may run one additional query at that point to identify the
+specific blocking learner-status row(s) for the error message (``409 Conflict``, naming what
+blocks it). Deletes are rare, so this extra query is acceptable exactly where a per-request
+existence check on every read would not be.
+
+Rejected Alternatives
+----------------------
+
+Compute ``deletable`` at read time via ``Exists()``/``OuterRef()``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Annotate the queryset for each GET endpoint with a correlated ``Exists()`` subquery per row,
+computing the signal fresh on every request instead of persisting it.
+
+**Pros:** always correct by construction, no write-path bookkeeping needed, no migration
+considerations.
+
+**Cons:** adds a JOIN to every list/detail read of these models to answer a question that's only
+relevant on the rare delete attempt. Not chosen, since reads vastly outnumber deletes for this
+data.
+
+Cache the computed value with a generous TTL
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Keep the ``Exists()`` computation, but cache its result instead of persisting it as a column.
+
+**Pros:** no schema change, no write-path coupling to learner-status writes.
+
+**Cons:** still requires the same per-row ``Exists()`` check to populate the cache on a miss, and
+introduces a staleness window and invalidation question a persisted, monotonic field avoids by
+construction: since the value only ever moves ``True`` to ``False`` and never back, there is no
+"is my cache stale" question to answer at all. Not chosen, since it solves a narrower version of
+the same problem the persisted field solves outright.
+
+Changelog
+---------
+
+2026-07-15:
+
+* Initial draft.


openedx-webhooks:


</details>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





</details><details>


---

</details>

- The need for product review

</details>

bradenmacdonald:
*If* we really need to show per-tag deletable status in the taxonomy tag list / editor UI, _and_ that deletability is primarily related to this criteria (whether or not any related learner status exists in ``StudentCompetencyStatus``,
``StudentCompetencyCriteriaGroupStatus``, or ``StudentCompetencyCriteriaStatus``), then I think this is a good approach to solving the problem efficiently. However, I'm not sure either of those assumptions are true. And I'm not a big fan of the vagueness of "deletable" nor how it's not really clear where/why/when that field is set, unless you happen to be familiar with the competency requirements.

--------

I would personally prefer a simpler approach. Before I get to that, I want to review the status quo:

First of all, note that in the taxonomy UI, we already show how many times each tag is used:

Screenshot 2026-07-20 at 4 26 27 PM

If the usage count is > 0, we probably want to discourage deleting that tag in general (but we could still allow it, unless it's used for learner status of CBE). If the usage count is 0, we definitely don't care if it gets deleted or not, and there's no need to check anything else. So already we see one potential point of confusion: the most basic criteria for "how deletable is this tag?" has to do with `usage_count`, not the value of the proposed `deletable` field.

Second of all, we don't have a UI that supports selecting and deleting multiple tags at once. You must choose one tag to delete, at which point this dialog displays:

Screenshot 2026-07-20 at 4 30 28 PM

**My suggestion** is that _when we load that dialog_, whether it's deleting one tag or a future version that allows you to select multiple tags then click delete, we can make an `is_deletable` REST API request to the server, which hard-deletes the selected tag + associated ObjectTags, and then immediately rolls back the transaction without committing. That will instantly and performantly tell you whether the tag is used in any foreign keys that are blocking the delete via `ON DELETE RESTRICT`.

Screenshot 2026-07-20 at 4 35 04 PM

Screenshot 2026-07-20 at 4 39 15 PM

* No new field required, no potential confusion around what `deletable` is or when/how it gets set
* No need to keep `deletable` in sync with status of external tables.
* Clearly explains to the user why something cannot be deleted, rather than just disabling some "Delete" menu entries
* No expensive per-row Exists() join/subquery.
* Automatically works with any other Django apps that need the same functionality and use foreign keys to ObjectTag/Tag with `ON DELETE RESTRICT`.

However, this approach only works for Tag/Taxonomy, and I'm not sure how good a fit it is for the other things.

mgwozdz-unicon:
Thanks for taking a look and reviewing @bradenmacdonald. The `is_deletable` dry-run (attempt the real delete in a transaction, roll back, let `ON DELETE RESTRICT` do the work) is appealing: no new field, no sync question to maintain, and it gets you the specific blocking record for free.

Before we commit to it, there's a dependency worth flagging. @jesperhodge's #657 (still unreviewed, per its own description) proposes moving the learner-status tables (`StudentCompetencyStatus`, `...GroupStatus`, `...CriteriaStatus`, plus their HISTORY siblings) onto a separate database alias for scale, and updates ADR-0002 Decision 6/7 so those tables reference the definition tables (`Tag`, `CompetencyCriteriaGroup`, `CompetencyCriteria`, etc.) via logical foreign keys with no database-level constraint, declared that way from day one so the schema doesn't have to change if those tables do move later. Delete protection ends up enforced in application code instead of a DB constraint as a result.

If that design lands, there's no `ON DELETE RESTRICT` to trial against for any of the five models this ADR covers, not just the `ObjectTag` case you already flagged, since the constraint is absent from the start rather than only once the tables are physically split out. A hard delete would just succeed at the DB layer regardless of learner status, so the dry-run wouldn't tell us anything reliable.

Given that, I'd like to get consensus on approach here before changing the ADR much further, since which mechanism is even viable depends on how #657 resolves. Curious whether either of you sees a way to keep an FK-based check under a cross-database status table design. If not, I see two app-code options worth weighing: keep this ADR's persisted `deletable` field, or drop it and run a fresh existence check against the status tables only at delete time. The ADR already rules out that same existence check on every list/detail read for its per-request JOIN cost, but running it once, only at the point of deletion, wouldn't have that cost.

bradenmacdonald:
@mgwozdz-unicon Ah, good call out. Yep, let's hold off on a decision until we know which way that is going. I'd definitely prefer to land on an approach that uses foreign key constraints as much as possible, but I guess we have to do what's performant.

jesperhodge:
@bradenmacdonald @mgwozdz-unicon see if you agree with [my comment on 657 here](https://github.com/openedx/openedx-core/pull/657#discussion_r3626640569).
If you don't see a problem with that, it would mean:
- The only table using a db alias and router is `StudentCompetencyCriteriaStatusHistory`. The non-history active `StudentCompetencyCriteriaStatus` table as well as all the other tables mentioned here (group status, mastery) will stay normal, and can have normal foreign keys.
- Be mindful of large read / write operations on the `StudentCompetencyCriteriaStatus` table as it's huge and we want to avoid blocking the db. Potentially we can specify using a read replica, if available, like on the `StudentModule` table.
- All very large tables in edx-platform, like `StudentModule`, avoid db constraints to users. For example, in `StudentModule`, `student = models.ForeignKey(User, db_index=True, db_constraint=False, on_delete=models.CASCADE)` specifies `db_constraint=False`. I assume there is a good reason for that, so we probably want to follow that principle.

bradenmacdonald:
@jesperhodge Ok, so if `History` is the only table that's in a separate DB, then we can still expect that the `StudentCompetencyCriteriaStatus` table and others have normal foreign keys and will block the deletes using `ON DELETE RESTRICT` rules, right? That's great, it gives us more options here.

jesperhodge:
> @jesperhodge Ok, so if `History` is the only table that's in a separate DB, then we can still expect that the `StudentCompetencyCriteriaStatus` table and others have normal foreign keys and will block the deletes using `ON DELETE RESTRICT` rules, right? That's great, it gives us more options here.

@bradenmacdonald Yes, assuming I am not missing any necessary performance considerations that I'm not aware of. But other tables like `StudentModule` have a similar size, so anything that works there should be fine here as well.

jesperhodge:
@mgwozdz-unicon @bradenmacdonald I don't have any reservations about Braden's approach.

mgwozdz-unicon:
@jesperhodge @bradenmacdonald posting an update since a lot has moved underneath this ADR since this was last visited.

**`CompetencyCriteria`, `CompetencyCriteriaGroup`, and `ObjectTag` no longer need this ADR.** #674 and #675 have since specced the actual delete mechanism for the first two models: on DELETE, check whether any `StudentCompetencyCriteriaStatus`/`StudentCompetencyCriteriaGroupStatus` row references the record; if none, hard-delete; if one exists, set `archived=True` instead of deleting. That settles the enforcement question this ADR was written next to. The GET-side signal for those two can now just be a plain per-row `Exists()` annotation added directly to #681's endpoint, since that list is scoped to one competency's criteria tree. `ObjectTag` turns out to be in the same position for a different reason: its GET endpoint (`ObjectTagView.get_queryset`, `openedx_tagging/rest_api/v1/views.py:459-484`) only ever returns tags for one object at a time, fetching multiple objects in one request is explicitly not supported yet, so that list is just as small and scoped as #681's. That's implementation detail belonging to each of those three endpoints, not a design decision needing its own ADR, so I'm proposing to narrow this ADR to just `Tag` and `Taxonomy`/`CompetencyTaxonomy`.

**For those two, Jesper's original cost concern is still the live question, and I want to bring back a concrete data point.** Jesper, your July 13th comment raised two separate points: reading happens far more than deleting, so adding a per-row JOIN to the `Tag`/`Taxonomy` list endpoint (`TaxonomyTagsView`, used throughout Studio's tagging UI, not just this feature) is worth avoiding, and if we persist a `deletable` field instead, that needs real effort to keep correct. That same list endpoint already computes something analogous: `add_usage_counts()` (`openedx_tagging/api.py:529`) adds a per-tag usage count derived from `ObjectTag` to every row in the response. Its docstring says it deliberately does *not* annotate the queryset row-by-row, calling that "a very expensive annotation," and instead runs one query for the whole page of tags up front, then attaches each tag's count in Python before the response goes out.

That's a real precedent against a per-row `Exists()` annotation for `deletable`, for the same reason, but it also shows the cheaper alternative: apply the exact same technique to `deletable`. When `TaxonomyTagsView` builds its response, run one additional query up front for that page of tags (which of them have a referencing learner-status row), then set `deletable` on each row from that result, the same way `usage_count` is set today. This is the same "embed the signal in the GET response" approach we already settled on for `CompetencyCriteria`, `CompetencyCriteriaGroup`, and `ObjectTag` above, just computed with one bulk query per page instead of a per-row annotation, since this list is hit far more often. `deletable` is also a plain existence check with no lineage traversal or dedup, so it should cost meaningfully less than `usage_count` already does on this same endpoint today. That resolves the read-cost point without touching the write path at all, which also removes the "needs real effort to keep correct" concern on the persisted-field option, since there's no persisted field or write-side bookkeeping involved.

Given the narrower scope, I'd like input on one thing: Does the bulk-computed-in-Python approach address your read-cost concern for `Tag`/`Taxonomy`, or is there a reason it still doesn't hold up at real taxonomy scale?

Separately, if we settle on this approach, I'm thinking that I'll close this ADR PR. At that point it's implementation detail on the same footing as #674/#675, not a design decision that needs a standalone ADR, in line with trying to keep implementation detail out of ADRs.

Tagging in case others want to weigh in: @kdmccormick @ormsbee .

kdmccormick:
just catching up on this. let me try to summarize my understanding:

* we want to allow Tag deletions, unless they are associated with 1+ LearnerCompetencyCriteriaStatuses, in which case we want to block the deletion.
* we already surface the `usage_count` each of Tag in the Taxonomy editor. this is an on-demand computation, not a persisted one.
* there is a modal which warns you when deleting a Tag with usages. in this modal, we can compute on-demand whether the Tag is associated with LearnerCompetencyCriteriaStatuses, and if it is, we can block the delete?
  * how do we compute this? two viable ways:
    1. compute delete-ability on-demand just like we do with `usage_count`
    2. presuming we use django's built in `on_delete=RESTRICT` feature, we can do a dry-run test of deleting it; a RestrictedException will be raised if there are any associated LearnerCompetencyCriteriaStatuses. the "dry-run" is orchestrated by doing the deletion in a transaction and then rolling back the transaction whether or not there's an RestrictedException.
       * there was a concern that this wouldn't be viable if competency statuses are stored in a separate db, but we recently decided that we will **not** be storing anything in a separate db.
* in any case we will **not** build and maintain a bespoke `deletable` row (I am strongly in favor of this decision)

assuming that's directionally correct, I'm in favor of closing this ADR 👍🏻

mgwozdz-unicon:
@kdmccormick I'll respond to each of your points inline.

> just catching up on this. let me try to summarize my understanding:
> 
> * we want to allow Tag deletions, unless they are associated with 1+ LearnerCompetencyCriteriaStatuses, in which case we want to block the deletion.

Yes.

> * we already surface the `usage_count` each of Tag in the Taxonomy editor. this is an on-demand computation, not a persisted one.

Yes, and worth being precise about the cost comparison: `usage_count` is a per-tag `COUNT` over potentially many `ObjectTag` rows, while `deletable` is a plain existence check with no counting or dedup, so it should cost meaningfully less than `usage_count` already does on this same endpoint. Both get the same page-at-a-time treatment either way: one bulk query for the page, attached to each row in Python, not a per-row database annotation.

> * there is a modal which warns you when deleting a Tag with usages. in this modal, we can compute on-demand whether the Tag is associated with LearnerCompetencyCriteriaStatuses, and if it is, we can block the delete?
>   
>   * how do we compute this? two viable ways:
>     
>     1. compute delete-ability on-demand just like we do with `usage_count`
>     2. presuming we use django's built in `on_delete=RESTRICT` feature, we can do a dry-run test of deleting it; a RestrictedException will be raised if there are any associated LearnerCompetencyCriteriaStatuses. the "dry-run" is orchestrated by doing the deletion in a transaction and then rolling back the transaction whether or not there's an RestrictedException.
>        
>        * there was a concern that this wouldn't be viable if competency statuses are stored in a separate db, but we recently decided that we will **not** be storing anything in a separate db.

The reason for putting `deletable` on the GET response, rather than checking at delete time, is so that Studio can disable the delete action before the user ever opens a confirmation dialog, instead of only surfacing the block after they've already tried. Whether the frontend disables the control outright or shows it enabled with a warning explaining why (e.g. that deleting this tag would remove the record of learners who've demonstrated mastery of it) is an open question (likely to be settled in #655), but either way the API needs to expose the same signal up front. My own preference is disabling with an explanation over letting someone click through to a failed attempt.

To be clear on the mechanism itself: we compute it before the frontend gets the list, the same way `usage_count` already is, since that's what makes disabling the control possible at all. A dry-run at delete-time only gets you a more specific error after the fact, it doesn't unlock anything the bulk computation doesn't already give us. And ADR-0002 currently blocks delete outright for Tags and Taxonomies once a learner status exists for them, so we wouldn't be showing a modal to warn the user and then let them choose to do something destructive despite the warning.

> * in any case we will **not** build and maintain a bespoke `deletable` row (I am strongly in favor of this decision)

Yes.

jesperhodge:
@mgwozdz-unicon sounds like a thoughtful solution to me. I don't have any objections.

kdmccormick:
Gotcha, thanks @mgwozdz-unicon . Still no objections from me.

mgwozdz-unicon:
Closing this PR. The persisted `deletable` field it proposed (ADR-0005) was the original answer to the read-cost concern jesperhodge raised on #655: adding a per-row database check to `TaxonomyTagsView`, a high-traffic endpoint, isn't worth it. After more discussion, we believe that the read-cost concern can instead be addressed by computing the signal in bulk per page and attaching it in Python, the same technique `TaxonomyTagsView` already uses for `usage_count` (`openedx_tagging/api.py:529`), with no persisted field or write-side bookkeeping needed.

Full history of how this proposal evolved, and the current state of the broader delete guardrails work, is in two comments just posted on #655: [here](https://github.com/openedx/openedx-core/issues/655#issuecomment-5361337783) and [here](https://github.com/openedx/openedx-core/issues/655#issuecomment-5361338847).

Issue #642, opened 2026-07-06
Mastery status lookup + learner progress models


jesperhodge:
# Mastery status lookup + learner progress models

## User Story

As a learner, I want my current mastery status recorded for each competency, criteria
group, and individual criterion, in order to see which requirements I have already met and
which are still outstanding.

## Acceptance Criteria

- [ ] The three mastery status values, `AttemptedNotDemonstrated`, `PartiallyAttempted` and `Demonstrated`, exist and their order is available to the database, so that raising a status can be written as one conditional `UPDATE` rather than a read followed by a write. A test asserts that a write of a lower value against a higher stored value changes no row, using a single statement.
- [ ] `StudentCompetencyStatus` rejects `AttemptedNotDemonstrated` and accepts only `Demonstrated` and `PartiallyAttempted`. The rejection holds on every write path, including `QuerySet.update()` and `bulk_create()`, which never call `clean()`. Tests cover a direct save and a bulk write.
- [ ] Learner status rows are updated in place, one row per learner and node under a unique constraint, per ADR-0003 Decision 5 and ADR-0002 Decision 6, which lists `created` and `modified` on all three tables. Each table carries both `created` (`auto_now_add=True`) and `modified` (`auto_now=True`). No history package is applied. #613 cites this as "ADR-0003 Decision 5 as amended on 2026-07-27"; that changelog entry no longer exists, because ADR-0003 Decision 5 was rewritten again afterwards. The requirement is unchanged, only the citation.
- [ ] No monotone-write logic and no staff-edit path land here. The models accept any status value the caller writes; the rules that decide which writes are allowed, that an automatic write may raise a status but never lower it and that a staff correction may lower one, are enforced in the API layer. Those two rules now live only in ADR-0004 Decisions 4 and 6. #613 also attributes them to ADR-0003 Decision 5, which no longer contains them.
[...]

Issue #641, opened 2026-07-06
Competency criteria models (authoring/definition layer)


jesperhodge:
# Competency criteria models (authoring/definition layer)

## User Story

As a course author, I want a competency's completion requirements stored as an ordered
AND/OR tree of criteria, in order to express a rule like "pass the final and either lab"
rather than a single flat threshold.

## Acceptance Criteria

- [ ] `CompetencyTaxonomy` has the `taxonomy_overrides_org` boolean, default `false`. The model itself shipped in PR #712 without this column.
- [ ] `CompetencyCriteriaGroup` has all required columns: `id`, `parent_id` (nullable self-FK), `oel_tagging_tag_id`, `course_id` (nullable `ForeignKey` to `openedx_catalog.CourseRun`), `name`, `ordering`, `logic_operator` (AND/OR/null).
- [ ] `openedx_catalog` is added to `.importlinter`'s `root_packages` and placed in the `src_layering` contract below `openedx_learning`. Today it appears in neither, so the first `openedx_learning` to `openedx_catalog` import would pass unexamined. `lint-imports` passes with no rule loosened. It is placed as an independent sibling of `openedx_content`, written `openedx_content | openedx_catalog`, rather than as a layer of its own above or below it: `layers` is a strict total order, so a layer of its own would also decide the catalog-to-content direction that `src/openedx_catalog/ARCHITECTURE.md` records as undecided. #613's wording says only "below `openedx_learning`" and needs the same clarification.
- [ ] `logic_operator` accepts AND, OR or null, per ADR-0002 Decision 2, and nothing at the data layer constrains it by child count. That rule cannot hold here: a group's children need its primary key, so the group's own `clean()` always sees zero children, and adding a child later calls the child's `clean()`, never the parent's. Whatever rule governs null is enforced in the authoring API, when a tree is saved as a unit.
- [ ] No `UniqueConstraint` on `(parent_id, ordering)` is added. ADR-0002 requires none, and it would settle only half the ordering question: a group's children are both child groups and leaf `CompetencyCriteria` rows, and the leaf model has no `ordering` column at all, so sibling order among leaves would stay undefined while looking solved.
- [ ] `CompetencyRuleProfile` has every column from ADR-0002 Decision 3: `id`, `organization_id`, `course_id`, `competency_taxonomy_id`, `scope_code`, `rule_type`, `rule_payload`, `archived`. `rule_payload` is a validated JSON field with shape enforced per `rule_type`.
[...]

Issue #640, opened 2026-07-06
CBE app foundation + CompetencyTaxonomy model


jesperhodge:
## EDIT:

This subtask was mixed up accidentally - it was just a copy of #641 . I had Claude fix it; but I haven't reviewed this in-depth yet.
Either way the parent ticket can be viewed as source of truth.

## Description

**Parent:** #613. This ticket is the foundation of the three-way split: it creates the Django
app the CBE models live in and the single model that both follow-up tickets depend on,
`CompetencyTaxonomy`. See #613 for full background, [ADR-0001](https://github.com/openedx/openedx-core/blob/main/docs/openedx_learning/decisions/0001-competency-criteria-location.rst)
for app placement, and ADR-0002/ADR-0003 for the model.

Nothing else can start until this merges: #641 needs the app to exist and needs
`CompetencyTaxonomy` for `CompetencyRuleProfile.competency_taxonomy_id`; #642 needs the app to
exist. Once this is in, #641 and #642 can be developed in parallel.

## What to build

**1. The `openedx_learning` umbrella app and the `cbe` applet**

ADR-0001 places CBE at `src/openedx_learning/applets/cbe/`, inside a new top-level
`openedx_learning` app that aggregates its applets into one Django app and Python API, the same
way `src/openedx_content` does today. That umbrella package **does not exist in this repo yet** —
ADR-0001 assumed the Learning Pathways workstream would have created it, and it hasn't. So this
ticket creates it:

- `src/openedx_learning/__init__.py`, `src/openedx_learning/apps.py` — one `AppConfig` for the
  whole app (`name`/`label` = `openedx_learning`), mirroring `ContentConfig` in
  `src/openedx_content/apps.py`. Individual applets do not get their own `apps.py`.
- `src/openedx_learning/applets/__init__.py`
- `src/openedx_learning/applets/cbe/__init__.py`, `models.py` (or `models/`), `admin.py`
- `src/openedx_learning/applets/cbe/migrations/0001_initial.py`

**2. Wire it up**

- Add `"openedx_learning"` to `INSTALLED_APPS` in `test_settings.py`.
- Add `openedx_learning` to `root_packages` in `.importlinter`, and place it in the
  `src_layering` contract **above** `openedx_tagging`. The constraint that matters (agreed in
  #613's comments and consistent with #614): the CBE applet may depend on `openedx_tagging`;
  `openedx_tagging` must never depend on or know about CBE.

[...]

Issue #631, opened 2026-07-02
[BE] Build Create endpoint for a Competency Rule Profile (scoped to user's permissions)


jesperhodge:
**Blocked by:** \#613/\#641 (the CBE data model, incl. PR \#636's `CompetencyRuleProfile` additions. Also depends on the one-time `openedx-platform` wiring ticket \#664 establishes (`INSTALLED_APPS` in `cms/envs/common.py`/`lms/envs/common.py` \+ the `cms/urls.py` include registering `openedx_learning.urls`); this endpoint registers inside the same `rest_api/v1/urls.py`, no repeated platform change. Also depends on \*\*6.8\*\* (\`CompetencyCriterion\` reassignment mechanics) for the function this endpoint calls to reassign existing criteria on create.

Repo: `openedx-core`.

## Use Case

As a Platform Administrator (or taxonomy/course author with taxonomy-edit rights), I want to create a Competency Rule Profile scoped to a specific Competency Taxonomy, so that criteria under that taxonomy inherit a default mastery rule (e.g. "score 75% or higher") instead of being configured one by one.

## Description

**Current state**

No `CompetencyRuleProfile` API exists. Per the merged ADR (0002/0003, PR \#636), a rule profile has three nullable scope columns (`organization_id`, `course_id`, `competency_taxonomy_id`, at most one set — enforced by a CHECK constraint and a unique `scope_code`), plus `rule_type`, `rule_payload`, and `archived`. The **system-default** profile is the single all-null-scope row, seeded via migration — it is never created through the API. **Only system-default and taxonomy-scoped profiles are in MVP; organization and course scopes are deferred** (their columns exist but no code creates or reads them yet — `taxonomy_overrides_org` is provisioned but dormant).

**Requested change**

A Create endpoint for a **taxonomy-scoped** rule profile (the only user-created scope in MVP):

- `POST /cbe/rest_api/v1/taxonomies/<taxonomy_id>/rule-profiles/` (competency taxonomy in the URL, not the body).  
- Requires `rule_type` (only `"Grade"` is supported now) and a valid `rule_payload` (for `Grade`: `op` in `gte`/`lte`/`eq`, `value` a 0.0–1.0 fraction).  
- **One profile per taxonomy** Reject a second non-archived profile for the same taxonomy with a clean 400/409, not a raw `IntegrityError` from the `scope_code` unique constraint.  
[...]

PR #712, opened 2026-08-07
feat: provide openedx_learning djangoapp and CompetencyTaxonomy model

jesperhodge:
## AI-first work.

This was coded with human-in-the-loop. I created the initial instructions; Claude Code created a plan, I (Jesper) reviewed the plan and edited as needed; Claude Code implemented, and I reviewed the implementation and tested.

## Related links

Implements #640.

## What this does

Open edX already has "taxonomies", which are just collections of tags. This PR adds a way
to mark a taxonomy as holding **competencies** rather than ordinary tags, so that
competency-based education (CBE) features can later be built on top of it.

- Adds a new Django app, `openedx_learning`, and its first sub-package, `cbe`. This is a
  new home for learning-side features. It follows the same layout as the existing
  `openedx_content` app: one Django app, split into subfolders for readability.
- Adds one model, `CompetencyTaxonomy`. Marking a taxonomy as a competency taxonomy means
  creating one of these. It does not copy the taxonomy or replace it; it points at the
  existing taxonomy and adds a single extra setting to it.
- Adds a small public API so other code can ask "is this taxonomy a competency taxonomy?"
  without needing to know how that link is stored.
- Wires it all up: app registration, the database migration, a Django admin page, and a
  rule that keeps the dependency pointing one way (the CBE code may use the tagging code,
  never the reverse).

There is no REST API and no user interface here. This is the groundwork that the follow-up
tickets #641 and #642 are waiting on.

## Depends on work in edx-platform

openedx-core is a library, so this app only becomes active once edx-platform lists it in
its settings. That platform-side work is tracked in openedx/openedx-platform#38958.

## How to test

This runs entirely inside this repo against a local SQLite file, so you do not need
devstack, tutor, or MySQL. `manage.py` defaults to the `projects.dev` settings, which are
already configured to use SQLite at `./dev.db`, so there is nothing to configure.

**1. Set up a virtualenv** (skip if you already have one for this repo):

```bash
python3 -m venv .venv
.venv/bin/pip install -r requirements/dev.txt
.venv/bin/pip install -e .
```

`.venv/` is already in `.gitignore`.

**2. Create the database and a login.**

```bash
.venv/bin/python manage.py migrate
.venv/bin/python manage.py createsuperuser
```

The first command creates `dev.db` and applies all migrations, including the new
`openedx_learning.0001_initial`.

**3. Start the server.**

```bash
.venv/bin/python manage.py runserver
```

**4. Add a competency taxonomy through the admin.**

Open http://127.0.0.1:8000/admin/ and log in. You should see a section called
**Open edX Core > Learning** containing **Competency Taxonomies**. Click **Add**, fill in
`Name` (for example `Nursing`) and `Export id` (for example `nursing-v1`), and save.

You should land back on a list showing the columns Name, Export id, Enabled, and Taxonomy
overrides org.

**5. Confirm it wrote to both tables.**

```bash
sqlite3 dev.db "SELECT t.id, t.name, c.taxonomy_overrides_org
  FROM oel_tagging_taxonomy t
  JOIN openedx_learning_competencytaxonomy c ON c.taxonomy_ptr_id = t.id;"
```

You should get your row back, something like `1|Nursing|0`.

That join returning a result is the thing worth seeing: one save through the admin created
a row in the existing tagging table and a linked row in the new CBE table. That is the
link this PR is really about.

**Cleaning up:** `rm dev.db` and re-run step 2 to start over.


src/openedx_learning/applets/cbe/api.py
```diff
@@ -0,0 +1,38 @@
...
+
+__all__ = [
+    "is_competency_taxonomy",
+    "select_competency_taxonomies",
+]
+
+# The accessor Django generates for the multi-table-inheritance link from Taxonomy to
+# CompetencyTaxonomy. Deliberately private: callers use the functions below rather than
+# spelling this out, so a model rename is a one-line change here and nowhere else.
+_COMPETENCY_TAXONOMY_RELATION = "competencytaxonomy"
```
kdmccormick: i think that this is unnecessarily indirect, and as an abstraction it's leaky. you cannot stop callers from accessing `taxonomy.competencytaxonomy` directly, so renaming the CompetencyTaxonomy model would be a breaking change--we'd best treat it that way rather than pretending that all references to `.competencytaxonomy` will go through the two functions below.

I would just use the `"competencytaxonomy"` string literal in the function below, as is idiomatic in django code. mypy is sometimes actually smart enough to do type checking on expressions like `taxonomies.select_related("competencytaxonomy")`, but when it's been abstracted out to `taxonomies.select_related(_COMPETENCY_TAXONOMY_RELATION)`, it will never be able to do any static analysis.

src/openedx_learning/applets/cbe/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+    .. no_pii:
+    """
+
+    taxonomy_overrides_org = models.BooleanField(
+        default=False,
+        help_text=_(
+            "When both an organization-scoped and a taxonomy-scoped rule profile "
+            "could apply to a criterion, this decides which one is assigned: false "
+            "assigns the organization's, true assigns this taxonomy's."
```
kdmccormick: could you add a note stating that this isn't yet used anywhere?  bonus points if you add a comment linking to a task or epic issue that would implement it.

kdmccormick (changes requested):
two stylistic comments, but otherwise this looks good to me.

@ormsbee let me know if you'd like to do a pass before I approve and merge.

src/openedx_learning/applets/cbe/models.py
mgwozdz-unicon: We do not have a task or epic that would implement this since it's out of scope. Epic 6 is what most closely relates to it though, but we don't currently have Epic level Github Issues, so that's a handful of issues. The one most closely related is probably #631 though.

kdmccormick: If it's out of scope for Willow, then can we omit the field for now @mgwozdz-unicon ?

mgwozdz-unicon: I think it would be fine to omit the field for now.

jesperhodge: Removed

src/openedx_learning/applets/cbe/api.py
jesperhodge: Fixed

mgwozdz-unicon (approved):
Thanks for addressing all of these. I've reviewed it again and this looks good to me.

On your question: Sonnet 5, high reasoning effort. I didn't use the /code-review skill here, my prompt included a specific list of questions plus an open "anything else worth flagging?", and that didn't happen to trigger it to use the skill (though I've found on my other projects that it often does), but I didn't force it to use the skill afterward either. This particular point came out of that open-ended last question rather than a targeted look at that one method. Going forward I'll make a point of invoking /code-review explicitly for these.

It's called out separately from my own points since it came straight from Claude as a flag to double-check rather than a verified finding. This time it was wrong on the specific exception type, but right that it was worth a test rather than an assumption.

src/openedx_learning/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+Learning-domain features for Open edX Core.
+
+Currently one applet, cbe, holding the Competency-Based Education models.
```
kdmccormick: this is bound to go out-of-date, let's omit it.
```suggestion
```

bradenmacdonald (commented):
From a quick look, LGTM! I didn't do a thorough review.

tests/openedx_learning/applets/cbe/test_models.py
```diff
@@ -0,0 +1,62 @@
...
+import pytest
+
+from openedx_learning.models import CompetencyTaxonomy
+from openedx_tagging.models import Taxonomy
+
+pytestmark = pytest.mark.django_db
+
+# The default MTI reverse accessor. django-stubs cannot see dynamically added
+# accessors, so these tests reach it by name; that name is the ADR-0013 contract.
+RELATION = "competencytaxonomy"
```
kdmccormick: again, this is strangely indirect. just access `.competencytaxonomy` like app code would. if we messed up, then there'll be an AttributeError.

src/openedx_learning/README.rst
```diff
@@ -0,0 +1,11 @@
+Learning App
```
kdmccormick, line 1: Please bump the version to 1.3.0 https://github.com/openedx/openedx-core/blob/main/src/openedx_core/__init__.py

Going forward: do a patch bump for every bugfix, a minor bump for every backwards-compatible feature, and a major bump for every breaking change. (We'll soon have tooling set up to do this automatically based on conventional commits -- we're testing that out on other repos first).

kdmccormick (approved):
two small nits but if you agree with those, then good to go 👍🏻

PR #713, opened 2026-08-07
docs: ADR 0004 competency mastery concurrency

jesperhodge:
Splits ADR 0004 out of #657 so it can be reviewed on its own. ADR 0005 is **not** included here.

## What's in this PR

* **New:** `docs/openedx_learning/decisions/0004-competency-mastery-concurrency.rst`, unchanged from #657.
* **ADR 0002:** the learner-status index list (paired ACTIVE/HISTORY indexes, including the unique leaf HISTORY index that ADR 0004 cites as the idempotency key for its append), the list of append-only history tables, and the matching changelog entry.
* **ADR 0003:** Decision 5 reworked to in-place, monotonic learner status updates, new Decision 6 for the append-only HISTORY tables, and the matching changelog entry.

Changes to 0002/0003 were filtered from #657, not rewritten. Left behind as ADR-0005-only or unrelated:

* pointing the learner-status `user_id` foreign keys at `settings.AUTH_USER_MODEL` (that comes from ADR 0005's rejected alternative 7)
* a whitespace-only fix in ADR 0002

## Known gap

ADR 0004 references `:ref:`openedx-learning-adr-0005`` in five places, and the 0002/0003 edits reference it in four more. Until ADR 0005 lands, those are undefined labels: `sphinx` emits nine `ref.ref` warnings. The build still succeeds (`SPHINXOPTS = -W` is commented out in `tox.ini`), but the links are dead. Either 0005 merges soon after this, or the references need rewording.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/openedx_learning/decisions/0002-competency-criteria-model.rst
+++ b/docs/openedx_learning/decisions/0002-competency-criteria-model.rst
@@ -241,13 +241,10 @@ Decision
    4. ``CompetencyCriteria(oel_tagging_objecttag_id)``
    5. ``CompetencyCriteria(competency_criteria_group_id)``
    6. ``StudentCompetencyCriteriaStatus(user_id, competency_criteria_id)`` (unique)
-   7. ``StudentCompetencyCriteriaStatusHistory(user_id, competency_criteria_id, status_id)`` (unique -- at most one HISTORY row per learner, leaf, and status level, which also serves as the idempotency key for the append in :ref:`openedx-learning-adr-0004`)
-   8. ``StudentCompetencyCriteriaGroupStatus(user_id, competency_criteria_group_id)`` (unique)
-   9. ``StudentCompetencyCriteriaGroupStatusHistory(user_id, competency_criteria_group_id)``
-   10. ``StudentCompetencyStatus(user_id, oel_tagging_tag_id)`` (unique)
-   11. ``StudentCompetencyStatusHistory(user_id, oel_tagging_tag_id)``
-   12. ``CompetencyRuleProfile(scope_code)`` (unique -- at most one profile per distinct scope value; a plain unique constraint on the three raw nullable scope columns would not enforce this, since SQL never treats two ``NULL`` values as equal and this project's MySQL backend does not support the conditional/partial unique indexes that would otherwise route around that; see the ``scope_code`` column in Decision 3)
-   13. ``CompetencyMasteryStatuses(status)`` (unique)
+   7. ``StudentCompetencyCriteriaGroupStatus(user_id, competency_criteria_group_id)`` (unique)
+   8. ``StudentCompetencyStatus(user_id, oel_tagging_tag_id)`` (unique)
+   9. ``CompetencyRuleProfile(scope_code)`` (unique -- at most one profile per distinct scope value; a plain unique constraint on the three raw nullable scope columns would not enforce this, since SQL never treats two ``NULL`` values as equal and this project's MySQL backend does not support the conditional/partial unique indexes that would otherwise route around that; see the ``scope_code`` column in Decision 3)
+   10. ``CompetencyMasteryStatuses(status)`` (unique)
 
 6. Learner progress status concepts (``StudentCompetency*Status`` database tables)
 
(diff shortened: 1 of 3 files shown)


mgwozdz-unicon (changes requested):
Thanks for the quick turnaround. I have 7 items from my review as well as some final considerations for us.

**1. Read-after-write isn't reflected anywhere in the ADR.** It was one of the two options coming out of the core arch meeting, alongside dirty-marking plus a scheduled job. Right now only one of the two made it in, not even as a rejected alternative. Could you add read-after-write, at minimum as a rejected alternative, with the reasoning for choosing dirty-marking over it? These two may not actually be solving the same problem: read-after-write recovers from a concurrent race within the same request, but doesn't obviously recover from a worker crashing mid-cascade, which is what the scheduled job in Decision 6 is for. If that is the real distinction, say so in the ADR; if the reason is something else, that's fine too, but it would help to have it written down rather than living only in meeting notes and side conversations.

**2. Context's transaction-boundary reasoning conflates two different constraints.** The grade itself is stored in SQL (`PersistentSubsectionGrade`), not MongoDB, and so is the competency status tree, so grade-plus-leaf being one transaction (Decision 1) was never blocked by anything in this paragraph, Mongo/cache/celery/events. The reason the rollup above the leaf isn't also folded into that same transaction is a separate one, already in Rejected Alternative 2's second con: two different grading tasks, each finishing a different assignment for the same group, still race on that group's row no matter how large either task's own transaction is, since each task's transaction hides its writes from the other until it commits. As written, Context reads like the Mongo/cache/celery/events constraint is the whole reason nothing above the leaf is transactional with the grade, when for the SQL-only portion of that question, the real reason is the cross-task race. Worth a sentence separating the two.

**3. Decision 2's correctness argument seems incomplete without Decision 3.** As written, Decision 2 justifies taking no locks by saying whichever writer reads a parent last computes the correct result. But the race described in the Context is exactly a case where the last reader can still compute a too-low value, both writers see the other's assignment as outstanding. What actually prevents that from sticking is Decision 3's rule to store whichever value is higher. Suggest reordering so Decision 3 comes before Decision 2, or adding a one-line forward reference in Decision 2 pointing to Decision 3.

**4. The dirty marker's clear step should say it's atomic.** Decision 5 says clearing the marker is conditional on the value passed up still being current, which is the right idea, but it only prevents the race it's meant to prevent if that check and the clear happen as a single atomic statement, for example `UPDATE ... WHERE value = :read_value`, rather than a read followed by a separate write. Please state that explicitly, the way Decision 3 already calls out "in a single statement" for the raise.

**5. Rejected Alternatives 3 and 4 may be saying the same thing twice.** Alternative 3, a lock on "the learner's root group row," lists as a con that it makes a learner's unrelated competencies wait for each other. That con only makes sense if "root group row" is one row spanning all of a learner's competencies rather than one per competency tree, in which case it's the same lock scope as Alternative 4's "one per learner" option, making that bullet in Alternative 4 redundant. If "root group row" is meant to be per competency instead, the "unrelated competencies wait" con doesn't belong on Alternative 3. Please clarify which it is, since Decision 7 reuses the same "root group row" language, and trim whichever alternative ends up restating the other.

**6. Two sentences read as PR history rather than ADR content:** "This was the previous decision here" in Rejected Alternative 1, and "This ADR originally assumed this was available" in Rejected Alternative 2. Neither means anything to a reader who wasn't following this PR. Suggest cutting both.

**7. Is "writes to file storage" in the Context accurate?** I couldn't find a file-storage write in the current certificate generation path, certificates are web-rendered rather than file-based, but I may be looking at the wrong part of the pipeline. If it isn't accurate, it should come out rather than stay as an uncited claim.

**Final Tradeoffs & Implementation Question:** One more, for discussion here rather than something to write into the ADR: the choice between read-after-write and dirty-marking-plus-scheduled-job seems to come down to one question: is the crash-safety the scheduled sweep buys worth the operational overhead it adds? Read-after-write's downside is that data can be left wrong if a worker crashes mid-cascade, with nothing scheduled to notice or fix it until an unrelated change happens to touch the same node again. The scheduled sweep's downside is real added scope: something has to be built and then run continuously by whoever operates the deployment. Separately, if a scheduled sweep is the direction we stick with, how much that overhead actually is depends on implementation: `instructor_task`'s pattern (a management command cron'd by whatever the operator already runs) is much lighter than `openedx-completion-aggregator`'s (a separate plugin plus scaled-up Celery workers), which is the closer analog to what's being proposed here. I'm curious for everyone's thoughts on these points. 
@jesperhodge @kdmccormick @ormsbee @bradenmacdonald

mgwozdz-unicon (approved):
Looks good, thank you!

ormsbee (approved):

PR #761, opened 2026-08-20
docs: add ADR for pathway catalog and content split

Agrendalath:
This adds an ADR describing a separation of concerns between the marketing and authoring parts of a pathway.

Related PRs:


docs/openedx_learning/decisions/0004-pathway-catalog-content-split.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Decisions
+---------
+
+1. A Pathway is split into two parts:
+
+   - **Catalog Pathway** - the learner-browsable, enrollable thing. It includes the display name, the description
+     shown in the catalog, SEO metadata, and a **Category**: a student-facing label for the kind of Pathway it is
+     (e.g. "Master's Degree", "Annual Training"). If it is specified, learners see the Category instead of the word
+     "Pathway". The Catalog Pathway is **not versioned**.
```
ormsbee: From a code point of view, I think we should always force it to be specified, and we can create a default database entry for "Pathway". If we do it with a hardcoded fallback-in-code, we're more likely to get inconsistent behaviors.

ormsbee (changes requested):
Can you also please add something about the dependency relationship, i.e. `openedx_content` knows about `openedx_catalog`, but not vice-versa? That has some implications elsewhere (e.g. Items are allowed to directly reference CourseRuns, and the mapping of CatalogPathway to Pathway content has to be in `openedx_content`).

docs/openedx_learning/decisions/0004-pathway-catalog-content-split.rst
```diff
@@ -0,0 +1,75 @@
...
+  apart makes that possible without inventing field-level permissions.
+- **Auditability.** Progress and credentials must be judged against the definition that was in effect at the time,
+  which requires versioning the definition - but versioning catalog copy would be pure overhead.
+
+Decisions
+---------
+
+1. A Pathway is split into two parts:
+
+   - **Catalog Pathway** - the learner-browsable, enrollable thing. It includes the display name, the description
```
kdmccormick, line 36: Can you say more about this? It sounds like a credential will tie together a learner, a catalog pathway, and particular content pathway that "implemented" that catalog pathway.

Is that right? Will an enrollment do the same?

docs/openedx_learning/decisions/0004-pathway-catalog-content-split.rst
```diff
@@ -0,0 +1,75 @@
...
+
+- **Different change rates.** The display name, description shown in the catalog, and SEO metadata are revised
+  frequently and casually. The definition of what a learner must do to complete the Pathway changes rarely and
+  deliberately.
+- **Different people doing the editing.** Catalog data is typically maintained by marketing or communications staff;
+  the Pathway definition is maintained by content authors. Both are visible to learners, so the distinction is about
+  who edits what, not about who can see it.
+- **Different permissions follow from that.** We expect instances to want to let marketing staff update catalog
+  copy without granting them the ability to change what learners must complete, and vice versa. Keeping the two
+  apart makes that possible without inventing field-level permissions.
```
bradenmacdonald: > Keeping the two apart makes that possible without inventing field-level permissions.

In our new RBAC world, I would expect that there's not much difference between implementing "field-level" permissions and "model-level" permissions, since we're using Casbin to define whatever roles and permissions make sense for each API. Either way, it's simply permissions and needs to be enforced in the REST API via `permission_classes` as appropriate.

However, I still think this split makes sense for the other reasons you've stated.

bradenmacdonald (approved):
Overall sounds good, though +1 to @ormsbee's comment.

docs/openedx_learning/decisions/0004-pathway-catalog-content-split.rst
Agrendalath: @ormsbee, good point - added in 9218009.

docs/openedx_learning/decisions/0004-pathway-catalog-content-split.rst
Agrendalath: @bradenmacdonald, ah, that's good to know - I'm not very familiar with the recent RBAC changes. Should we remove this point from the ADR?

docs/openedx_learning/decisions/0004-pathway-catalog-content-split.rst
Agrendalath: @kdmccormick, I added more details about the enrollments in 9218009. I will add the details about the credentials to https://github.com/openedx/openedx-core/pull/764, since it's out of scope for the current ADR (and I wanted to avoid forward refs here).

In short:
1. Enrollment binds User and Catalog Pathway, because a learner enrolls in a pathway, not its content - i.e., any active authoring (content) changes are visible to the learner. Therefore, I believe there is no real reason to track, at the DB level, the exact state in which the learner was enrolled.
2. Credential is different - it binds User and a specific version of the pathway content, because a credential is a claim that a particular definition of requirements (that was active at that point in time) was met.

ormsbee (approved):
I plan to merge this tomorrow afternoon (1 week after it was initially opened), unless anyone has objections.

docs/openedx_learning/decisions/0004-pathway-catalog-content-split.rst
bradenmacdonald: I'd say we should add a note that this reason may not be as important with the new RBAC system. But as I said, I still think the overall decision makes sense and is a nice way to structure it.

Agrendalath: Got it - added in 4966440.

PR #762, opened 2026-08-20
docs: add ADR for pathway and pathway item boundaries

Agrendalath:
This adds an ADR defining the boundary between a Pathway and its Items.

Related PRs:


docs/openedx_learning/decisions/0005-pathway-and-pathway-items.rst
```diff
@@ -0,0 +1,62 @@
...
+A Pathway is a set of requirements that a learner works through to earn some larger achievement, such as a
+certificate. Each requirement is represented by a **Pathway Item**. We want to draw the boundary between these two
+concepts so that the Pathway level stays stable while the Item level - and especially the ways Items get fulfilled -
+can evolve.
+
+Decisions
+---------
+
+1. A Pathway holds an ordered list of Pathway Items. The order is author-defined and is the order in which Items are
+   presented to learners. In the future, we plan to also support enforcing the order of completion.
```
ormsbee, line 23: The fact that we know we're going to have to later accommodate much more sophisticated display rules for Pathways makes me think that the Pathway Item ordering is something that's going to be modeled at arm's length, i.e. a M:M simple learner display ordering model, rather than making ordering an intrinsic part of the PathwayItem. With the idea that eventually other more sophisticated M:M models will be added to encode fancier dependency/ordering/grouping rules. Does that sound right to you?

ormsbee (approved):
Minor comment just to get clarification, but no change required.

docs/openedx_learning/decisions/0005-pathway-and-pathway-items.rst
```diff
@@ -0,0 +1,62 @@
...
+Status
+------
+
+Draft
+
+Context
+-------
+
+A Pathway is a set of requirements that a learner works through to earn some larger achievement, such as a
+certificate. Each requirement is represented by a **Pathway Item**. We want to draw the boundary between these two
```
bradenmacdonald, line 15: very low importance: I'm wondering why we're calling each pathway requirement a pathway "Item" and not a pathway "Requirement". Without context "pathway item" could be interpreted as some object representing the pathway itself ("an item, specifically a pathway"), not a sub-component.

Is it because some will be optional in the future, and "optional requirement" sounds contradictory?

bradenmacdonald (approved):

docs/openedx_learning/decisions/0005-pathway-and-pathway-items.rst
Agrendalath: @ormsbee, my idea for the ordered list in the MVP was to allow authors to specify the order of item display (e.g., either the suggested order of completion for learners or the content author's pure preference).
I agree that M:M relations will be reasonable for content groups and prerequisites. However, I'm wondering whether we should actually mix the display order with the prerequisites. If not, perhaps keeping the order directly as a PathwayItem field would be a better choice?

docs/openedx_learning/decisions/0005-pathway-and-pathway-items.rst
Agrendalath: @bradenmacdonald, ah, yes, the "requirement" was worded inaccurately on my end. Items == requirements in the MVP, but in the next iterations, these are meant to be completely different concepts.

> Without context "pathway item" could be interpreted as some object representing the pathway itself ("an item, specifically a pathway"), not a sub-component.

We used to call them "pathway steps", which (in my opinion) is a bit more intuitive. @ormsbee, do you know why we decided to change this name to "pathway items"?

ormsbee: @Agrendalath: I don't remember the rationale, just that it was a deliberate decision.

@cassiezamparini, @jmakowski1123, @crathbun428: Do you folks remember what the discussion was around "Pathway Item" vs. "Pathway Step"? I'll write the justification into the [definitions](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5481594883/Learning+Pathways+Studio+Feature+Set+Product+Requirement+Document+PRD#Scope-and-Definitions) wiki entry.

docs/openedx_learning/decisions/0005-pathway-and-pathway-items.rst
ormsbee: I think we have the same concern—I was just thinking both orderings would be M:M relations. Thinking on it more, I'm fine with putting author-ordering on the PathwayItem directly for student display purposes as you describe.

docs/openedx_learning/decisions/0005-pathway-and-pathway-items.rst
cassiezamparini: I suggested "Pathway Item," and here's the reasoning...

"Requirement" doesn't work because of where we're headed post-MVP. Future releases may let a pathway define something like "complete 2 of 3 items," where individual items become optional. "Optional requirement" is a contradiction, so the term needs to stay neutral.

At this stage we're not surfacing the word in the LMS (item/step), learners will just see copy like "complete in any order." To me, "step" reads as sequential, and this release has no fixed order. So the language is contradictory.

Open to other suggestions, but whatever we decide on needs to account for optional items, any order items, and other item types added in future (components, units etc).

Agrendalath: @cassiezamparini, we're not considering calling them "requirements". I'm just wondering why we abandoned the "step" terminology, as an "optional step" is a valid term. Even in this release, the courses must be displayed in some order (even though it's not enforced), so we do not need to worry about sequential/non-seuqential distinction. A "step" is a rather self-explanatory term here. An "item" can be basically anything related to a pathway.
A few years ago, we used to call XBlocks "items" across the whole platform, which resulted in tens of hours of refactoring. I just want to ensure that we do not make the mistake of using a confusing/too broad name once again. Especially if this is how we refer to these things internally (or in the code).

jmakowski1123: Agree re: Pathway Item for all the reasons Cassie mentions.

cassiezamparini: @Agrendalath 

> we're not considering calling them "requirements"

I know, I was just responding to Braden's query 😄

bradenmacdonald: Thanks for the confirmation everyone :)

PR #764, opened 2026-08-20
docs: add ADR for pathway credentials

Agrendalath:
This adds an ADR defining the initial architecture of the pathway credentials.

Related PRs:


docs/openedx_learning/decisions/0007-pathway-credentials.rst
```diff
@@ -0,0 +1,58 @@
...
+   organization, signature image).
+
+4. **Rendering.** An issued certificate is rendered as an HTML page (printable by the learner), analogous to course
+   certificates. W3C Verifiable Credentials / Open Badges will not be supported in the MVP.
+
+5. **Issued certificates are records.** An issued certificate stores everything needed to display it later: recipient
+   name, Pathway name, issuer, signatories, the criteria met (the list of courses passed), and the date earned. It
+   remains valid and viewable even if the Pathway is later updated or archived.
+
+6. **Events.** Issuing a certificate emits an event, so that instances can report on credentials (e.g. in Aspects).
```
bradenmacdonald: In addition to reporting in Aspects, I would assume it should also send emails both to the learner who earned it and to the team that manages the pathway?

bradenmacdonald (commented):
The decisions make sense, but how will this be implemented? I'm not very familiar with the credentials service / certificates code that exists today, though my impression has always been that it's a bit messy.

Does it already have the primitives you need to issue "pathway" certificates in addition to course certificates?

docs/openedx_learning/decisions/0007-pathway-credentials.rst
Agrendalath: @bradenmacdonald, I added a note about sending an email to the learner in 3ae128c. I don't think we should send it to the managing team, though (e.g., if a course has thousands of learners, checking the credentials manually could be more efficient).

PR #792, opened 2026-08-31
[WIP] Re-implementing Restore to use Pydantic (attempt 2)

ormsbee:
This code will import things through the existing Studio interface. It replaces the existing restore code. The backup code has not been changed as part of this PR. You can also test with:

```
python manage.py cms lp_load [path] [package_ref] [username]
```

Starting point for all the logic is in [`backup_restore.api.load_learning_package`](https://github.com/openedx/openedx-core/pull/554/changes#diff-2e119ea0879d0fd914fed8e74c38e158efcd29dbd949a29c9556bb37fa5b1755R33).

New modules:

* `archive.py` handles the container archive format, though this is mostly just delegating to `fsspec`.
* `payload.py` is where the extraction of TOML files happens, and things get assembled into a big dict for validation. It understands the backup archive's internal format.
* `validation.py` is where that's gets turned into a validated Pydantic model and errors will be accumulated.
* `schema.py` defines the actual Pydantic models. It needs a lot of validation rules (just the bare field stuff exists now).
* `loading.py` is the logic for actually loading validated input into the database. It initializes a `Loader` object with input data and then calls  `loader.load_into()` to push that data into a target `LearningPackage`.

I think I'm comfortable with this as the basic structure. I intend to keep full compatibility with the existing REST endpoint and just do conversions as needed.

One note is that much of `validator.py` was Claude-generated, along with many of the tests. I'm not at all happy with the resulting code in `validation.py`, and I plan to make major revisions before I open this for review.


src/openedx_content/applets/backup_restore/schema.py
```diff
@@ -0,0 +1,292 @@
...
+
+    key: Annotated[str, REF_CONSTRAINTS]
+
+    can_stand_alone: bool = True
+
+    created: AwareDatetime
+
+    # Weird edge case: If you create something, never publish it, and then do a
+    # "reset to published state", the resulting export in Ulmo would omit the
+    # [entity.draft] section entirely, rather than it being an empty dictionary.
```
ormsbee, line 207: Note to self, make sure we're actually testing this edge case.

src/openedx_content/applets/backup_restore/api.py
```diff
@@ -1,42 +1,132 @@
...
+from . import archive, loading, payload, validation
+from .errors import BackupRestoreError, RestoreFailedError
+from .results import RestoreResult, generate_staged_package_ref
+from .zipper import LearningPackageZipper
 
-# The public API that will be re-exported by openedx_content.api
-# is listed in the __all__ entries below. Internal helper functions that are
-# private to this module should start with an underscore. If a function does not
-# start with an underscore AND it is not in __all__, that function is considered
-# to be callable only by other applets in the openedx_content package.
```
ormsbee, line 15: Note: restore this

Issue #716, opened 2026-08-12
[BE] Add archived field to CompetencyCriteriaGroup and CompetencyCriterion


thelmick-unicon:
**Blocked by:** \#613 (the CBE data model these two models belong to).

Repo: `openedx-core`, single-repo. No `openedx-platform` changes.

## Use Case

As a developer implementing CBE's archive-vs-delete behavior across several tickets (\#674, \#675, the create-side archived-branch guard, and eventually \#681), I want `CompetencyCriteriaGroup` and `CompetencyCriterion` to have a settled `archived` field from the start, so that none of those tickets has to race to add it themselves or coordinate around who lands first.

## Description

**Current state**

Neither `CompetencyCriteriaGroup` nor `CompetencyCriterion` has an `archived` field. Multiple downstream tickets each need it to implement ADR 0002 Decision 7's archive-only retirement rule, and were drafted with "whichever lands first adds it" coordination language instead of a single owner.

**Requested change**

Add `archived = models.BooleanField(default=False)` to both `CompetencyCriteriaGroup` and `CompetencyCriterion`, in one migration. No `archived_at` companion field: `django-simple-history` (already applied per ADR 0003\) covers the "when." This ticket makes no behavior changes: nothing here sets the field to `True` or reads/filters on it, that's downstream tickets' scope.

**Explicitly out of scope**

- Setting `archived=True` anywhere.  
- Filtering or checking `archived` anywhere.  
- Any UI.

## Acceptance Criteria

There is no manual QA path. Acceptance is determined by a PR reviewer verifying the following against ADR 0002 and ADR 0003:

- [ ] `archived = models.BooleanField(default=False)` exists on both `CompetencyCriteriaGroup` and `CompetencyCriterion`.  
- [ ] One migration adds both columns.  
- [ ] No `archived_at` column was added.  
- [ ] `django-simple-history` already tracks both models (per ADR 0003); the new field is captured in the history table without further changes.  
- [ ] `archived` is registered as non-PII, matching `CompetencyRuleProfile.archived`'s existing category; `make pii_check` passes.  
- [ ] No backfill needed: greenfield field, no existing rows depend on a different default.

## Context

- \#613: the model definitions this ticket adds a field to.  
[...]

PR #800, opened 2026-09-01
feat: add competency criteria models for CBE authoring layer

### LINKED ISSUE #641: Competency criteria models (authoring/definition layer)

jesperhodge:
Adds the authoring and definition half of the competency-based education (CBE) data model, per
[ADR-0002](https://github.com/openedx/openedx-core/blob/main/docs/openedx_learning/decisions/0002-competency-criteria-model.rst)
and [ADR-0003](https://github.com/openedx/openedx-core/blob/main/docs/openedx_learning/decisions/0003-competency-criteria-versioning.rst).
Three models, one column on an existing model, two migrations, four configuration edits. No REST
endpoints, no UI, no evaluation logic.

Closes #641.

> [!WARNING]
> **The deletion behavior here is work in progress and will change within this PR**, pending
> discussion with @mgwozdz. Three separate things are still open: the individual `on_delete`
> values, the cascade-versus-protect split as a design question, and the archive-versus-delete
> story as a whole. ADR-0002 Decision 7 was amended twice in three days (`f9ec022`, then
> `b5fae6b`) while this branch was being written, and the second amendment reversed the reasoning
> behind the first. Treat the `on_delete` table below as the current position, not a settled one.
> Nothing else in this PR depends on how that discussion resolves.

> [!NOTE]
> Implemented by an AI agent (Claude Code), with a human directing the work and reviewing the
> decisions. Please review it as you would any other PR.

## The three models

- **`CompetencyCriteriaGroup`** is an internal AND/OR node of a criteria tree. A tree hangs off one
  competency, which is a `Tag` in a competency-enabled taxonomy.
- **`CompetencyCriterion`** is a leaf. It points at one `ObjectTag`, meaning one specific piece of
  tagged content, and takes its pass rule either from a shared profile or from its own inline
  override pair.
- **`CompetencyRuleProfile`** is a reusable set of evaluation settings, scoped to at most one of an
  organization, a course, or a taxonomy. One row is scoped to none of them: the **system default**,
  seeded by migration, which every criterion falls back to when nothing more specific applies. In
  this MVP it is the only profile that exists, so all three scope columns are always null.

Plus `CompetencyTaxonomy.taxonomy_overrides_org`, the boolean ADR-0002 Decision 1 asks for.

## Decisions

**`scope_code` is an ordinary column written in `save()`, null when the profile is archived, rather
than a database `GeneratedField`.** At most one profile may exist per distinct scope, and a unique
constraint over the three nullable scope columns cannot enforce that, because SQL never treats two
`NULL`s as equal. The idiomatic fix, a conditional `UniqueConstraint`, compiles to a partial index
that MySQL silently skips (ADR-0002 Rejected Alternative 6). Deriving the value works, but deriving
it *in the database* broke twice: an archived profile kept occupying its scope's unique slot, so no
replacement could ever be created for that scope; and Django's collector nulls a nullable foreign
key before deleting the row it points at, which recomputed `scope_code` mid-delete and collided with
the seeded default row. The collector only does this where `can_defer_constraint_checks` is false.
MySQL has that flag false and SQLite has it true, so this failed only on MySQL. Writing the column
in `save()` fixes both: the collector's update no longer rewrites it, and archived rows carry
`NULL`, so any number of them share a scope while exactly one live row holds it, identically on
every backend. A `CheckConstraint` ties `archived` and `scope_code` together, so a
`QuerySet.update()` bypassing `save()` is refused by the database rather than silently breaking the
invariant. This deviates from #641, which asks for a generated, never-null column; ADR-0002
Decision 3 needs a matching amendment.

**Validation and immutability each collapsed to one path.** `save()` now calls `full_clean()` on
both models rather than repeating a hand-picked list of checks that could drift from `clean()`,
following the precedent `CourseRun.save()` sets. Scope immutability drops its cached copy of the
loaded scope and its `from_db()` override in favor of always reading the persisted scope, on
`self._state.db` so a non-default database alias is not silently skipped. The rule payload schema
moved out of the models module into `rule_payloads.py` and now returns the parsed `GradeRule` rather
than discarding it, so #642's evaluation code can consume typed fields without importing five
models. Both models derive their `rule_type` choices from the payload-spec registry, so a rule type
can never be offered to an author and then rejected on save.

**Two `on_delete` values differ from #641, and this is the part under discussion.**
`CompetencyRuleProfile.course` becomes `CASCADE` because `b5fae6b` says a profile is deleted along
with "a taxonomy or course" it is scoped to. `CompetencyCriteriaGroup.course` becomes `CASCADE` for
that amendment's own stated reason: a course is only hard-deleted once nothing beneath it needs
protecting, so a course-scoped criteria tree is safe to remove with it rather than blocking the
delete permanently. `CompetencyRuleProfile.organization` stays `PROTECT`, since the amendment names
only taxonomy and course, and an organization is not a competency-definition record. That asymmetry
is deliberate and is one of the things to settle.

| Foreign key | Value |
| --- | --- |
| `CompetencyCriteriaGroup.tag`, `.parent`, `.course` | `CASCADE` |
| `CompetencyCriterion.group`, `.object_tag` | `CASCADE` |
| `CompetencyRuleProfile.competency_taxonomy`, `.course` | `CASCADE` |
| `CompetencyCriterion.rule_profile` | `PROTECT` |
| `CompetencyRuleProfile.organization` | `PROTECT` |

## Other deviations from #641

- `.importlinter` ranks `openedx_content` above `openedx_catalog` rather than making them
  independent siblings. The sibling form forbids imports both ways, including the direction
  `0007-pathway-catalog-content-split.rst` requires, so it would have to be loosened to do an
  already-decided thing.
- `RuleType` declares only `Grade`. ADR-0002 also names `View` and `MasteryLevel`, but neither has a
  payload shape, so declaring them offers an author a choice that always fails on save. Adding one
  later means a spec class, a registry entry, and the matching member together.

## Known gaps

- Deleting a `CompetencyTaxonomy` whose taxonomy-scoped profile is assigned to a criterion raises
  `ProtectedError`. Django's collector looks up referencing rows in the database rather than in the
  set it has already decided to delete, so `CompetencyCriterion.rule_profile`'s `PROTECT` fires even
  for criteria being deleted in the same operation. This is unreachable in the MVP, where the only
  profile is the system default. A test pins it and names the fix: a fifth reassignment event in
  ADR-0002 Decision 4, in an application-layer function.
- No test proves the scope-immutability query targets `self._state.db`. That needs a second database
  alias, and configuring one breaks the whole test session on a pre-existing bug in
  `openedx_content/backcompat/collections/migrations/0004_collection_key.py`, whose `generate_keys`
  step queries without `.using(schema_editor.connection.alias)` and so always hits `default`.
  Worth its own issue.
- Out of scope: the application-layer archive-and-reassign function, the archive-versus-delete
  branch from #655, the `archived` column on the group and leaf models (#716), and Django admin
  registration.

## Testing

Every acceptance criterion in #641 has a test, written before the implementation. Tests are named
for the behavior they assert rather than the mechanism, and `test_criteria_trees.py` covers
whole-tree deletion, so a test proves the bad outcome is avoided rather than only that a cascade
fired. The deletion paths also run under MySQL's collector semantics while still on SQLite, by
setting `can_defer_constraint_checks` to false, which makes this class of bug visible in the fast
local suite instead of only in CI.

Verified against **both backends**, since a green SQLite run is not evidence for the `scope_code`
work: 887 passed on SQLite, 888 on a real MySQL 8. `mypy`, `pylint`, `pycodestyle`, `pydocstyle` and
`isort` are clean, `lint-imports` keeps both contracts, and `makemigrations --check` reports no drift
in `openedx_learning`. No `# noqa`, `# pylint: disable`, `# type: ignore` or `TODO` anywhere in the
diff.

`make pii_check` still fails at two pre-existing lint conflicts (`openedx_content.Draft`,
`openedx_content.PublishableEntityVersion`), both identical on `main` and in code this PR does not
touch. The models added here are annotated, including the three `Historical*` models
`django-simple-history` generates.


src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+class LogicOperator(models.TextChoices):
+    """How a CompetencyCriteriaGroup combines its child nodes."""
+
+    AND = "AND", _("And")
+    OR = "OR", _("Or")
+
+
+def validate_rule_payload(rule_type: str, payload: Any) -> None:
```
jesperhodge: I don't like `payload: Any`, there should be a type for the dict.
And I want to validate against that type.

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+    loaded. That comparison covers every ``instance.save()``, including one loaded with
+    ``.only()``/``.defer()`` that skipped some scope columns, in which case the comparison falls
+    back to reading the persisted scope directly rather than skipping the check. It does not
+    cover a bulk ``QuerySet.update()``, since that path never loads or constructs a model
+    instance at all.
+
+    .. no_pii:
+    """
+
+    # Set at from_db() time to the scope this row had when it was loaded from the database, so
```
jesperhodge: Comment too hard to understand

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+    # clean()/save() can detect an attempt to change it. None for a newly-constructed instance,
+    # meaning there's nothing yet to compare against. Deliberately not underscore-prefixed:
+    # from_db() is a classmethod, so it sets this through a local `instance` variable rather than
+    # `self`, which pylint's protected-access check can't tell apart from reaching into another
+    # object's internals.
+    loaded_scope: tuple[int | None, int | None, int | None] | None = None
```
jesperhodge: What is loaded_scope and why is it a tuple of ints?

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+
+    organization = models.ForeignKey(
+        Organization,
+        null=True,
+        blank=True,
+        on_delete=models.PROTECT,
```
jesperhodge, line 193: organization & course on_delete should also be CASCADE, with Python logic elsewhere ensuring that no learners are linked to this (if they are linked, organization / course can still be deleted but rule profile stays.)

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+        on_delete=models.CASCADE,
+        related_name="rule_profiles",
+        help_text=_("The competency taxonomy this profile is scoped to, if any."),
+    )
+    # Always non-null, including for the system-default row (all three scope columns null), so a
+    # plain UniqueConstraint on this one column enforces "at most one profile row per distinct
+    # scope" identically on every backend. SQL never treats two NULLs as equal, so a unique
+    # constraint directly on the three nullable scope columns would let e.g. two rows that both
+    # set only organization_id=5 both exist. See ADR-0002 Decision 3.
+    scope_code = models.GeneratedField(
```
jesperhodge: Where's the UniqueConstraint? I don't see anything here or in the migration, but maybe I'm just missing it.

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+        db_column="oel_tagging_tag_id",
+        on_delete=models.CASCADE,
+        related_name="competency_criteria_groups",
+        help_text=_("The competency (tag) that this criteria tree evaluates mastery of."),
+    )
+    course = models.ForeignKey(
+        CourseRun,
+        null=True,
+        blank=True,
+        on_delete=models.PROTECT,
```
jesperhodge: This also needs to be CASCADE but then it should stay if there are any actual learner's already connected to the mastery. Same as elsewhere

src/openedx_learning/applets/cbe/models/criteria.py
jesperhodge: I think this is not quite right. CASCADE should result in archival, not deletion.
Question to flag for later: what should happen if someone actually wants to modify (or rather, delete and then replace) a rule profile? Assuming this gets archived, do the archived ones still need to be unique? In that case they are never modifiable and we need some hard delete or overwrite mechanism, I guess.

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+                    "competency_taxonomy."
+                ),
+            ),
+        ]
+
+    @classmethod
+    def from_db(cls, db, field_names, values):
+        """Capture the scope this row had when loaded, so clean()/save() can detect an edit to it."""
+        instance = super().from_db(db, field_names, values)
+        # field_names holds attnames (e.g. "organization_id"), not field names. Only capture when
```
jesperhodge: I have no idea what this is supposed to mean. Clarify what the intent is here.

jesperhodge: Why do we need to override from_db at all?

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+        scope_attnames = {"organization_id", "course_id", "competency_taxonomy_id"}
+        if scope_attnames.issubset(field_names):
+            instance.loaded_scope = (
+                instance.organization_id,
+                instance.course_id,
+                instance.competency_taxonomy_id,
+            )
+        return instance
+
+    def _check_scope_immutable(self) -> None:
```
jesperhodge, line 294: This seems pretty complicated code. Can it be simplified following KISS principle?

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+        CompetencyRuleProfile,
+        null=True,
+        blank=True,
+        db_column="competency_rule_profile_id",
+        on_delete=models.PROTECT,
+        related_name="criteria",
+        help_text=_("The profile this criterion uses by default. Null only when overrides are set instead."),
+    )
+    rule_type_override = models.CharField(max_length=32, choices=RuleType, null=True, blank=True)
+    rule_payload_override = models.JSONField(null=True, blank=True)
```
jesperhodge, line 395: Use `attrs`, as in my other comment.

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+        on_delete=models.CASCADE,
+        related_name="competency_criteria",
+        help_text=_("The tag/object association that this criterion evaluates."),
+    )
+    rule_profile = models.ForeignKey(
+        CompetencyRuleProfile,
+        null=True,
+        blank=True,
+        db_column="competency_rule_profile_id",
+        on_delete=models.PROTECT,
```
jesperhodge, line 390: I think this is okay because rule profiles should be archived not deleted in general.

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,461 @@
...
+    archived = models.BooleanField(
+        default=False,
+        help_text=_(
+            "Set instead of deleting a profile that's no longer wanted. Archived profiles are hidden from "
+            "authoring and new associations but remain queryable, so existing criteria stay resolvable."
+        ),
+    )
+    uuid = immutable_uuid_field()
+
+    history = HistoricalRecords(excluded_fields=["scope_code"])
```
jesperhodge, line 253: I wonder if the history and the `archived` flag somehow conflict with each other.

Also I wonder about this immutability idea anyway: if this should be immutable why a history?
Maybe immutability is not a clear concept in the issue. This will need further clarification from the architect.

tests/openedx_learning/applets/cbe/test_criteria_deletion.py
```diff
@@ -0,0 +1,348 @@
...
+# real trap of that kind, where CatalogCourse.org is also PROTECT), so a bare
+# `pytest.raises(ProtectedError)` would not actually prove which foreign key did the protecting.
+# The five that are CASCADE assert the delete succeeds and that the referencing row is actually
+# gone from the database afterward, not merely that no exception was raised, and assert the
+# referencing row existed beforehand, so the "gone" assertion can't pass because a fixture never
+# created it in the first place.
+# ==============================================================================================
+
+
+def test_group_parent_cascade(tag: Tag) -> None:
```
jesperhodge: Test names are bad. They should all state the expected behavior. I don't care if that makes them long. E.g. "test_criteria_group_deletion" is bad, while "test_delete_criteria_group_cascades_to_child_groups" is good.

jesperhodge: Here's some reading for proper naming of tests. https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices

tests/openedx_learning/applets/cbe/test_criteria_deletion.py
```diff
@@ -0,0 +1,348 @@
...
+    """
+    Deleting a CompetencyCriteriaGroup cascades to any child group referencing it via `parent`:
```
jesperhodge, line 65: At least some of the tests need to be a bit more integrative. In that: sure, we have tested that the cascade is there, but it's not clear why. There should be at least some tests that look at the actual bad outcome that we want to avoid: for example, do we suddenly have orphaned child groups that do not serve any purpose?

tests/openedx_learning/applets/cbe/test_criteria_deletion.py
```diff
@@ -0,0 +1,348 @@
...
+
+    with pytest.raises(ProtectedError) as exc_info:
+        default_rule_profile.delete()
+
+    protected = exc_info.value.protected_objects
+    assert any(isinstance(obj, CompetencyCriterion) and obj.pk == criterion.pk for obj in protected)
+
+
+# ==============================================================================================
+# Transitive deletion tests required by #641's Deletions criteria: deleting an oel_tagging.Tag,
```
jesperhodge, line 401: Too unclear. I have no patience to decipher what this comment means. Either it's clear at a glance or it's useless.

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -131,7 +131,7 @@ Decision
...
+   5. ``scope_code``: A plain column, recomputed by the model's ``save()`` and never set directly, in the fixed, trivially-parseable format ``"org:X,course:Y,taxonomy:Z"``, with each segment left blank when the corresponding scope column is null: for example ``"org:5,course:,taxonomy:"``, or ``"org:,course:,taxonomy:"`` for the system default row. It is null while a profile is archived, and non-null while it is live. Collapsing the scope into one column exists because SQL never treats two ``NULL`` values as equal for uniqueness purposes, so a plain unique constraint across the three nullable scope columns would not stop two rows from sharing a scope. Nulling it while archived is what frees an archived profile's scope for a replacement, without needing the conditional unique index MySQL does not support. It is a plain column rather than a ``GeneratedField`` because Django's delete collector nulls a nullable cascading foreign key before issuing the DELETE on backends that cannot defer constraint checks, and a generated column would recompute from that nulled value and collide with whichever row already holds the resulting blank scope. A plain column is untouched by that nulling.
```
jesperhodge, line 134: But isn't archiving an operation that is intended to possibly restored later? If we remove the scope, don't we loose that data? Maybe we should do `"org:X,course:Y,taxonomy:Z" with possible blank segments, but for archived items, we do something like "org:X,course:Y,taxonomy:Z,archive-version:1" to keep it restorable and unique

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -309,6 +309,8 @@ Decision
...
+   - ``on_delete`` on the criteria tables expresses containment, not protection: a row whose referent is gone is meaningless, so ``CompetencyCriteriaGroup.parent``, ``.tag`` and ``.course``, ``CompetencyCriterion.group`` and ``.object_tag``, and ``CompetencyRuleProfile.course`` and ``.competency_taxonomy`` all cascade. ``CompetencyCriterion.rule_profile`` stays ``PROTECT``, which is what makes "a profile is never hard-deleted by a direct delete" hold at the ORM layer. ``CompetencyRuleProfile.organization`` stays ``PROTECT`` because an ``Organization`` is not a competency definition record and ``edx-organizations`` deactivates organizations rather than deleting them. The tree links additionally have to cascade for a mechanical reason: Django's collector looks up referencing rows in the database rather than in the set it has already decided to delete, so a parent and child reached in the same batch would still trip ``PROTECT`` and abort the walk partway down. Those cascading edges are what carries a delete down to the ``PROTECT`` on the learner status tables, which is where this decision is actually enforced.
```
jesperhodge, line 312: Unreadable, unclear

docs/openedx_learning/decisions/0002-competency-criteria-model.rst
```diff
@@ -309,6 +309,8 @@ Decision
...
+   - Known limitation: deleting a ``CompetencyTaxonomy`` whose taxonomy-scoped profile is assigned to a ``CompetencyCriterion`` raises ``ProtectedError`` naming that criterion, even though the criterion would also be cascade-deleted in the same operation through the tag chain, for the same collector reason above. This is unreachable until scoped profiles can be authored. The fix at that point is a fifth reassignment event on Decision 4: when a profile's scope owner is being deleted, reassign every criterion off that profile before the cascade proceeds.
```
jesperhodge, line 313: Needs review

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+    Decision 3 for how a :class:`CompetencyCriterion` is assigned one of these, and Decision 4 for
+    what happens when more than one scope's profile could apply to the same criterion.
+
+    Editing a profile may change ``rule_type``/``rule_payload`` only: the scope fields
+    (``organization``, ``course``, ``competency_taxonomy``) are immutable after creation, so that
+    criteria already resolved to this profile's scope are never silently re-governed. ``clean()``
+    enforces this by comparing the current scope columns against what is actually persisted for
+    this row, so the check holds regardless of whether this instance was loaded with a partial
+    ``.only()``/``.defer()`` that skipped some scope columns. It does not cover a bulk
+    ``QuerySet.update()``, since that path never loads or constructs a model instance at all.
```
jesperhodge, line 132:
```suggestion
```

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+
+    ``rule_payload``'s shape (see :func:`~openedx_learning.applets.cbe.rule_payloads.validate_rule_payload`)
+    is likewise validated from ``clean()``, reached from both ``objects.create()`` and a plain
+    ``instance.save()`` via ``full_clean()``. A bulk ``QuerySet.update()``, ``bulk_create()``, and a
+    DRF serializer that writes straight to the database are NOT covered: none of them build or save
+    a model instance, so ``clean()`` never runs.
```
jesperhodge, line 138:
```suggestion
```

src/openedx_learning/applets/cbe/models/criteria.py
jesperhodge: Unnecessary explanation

src/openedx_learning/applets/cbe/models/criteria.py
jesperhodge: Unnecessary explanation

src/openedx_learning/applets/cbe/models/criteria.py
jesperhodge: PROTECT is no good. The org delete shouldn't be blocked. Instead, use `models.SET()` to run code to archive the rule_profile, but only if no learners are connected.

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+        blank=True,
+        on_delete=models.PROTECT,
+        related_name="competency_rule_profiles",
+        help_text=_("The organization this profile is scoped to, if any."),
+    )
+    course = models.ForeignKey(
+        CourseRun,
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
```
jesperhodge, line 156: Use `models.SET()` and archive. only if there are no learners connected

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+        related_name="competency_rule_profiles",
+        help_text=_("The course run this profile is scoped to, if any."),
+    )
+    competency_taxonomy = models.ForeignKey(
+        CompetencyTaxonomy,
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
```
jesperhodge, line 164: Maybe.

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+            ),
+        ]
+
+    def _check_scope_immutable(self) -> None:
+        """Raise ValidationError if the scope columns no longer match what is persisted for this row."""
+        if self.pk is None:
+            # A new, unsaved instance: there's no persisted scope yet to compare against.
+            return
+        # Queried rather than compared against a value cached at load time, so a deferred load or
+        # a refresh_from_db() cannot bypass the check. `using` keeps a non-default-database
```
jesperhodge, line 238: What are we talking about specifically with non-default-database instance? Why would there be such a thing?

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+        # instance from being compared against the wrong alias.
+        persisted_scope = (
+            CompetencyRuleProfile.objects.using(self._state.db)
```
jesperhodge, line 241:
```suggestion
            CompetencyRuleProfile.objects
```

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+            .filter(pk=self.pk)
+            .values_list("organization_id", "course_id", "competency_taxonomy_id")
```
jesperhodge, line 243: Better take all 3 values and compare them against all 3 current values

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+        super().clean()
+        self._check_scope_immutable()
+        validate_rule_payload(self.rule_type, self.rule_payload)
+
+    def _compute_scope_code(self) -> str | None:
+        """Return this profile's scope_code, or None while it is archived."""
+        if self.archived:
+            return None
+        scope_ids = (self.organization_id, self.course_id, self.competency_taxonomy_id)
+        return "org:{},course:{},taxonomy:{}".format(*("" if pk is None else pk for pk in scope_ids))
```
jesperhodge, line 268:
```suggestion
        return f"org:{self.organization_id},course:{self.course_id},taxonomy:{self.competency_taxonomy_id}"
```

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+
+    def save(self, *args, **kwargs):
+        """Recompute scope_code, then persist this profile after full_clean() re-validates it."""
+        self.scope_code = self._compute_scope_code()
+        # Ensure that we run the validations/defaults defined in clean().
+        # But don't validate_unique(); it just runs extra queries and the database enforces it anyways.
```
jesperhodge, line 274:
```suggestion
        # validate_unique() is already enforced by the database.
```

src/openedx_learning/applets/cbe/models/criteria.py
```diff
@@ -0,0 +1,368 @@
...
+
+    def _compute_scope_code(self) -> str | None:
+        """Return this profile's scope_code, or None while it is archived."""
+        if self.archived:
+            return None
+        scope_ids = (self.organization_id, self.course_id, self.competency_taxonomy_id)
+        return "org:{},course:{},taxonomy:{}".format(*("" if pk is None else pk for pk in scope_ids))
+
+    def save(self, *args, **kwargs):
+        """Recompute scope_code, then persist this profile after full_clean() re-validates it."""
```
jesperhodge, line 271:
```suggestion
        """On save: recompute and validate scope_code"""
```

mgwozdz-unicon (approved):
I reviewed the Stacked PRs here https://github.com/jesperhodge/openedx-core/pull/10 and approved.

Issue #628, opened 2026-07-01
[BE] openedx-core - Accept taxonomy_type on Create & Import endpoints


mgwozdz-unicon:
Part of #614. See parent for use case, acceptance criteria, and architecture rationale.

**Depends on:** data model ticket (CompetencyTaxonomy model must exist) #613 

### Files to modify

| File | Change |
| :---- | :---- |
| `src/openedx_tagging/rest_api/v1/serializers.py` | Add `taxonomy_type` as `write_only=True` `ChoiceField` with `required=False`, `choices=["tags", "competency"]`. Define the valid-value constants in a shared location so this field and the CBE dispatch logic below can't drift apart. |
| `src/openedx_tagging/rest_api/v1/views.py` | Update `perform_create()` and `create_import()` to surface `taxonomy_type` from validated data. Do not call `CompetencyTaxonomy` from this layer. |
| `openedx_learning/applets/cbe/api.py` | Add `create_competency_taxonomy()`: call `create_taxonomy()` then create the linked `CompetencyTaxonomy` row in `transaction.atomic()`. |

### Key constraint

`openedx_tagging` must not import from `openedx_learning`. The dispatch to
`create_competency_taxonomy()` must happen at a layer that can see both packages — a
CBE-provided view mixin overriding `perform_create()` is one viable approach.

#### Example Resolution Prompt

In `src/openedx_tagging/rest_api/v1/serializers.py`, add `taxonomy_type` as a
write-only `ChoiceField` with `required=False` and `choices=["tags", "competency"]`;
define the valid-value constants (e.g. `TAXONOMY_TYPE_COMPETENCY = "competency"`) in a
shared location so this field and the CBE dispatch logic below stay in sync. In
`src/openedx_tagging/rest_api/v1/views.py`, update both `perform_create()` (around line
262) and the `create_import()` action to surface `taxonomy_type` from validated data —
do not call `CompetencyTaxonomy` directly from this layer. In
`openedx_learning/applets/cbe/api.py`, add `create_competency_taxonomy()`: call
`create_taxonomy()` from `openedx_tagging`, then create the linked `CompetencyTaxonomy`
row wrapped in `transaction.atomic()`. Wire the dispatch to
`create_competency_taxonomy()` at a layer that can see both packages — a CBE-provided
view mixin overriding `perform_create()` is one viable approach.

PR #803, opened 2026-09-03
feat: accept taxonomy_type on taxonomy create and import endpoints

alezconsultant:
## Description

Adds `taxonomy_type` to the taxonomy create and import endpoints: `"competency"`
creates a `CompetencyTaxonomy` alongside the base `Taxonomy`, `"tags"` (default)
creates a plain one.

This PR lands the `openedx-core` side only. A companion PR against
`openedx-platform` will wire `TaxonomyOrgView` to `CompetencyTaxonomyView` and
pass `taxonomy_type` through from the client.

## Changes

- `src/openedx_tagging/api.py`: `TaxonomyType` enum (`TAGS`, `COMPETENCY`).
- `src/openedx_tagging/rest_api/v1/serializers.py`: `taxonomy_type`, a write-only
  `ChoiceField` on `TaxonomySerializer`, default `"tags"`.
- `src/openedx_tagging/rest_api/v1/views.py`: `TaxonomyView.perform_create()` and
  `create_import()` discard `taxonomy_type` after validation — this app must never
  import `openedx_learning`. `create_import()`'s taxonomy-creation step is extracted
  into an overridable `_create_taxonomy_for_import()` hook.
- `src/openedx_learning/applets/cbe/api.py`: `create_competency_taxonomy()` — creates
  the `Taxonomy` and linked `CompetencyTaxonomy` rows in one transaction.
- `src/openedx_learning/applets/cbe/views.py` (new): `CompetencyTaxonomyView`, a
  `TaxonomyView` subclass overriding `perform_create()`/`_create_taxonomy_for_import()`
  to dispatch to `create_competency_taxonomy()` for `taxonomy_type="competency"`.

## Design notes

- `openedx_tagging` stays unaware `CompetencyTaxonomy` exists (`lint-imports` enforces
  it never imports `openedx_learning`). Plain `TaxonomyView` accepts
  `taxonomy_type="competency"` as a valid value but always creates a plain `Taxonomy`
  for it, since it structurally can't do anything else.
- `CompetencyTaxonomyView` overrides `TaxonomyView`'s creation methods rather than
  building on top of them, because that layering rule blocks the other direction:
  `openedx_tagging` can't extend itself with competency-aware behavior, since it
  can't reference `openedx_learning` at all. Overriding in a subclass that lives in
  `openedx_learning` (where both apps are visible).
- `CompetencyTaxonomyView` is a real subclass, not a mixin: one consumer for now
  (openedx-platform's `TaxonomyOrgView`).

## Verification

- `pytest tests/openedx_learning tests/openedx_tagging --no-cov`: 492 passed.
- `pylint`, `pycodestyle`, `isort --check-only`, `mypy`, `lint-imports`: all clean.

Related to openedx/openedx-core#628


mgwozdz-unicon (changes requested):
Claude and I worked together on this code review and are requesting the enumerated changes below. The layering is done right: `openedx_tagging`'s serializer and view changes never import or reference `openedx_learning`/`CompetencyTaxonomy` anywhere, `create_competency_taxonomy()` is the only place that touches both packages, and `lint-imports` confirms both contracts still hold. That matches the architecture constraint in #614 ("`openedx_tagging` must not import or instantiate `CompetencyTaxonomy`").

**1. `CompetencyTaxonomyView.perform_create()` returns a 500 instead of a 400 on a validation failure.**

In `src/openedx_learning/applets/cbe/views.py`, the competency branch of `perform_create()` calls `create_competency_taxonomy(**serializer.validated_data)` directly, with no error handling. The base `TaxonomyView.perform_create()` (the branch taken for `"tags"`, in `src/openedx_tagging/rest_api/v1/views.py`) wraps the equivalent call in `try/except exceptions.ValidationError` and re-raises DRF's `ValidationError` so it comes back as a 400. The competency branch skips that.

Claude reproduced it: `create_competency_taxonomy()` calls `create_taxonomy()` internally, which calls `taxonomy.full_clean()`. Posting `taxonomy_type="competency"` with an `export_id` that already exists raises `django.core.exceptions.ValidationError` from that `full_clean()` call, and it propagates unhandled, so the response is a 500. The same duplicate `export_id` through plain `TaxonomyView` returns a 400. Can you wrap the competency branch's call the same way the base class does, so both paths return a 400 on the same kind of failure?

**2. Please add a test that exercises `CompetencyTaxonomyView` itself.**

Every new test in `tests/openedx_tagging/test_views.py` posts to `TAXONOMY_LIST_URL` / `TAXONOMY_CREATE_IMPORT_URL`, which route to plain `TaxonomyView` (see `src/openedx_tagging/rest_api/v1/urls.py`), not `CompetencyTaxonomyView`. The two new methods on `CompetencyTaxonomyView`, i.e. the actual `taxonomy_type` dispatch this ticket exists to add, have no test going through the view at all. That's how item 1 got through. `CompetencyTaxonomyView` isn't registered on a URL in openedx-core yet, but it can still be tested directly via `APIRequestFactory` + `.as_view()` without one. At minimum: the competency branch of `perform_create()` returns a `CompetencyTaxonomy`, and it returns a 400 (not a 500) on a validation failure, same as the `"tags"` branch.

**3. `test_create_taxonomy_type_tags_or_omitted` and `test_import_taxonomy_type_tags_or_omitted` claim more than they check, and nothing else in the repo checks it either.**

Both docstrings say the endpoint "creates a plain Taxonomy" with no way to create a `CompetencyTaxonomy` row, but both only assert `Taxonomy.objects.filter(name=...).exists()`. Since `CompetencyTaxonomy` is multi-table inheritance from `Taxonomy`, that assertion would pass either way, so it doesn't test the claim. It looks like there's no `test_views.py` under `tests/openedx_learning/` at all, so nothing in the repo actually asserts that a `CompetencyTaxonomy` row is absent here. Since `tests/openedx_tagging` can't import `CompetencyTaxonomy` without breaking the layering rule it's demonstrating, can you add that assertion as its own test in `tests/openedx_learning`, with a docstring describing what it actually checks, and amend these two docstrings to only claim what their own assertions verify?

**4. The comment on where `export_id` gets auto-generated got dropped, and the replacement docstring doesn't cover it.**

The old `create_import()` had `# If no taxonomy_export_id provided, a unique export id will be generated` above the `create_taxonomy()` call. That's gone, and the new `_create_taxonomy_for_import()` base method doesn't explain it either. `create_taxonomy()`'s own docstring in `src/openedx_tagging/api.py` still just says "Creates, saves, and returns a new Taxonomy with the given attributes," it never documents the auto-generation behavior (`if not export_id: export_id = f"{count+1}-{slug}"`).

Rather than restoring that exact comment at one call site, can you add a line to `create_taxonomy()`'s own docstring instead? After this PR there are three places that can pass `export_id=None` into that auto-generation path: the base `_create_taxonomy_for_import()`, `CompetencyTaxonomyView._create_taxonomy_for_import()`, and `create_competency_taxonomy()`. Documenting it once at the function that owns the behavior avoids it going stale at whichever call site happens to keep the comment.

**5. Please explicitly state that there will be a openedx-platform follow-up in the PR description itself.**

tests/openedx_learning/applets/cbe/test_views.py
```diff
@@ -0,0 +1,144 @@
...
+    assert not CompetencyTaxonomy.objects.filter(name="Imported Tags Type Test").exists()
+
+
+def test_base_taxonomy_view_competency_type_creates_no_competency_row() -> None:
+    """
+    Posting taxonomy_type="competency" through the base TaxonomyView creates a
+    plain Taxonomy and no CompetencyTaxonomy row.
+
+    tests/openedx_tagging/test_views.py can't check this directly: it would have
+    to import CompetencyTaxonomy, breaking the layering rule it's demonstrating.
```
mgwozdz-unicon:
```suggestion
```
This feels like an odd AI relic to me.

tests/openedx_tagging/test_views.py
```diff
@@ -404,11 +404,12 @@ def test_create_taxonomy_read_only(self, create_data):
...
-        (this raw TaxonomyView) creates a plain Taxonomy: this endpoint must
-        never import openedx_learning, so it has no way to act on "competency" beyond
-        accepting it as a valid choice and discarding it -- creating a CompetencyTaxonomy
-        row for it is a downstream consumer's job.
+        Posting any accepted taxonomy_type (or omitting it) to this raw TaxonomyView
+        succeeds and creates a Taxonomy row.
+
+        Doesn't check whether a CompetencyTaxonomy row also gets created for
+        "competency". See tests/openedx_learning/applets/cbe/test_views.py for
+        that assertion.
```
mgwozdz-unicon:
```suggestion
```
This feels like an odd AI relic to me.

tests/openedx_tagging/test_views.py
```diff
@@ -3286,11 +3287,12 @@ def test_import_no_export_id(self, file_format) -> None:
...
-        endpoint (this raw TaxonomyView) creates a plain Taxonomy: this
-        endpoint must never import openedx_learning, so it has no way to act on
-        "competency" beyond accepting it as a valid choice and discarding it -- creating
-        a CompetencyTaxonomy row for it is a downstream consumer's job.
+        Posting any accepted taxonomy_type (or omitting it) to this raw create/import
+        endpoint succeeds and creates a Taxonomy row.
+
+        Doesn't check whether a CompetencyTaxonomy row also gets created for
+        "competency".
+        See tests/openedx_learning/applets/cbe/test_views.py for that assertion.
```
mgwozdz-unicon:
```suggestion
```
This feels like an odd AI relic to me.

mgwozdz-unicon (approved):
Approved, but I left some suggestions that I think clean up the comments better. Could you please also update the PR description to say `Related to openedx/openedx-core#628` instead of `See #628` to see if that connects the Github Issue and PR together?

src/openedx_learning/applets/cbe/api.py
```diff
@@ -3,16 +3,55 @@
...
+            read_only=read_only,
+            export_id=export_id,
+        )
+        competency_taxonomy = CompetencyTaxonomy(taxonomy_ptr=taxonomy)
+        # Copy the parent's fields onto the child instance: save_base(raw=True) below writes
+        # only the child's own row and skips Taxonomy entirely, so it never reads these back
+        # off the DB itself the way a normal save() of the MTI chain would.
+        for field in Taxonomy._meta.fields:
+            setattr(competency_taxonomy, field.attname, getattr(taxonomy, field.attname))
+        competency_taxonomy.save_base(raw=True)
```
ormsbee: @alezconsultant: I think this is fine for the purposes of this PR, but we probably want to have a short discussion about whether to extract this into a helper for `openedx_django_lib`.  This pattern of extending via subclass is something that's likely to happen more and more often given how we set up our apps in openedx-core. I believe that 

@bradenmacdonald, @kdmccormick, @mgwozdz-unicon, @jesperhodge: Want to get your thoughts here as well. FWIW, django-polymorphic does this kind of downcasting via [create_from_super](https://django-polymorphic.readthedocs.io/en/stable/api/polymorphic.managers.html#polymorphic.managers.PolymorphicManager.create_from_super). With containers, we flip the responsibilities by [passing the subclass](https://github.com/openedx/openedx-core/blob/d929ad54f601057d6755fd4605e4770c8db431e0/src/openedx_content/applets/containers/api.py#L135-L160) to `create_container`. I prefer doing something like `create_from_super` (and the code in this PR), because I think it's less confusing to have the extension happen in the code dealing with the subclass... so long as we have some common and well-maintained utility for doing it in a consistent way.

src/openedx_learning/applets/cbe/views.py
```diff
@@ -0,0 +1,45 @@
...
+from rest_framework.exceptions import ValidationError
+
+from openedx_tagging.api import TaxonomyType
+from openedx_tagging.models import Taxonomy
+from openedx_tagging.rest_api.v1.views import TaxonomyView
+
+from .api import create_competency_taxonomy
+
+
+class CompetencyTaxonomyView(TaxonomyView):
```
ormsbee: Isn't this introducing a lot of implicit coupling? If a client calls the list endpoint on this class-based-view, they'll get all taxonomies instead of just CBE ones, right?

ormsbee (commented):
I am concerned that by subclassing `TaxonomyView`, we're adding an unnecessary amount of coupling. If we only need to change how these taxonomies are created and deleted, I would prefer that we implement a narrow endpoint for that and not inherit the rest of the API interface—with the assumption that the frontend would point to the existing REST API for tags for most operations.

I believe @kdmccormick has been more in the loop on the discussions regarding this endpoint, so I defer to him on this review.

src/openedx_learning/applets/cbe/views.py
alezconsultant: Yes, we get all taxonomies, GET logic is part of another ticket, if i got it right.

src/openedx_learning/applets/cbe/api.py
jesperhodge: Hi @ormsbee , before I look deeper into this:
- "we probably want to have a short discussion about whether to extract this into a helper for openedx_django_lib" -> Great idea! If this is something that doesn't need to block this PR, can we move this out into a separate followup discussion so the PR can move along? But if you want a decision before the PR gets approved, we can discuss it first, of course.
- "I prefer doing something like create_from_super (and the code in this PR)" - for clarification, is this an action item / suggestion for this PR? Or is it part of a separate discussion you mentioned above?

bradenmacdonald: Personally, I would suggest simplifying this. Remove the `for field in Taxonomy._meta.fields: ... competency_taxonomy.save_base(raw=True)` block above and just keep it simple with `competency_taxonomy = CompetencyTaxonomy.objects.create(taxonomy_ptr=taxonomy)`.

Yeah it may use one more query, but creating competency taxonomies is something that happens fairly rarely, and nobody is going to notice the performance difference at all. So why over-complicate the code will all this advanced Django usage.

> I prefer doing something like `create_from_super` (and the code in this PR), because I think it's less confusing to have the extension happen in the code dealing with the subclass... so long as we have some common and well-maintained utility for doing it in a consistent way.

I don't have a strong opinion, but in the case of `Container`, the raw base class cannot be created on its own, so where we have it now seems like a natural place to put the logic that enforces the consistency. In other words, it seems fine to enforce it in the one place that every usage already has in common, rather than expecting that every subclass implements subclassing using the correct helper. But either way is fine with me in general.

bradenmacdonald: I am in favor of having a helper in `openedx_django_lib` to make things consistent in general. But wouldn't it be even better to just use django-polymorphic?

src/openedx_learning/applets/cbe/views.py
alezconsultant: @ormsbee @bradenmacdonald @mgwozdz-unicon I agree with the concerns about coupling in TaxonomyView, but we still have only one consumer in `openedx-platform` which is `TaxonomyOrgView` . As far as I understand, the main goal of this task is to allow the existing endpoints to create and import competency taxonomies by providing `taxonomy_type` in the request body. So this seems more like an extension of the current endpoints rather than a reason to introduce new ones. To avoid tight coupling, I suggest creating a mixin that can be plugged into `TaxonomyOrgView` from `openedx-platform`, for example:
`TaxonomyOrgView(MixinName, TaxonomyView)` Alternatively, we could create separate endpoints dedicated to creating and importing competency taxonomy records. In that case, I need to understand for what taxonomy types it will be.

src/openedx_learning/applets/cbe/api.py
jesperhodge: Here's my take if it helps, but I'll defer to your thoughts on this since you have more experience there.

Yes I agree with Braden. This seems unnecessarily complicated. If I understand correctly, this is because of an extra query just in the create function, which is insignificant? Or am I wrong and this affects reads elsewhere in the code somehow?
If there is no important reason for downclassing, I would second Braden's simplification.
If there is a good reason, I also think it would be better to use django-polymorphic. Since that's already a standard way of doing things, I think wrapping this in another standard inside of `openedx_django_lib` is not adding any benefit.

src/openedx_learning/applets/cbe/views.py
jesperhodge: My thoughts are that the tight coupling is a problem, and that we have a natural separation of concerns here. The Competency Taxonomy endpoints are part of the CBE applet, which is a different API segment than tagging. The endpoints are supposed to be different from each other, and defined separately.
If we want to reuse pieces of code, that's a good idea. Following good DDD principles, the views ideally just define the API endpoints - in two different places, separately, not coupled. But the business logic for DDD doesn't belong in the view but in a shared service or helper file that is used by both endpoints. Exactly for this reason: to enable separation of concerns between different endpoints.
Serializers are already extracted to a separate place in Django. Using the same serializer for two different endpoint seems completely fine, but again I would go with composition over inheritance there.
So I would suggest to keep the two endpoints separate without inheritance, and to keep it simple without mixins as well, and just extract any shared logic into a file that is shared between tagging and CBE.

tests/openedx_learning/applets/cbe/test_api.py
```diff
@@ -3,13 +3,63 @@
...
 from openedx_learning.models import CompetencyTaxonomy
 from openedx_tagging.models import Taxonomy
 
 pytestmark = pytest.mark.django_db
 
 
+def test_create_competency_taxonomy_saves_both_rows() -> None:
+    """
+    create_competency_taxonomy() saves a CompetencyTaxonomy and Taxonomy row that both
+    carry the given field values.
```
jesperhodge, line 16: The field values in the database only exist on the Taxonomy table though, right? The CompetencyTaxonomy table only has a pointer id

tests/openedx_learning/applets/cbe/test_api.py
```diff
@@ -3,13 +3,63 @@
...
+
+    Re-fetching from Taxonomy.objects (not just CompetencyTaxonomy.objects) is the
+    regression check for using save_base(raw=True): a plain save() on the child
+    instance would re-save the parent Taxonomy row with blank/default field values,
```
jesperhodge: Is this true? My understanding was that save() will save the child and the parent, and save_base is just there to prevent a double write (but I think the double write is desired here...)
This sounds like it would overwrite the parent fields with empty values?

tests/openedx_learning/applets/cbe/test_views.py
```diff
@@ -0,0 +1,141 @@
...
+    force_authenticate(request, user=_staff_user())
+
+    response = create_view(request)
+
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+
+def test_create_import_tags_creates_no_competency_row() -> None:
+    """
+    Importing with taxonomy_type="tags" through CompetencyTaxonomyView creates a
```
jesperhodge: I don't think this is desired behavior. If you want to create a tag taxonomy, you shouldn't use the competency taxonomy endpoint. The test should probably say:
"""Importing with taxonomy_type="tags" through CompetencyTaxonomyView throws a ValidationError""".

This seems like an unfortunate side effect of inheriting from the taxonomy view, a good reason not to do inheritance there.

Also I would like to point out that tests like this can create significant problems. What happened here, in my guess, is that first faulty code was written, and then after the fact a test was created that just confirms whatever behavior the code has, including contained bugs.

The bigger problem here is that the test is telling the application: "Hey, this bug is expected behavior", explicitly building a bug into the application design.

So - especially now that we use AI, we need to be extremely cautious when reviewing tests, and write the tests before the code is written not to match the behavior of the code, but instead match the acceptance criteria of the ticket.

tests/openedx_learning/applets/cbe/test_views.py
```diff
@@ -0,0 +1,141 @@
...
+
+    assert response.status_code == status.HTTP_201_CREATED
+    assert Taxonomy.objects.filter(name="Imported Tags Type Test").exists()
+    assert not CompetencyTaxonomy.objects.filter(name="Imported Tags Type Test").exists()
+
+
+def test_base_taxonomy_view_competency_type_creates_no_competency_row() -> None:
+    """
+    Posting taxonomy_type="competency" through the base TaxonomyView creates a
+    plain Taxonomy and no CompetencyTaxonomy row.
```
jesperhodge: This does not sound correct. There is no such thing as a taxonomy with a type competency but no competencytaxonomy entry. Where does this come from?

This contradicts the acceptance criteria.
"Given a POST request to /api/content_tagging/v1/taxonomies/
And the request body includes taxonomy_type set to "competency"
When the endpoint processes the request
Then the response returns 201
And a CompetencyTaxonomy record is created linked to the new Taxonomy"

Now I'm not fully certain why the AC mainly talk about the /taxonomies/ endpoint and not the competency taxonomies endpoint, which is maybe something to discuss.

But I'm fairly certain for now we just follow the acceptance criteria above.
The test should still be there and test exactly the opposite behavior.

jesperhodge (changes requested):
Great work! Awesome to see this coming together.

Update: I am a bit confused about the requirements, so I need to look through this again. Some of my previous PR comments were incorrect, so I'm reevaluating.

src/openedx_learning/applets/cbe/views.py
mgwozdz-unicon: Worked through this with Claude, and I think "keep the two endpoints separate" reads two different ways, and they lead to different places.

If it means two different URLs, the existing tags create endpoint plus a second, CBE-owned endpoint for competency taxonomies, that's the "New combined competency REST API" alternative that ADR 0013 (competency-taxonomy-detection) already considered and rejected. The reasoning there was that a separate CBE endpoint is a new surface to design, build, and maintain, and the frontend still has to choose which one to call before it makes the request, since nothing upstream tells it in advance whether it's creating a Tags or Competency taxonomy. That just moves the type-awareness into the frontend instead of removing it, which is the opposite of what this PR's `taxonomy_type` field is meant to do.

If instead it means two separately-defined view classes in code, with only `TaxonomyOrgView` ever bound to a real URL, and `TaxonomyOrgView.perform_create()` / `_create_taxonomy_for_import()` calling `create_taxonomy()` or `create_competency_taxonomy()` directly based on `taxonomy_type`, with no CBE-owned view class sitting in between, then I agree with that approach. It keeps this as a single endpoint, avoids any inheritance between the tagging and CBE view layers, and matches what I've been describing as the plan for `openedx-platform`: that repo is where the branching lives to create a `CompetencyTaxonomy` record versus a plain `Taxonomy` record, not a shared view class in `openedx-core`.

Concretely, that would mean this PR doesn't need a `CompetencyTaxonomyView` in `openedx-core` at all. `TaxonomyView` keeps the `_create_taxonomy_for_import()` hook it already has, and `create_competency_taxonomy()` stays in `cbe/api.py`. The follow-up `openedx-platform` PR wires `TaxonomyOrgView` to call both directly based on `taxonomy_type`. That PR needs an edit to `perform_create()` regardless, since it currently calls `create_taxonomy()` directly without going through `super()`, so routing it through this dispatch instead isn't extra work on top of what's already required there.

src/openedx_learning/applets/cbe/api.py
mgwozdz-unicon: I agree with Braden's simplification suggestion. I'm not strongly opposed to django-polymorphic, but I don't think it fits in our scope, and we're already planning to take a different approach on https://github.com/openedx/openedx-core/issues/618 (which is where I think this part of the discussion might be more relevant) by adding `select_related` in `TaxonomyOrgView.get_queryset()` to avoid the extra per-row query without a new dependency.

mgwozdz-unicon: Claude is now flagging to me that the `CompetencyTaxonomy.objects.create(taxonomy_ptr=taxonomy)` simplification won't work for our needs. It's saying that this approach will call Django's normal `save()` which will overwrite the taxonomy's `name` and `export_id` values with blank strings, but the current code avoids this. It might be worth it to update the comment on this code to also warn that a plain `.save()` or `.objects.create()` there would blank the parent row, since we missed that this first time.

tests/openedx_learning/applets/cbe/test_api.py
mgwozdz-unicon: Yes

tests/openedx_learning/applets/cbe/test_api.py
mgwozdz-unicon: Yes, this is true as written. Claude tested this and confirmed that the parent fields would get overwritten by a plain `save()`.

src/openedx_learning/applets/cbe/api.py
kdmccormick: hm, I too feel like this is a hack and doesn't belong in the CompetencyTaxonomy constructor. I don't really like the idea of abstracting it out into `openedx_django_lib` either, though. Either way, it's bespoke and probably has edge cases that we're not fully considering.

instead, how about following in the footseps `create_container`, which accepts a subclass `container_cls` and instantiates that rather than instantiating the superclass? https://github.com/openedx/openedx-core/blob/d929ad54f601057d6755fd4605e4770c8db431e0/src/openedx_content/applets/containers/api.py#L135-L179

this would be a change to the `create_taxonomy` definition. It could be made a backwards-compatible change by making `taxonomy_cls` optional, e.g. `taxonomy_cls: type[Taxonomy] = Taxonomy`.

(i would prefer this change be made in this same PR)

alezconsultant: Thanks for pointing out the `create_container` approach. I've updated the PR to use it.

ormsbee (approved):

src/openedx_learning/applets/cbe/views.py
bradenmacdonald: I would have preferred that we kept the "create competency taxonomy" as a separate REST endpoint, and that the core taxonomy python API accepts `taxonomy_type` but the taxonomy REST API does not. However, that would require that creating a competency taxonomy only happens from a dedicated CBE UI, and the approach chosen by the product team here is to allow creating competency taxonomies from the regular taxonomy UI, with the regular taxonomy endpoint. So given that, I think @mgwozdz-unicon's summary here makes sense and matches the planned approach.

ormsbee (changes requested):
Please add a version bump to 1.4.0 with this and squash with a new commit message.

ormsbee (approved):

Issue #673, opened 2026-07-20
[BE] Support editing an existing tag's external_id (Competency ID) on taxonomy re-import


thelmick-unicon:
## Use Case

As a Platform Administrator, I want to correct or change the Competency ID of an existing competency by re-importing its taxonomy, so that a mistyped or externally-changed identifier can be fixed in place without deleting and recreating the competency (which may already carry content associations or be protected from deletion).

## Acceptance Criteria

These scenarios cover the import-time `external_id` rename described above. The import is performed through the Studio taxonomy import wizard: the admin uploads a CSV or JSON file that includes the optional `previous_id` column, the wizard shows a **Plan** preview and requires confirmation, and validation errors surface at the Plan step and block the admin from proceeding. Because `external_id` is **not** shown in the import wizard or the tag list, the resulting `external_id` value is verified by **exporting** the taxonomy from the wizard and inspecting the exported file: `external_id` round-trips as the file's `id` column (CSV) / `id` key (JSON), and `previous_id` is never exported. Export is available in the wizard today, so this verification does not depend on the Competency Management page (4.6). Both export formats are covered below.

The wizard renders validation errors as a single block of text supplied by the backend (`ImportTagsWizard.jsx` prints the API's error message verbatim in one alert; it holds no per-error-type copy of its own). Two consequences for every scenario below that says the wizard "identifies the offending row": the identification has to be inside the message the backend produces, since the wizard cannot add it, and when a file contains more than one rejected row every row's error must appear in that one block rather than only the first, so an admin correcting a large framework is not forced into one upload per mistake.

The Studio taxonomy import is always a **full replace** (the uploaded file is the complete desired state; tags omitted from the file are removed), so a test import file must contain the taxonomy's full desired state.

Two rules govern how a file's rows relate to one another, and both are validated at the Plan step:

- An `id` that is currently held by another competency may be claimed only if the same file also removes that competency (by omitting it). If the file keeps both, the row is rejected.
- Every `parent_id` must reference the `external_id` its parent will hold **after** the import, not the one it held before. A row whose `parent_id` names an `external_id` that the same file renames away is rejected.

```
Scenario: rename an external_id, verified via CSV export
  Given a competency taxonomy contains a tag whose external_id is "COMP-A"
  When the admin imports a file whose row has previous_id "COMP-A" and id "COMP-B"
  Then the import succeeds (a "<taxonomy> updated" toast) with no duplicate competency
  When the admin exports the taxonomy as CSV
[...]

PR #804, opened 2026-09-03
feat: support editing an existing tag's external_id on taxonomy re-im…

ufedaseyeuconsultant:
# Description
Implements [ADR 0010](https://github.com/openedx/openedx-core/blob/main/docs/openedx_tagging/decisions/0010-mutable-tag-external-id.rst): the tag import file format gains a new optional, import-only column, **previous_id**. When a row's **previous_id** matches an existing tag's **external_id** in the taxonomy, and the row's id differs from it, the import renames that tag's external_id in place (along with any other changed fields) instead of deleting the old tag and creating a new one. This preserves the tag's primary key and existing associations (e.g. **CompetencyCriteria** links) across an institution-driven identifier rename.

**previous_id** is transient: it's read from the import file, consumed while building the import plan, and never persisted on **Tag** or written back out on export. No model field, no migration.

# Changes

- src/openedx_tagging/import_export/parsers.py: Parser gets a new import_only_fields = ["previous_id"] class attribute, parsed with the same missing/blank → None coercion as optional_fields. Unlike optional_fields, it's never added to CSV headers, never required by _verify_header, and never emitted by export.
- src/openedx_tagging/import_export/import_plan.py: TagItem gains previous_id: str | None = None. In generate_actions's replace-mode delete sweep, a renamed tag's old id is now also excluded from deletion (its new id was never a delete-sweep key to begin with).
- src/openedx_tagging/import_export/actions.py: new RenameTagExternalId action (name = "rename_external_id"), registered in available_actions right before CreateTag. It fires when previous_id is set and differs from id; validates that previous_id matches an existing tag, that the new id doesn't collide with another tag (in the DB or already queued in this import), and value/parent as needed; execute() updates the matched tag's external_id, value, and parent in place. CreateTag.applies_for is guarded off for rename rows so they're never double-handled.
- Fixed a related gap surfaced during review: the shared _validate_value/_validate_parent helpers only recognized CreateTag/RenameTag as sources of a same-import forward reference. Without also checking RenameTagExternalId, a value collision between a rename row and another row in the same import could slip past validation and hit an uncaught IntegrityError at execute time, and a legitimate parent reference to a renamed-in tag would be falsely rejected. Both helpers now also check RenameTagExternalId entries.

# Test coverage
previous_id parsing (CSV + JSON, absent/blank → None, never exported), RenameTagExternalId applies/validate/execute, the CreateTag guard, unmatched-previous_id and new-id-collision rejections, the two validation-gap regressions above, single-action generation (no spurious create+delete pair), replace-mode delete protection, and an end-to-end import→export round-trip that preserves PK and drops the old id.

# Tested via
* Unit tests
* Studio UI
* API

🤖 Generated with help of [Claude Code](https://claude.com/claude-code)


mgwozdz-unicon (changes requested):
Claude and I worked through this together and are requesting the changes below before merge. The scope is right: everything stays inside `src/openedx_tagging/import_export/`, no model field or migration is added, `previous_id` is never persisted or exported, and the two validation fixes called out in the PR description (`_validate_parent` and `_validate_value` learning about `RenameTagExternalId`) are correct and covered by tests. The gaps below are all about same-import row interactions: several of them come down to validation checking only the current database state and rows processed earlier in the file, never rows later in the file or the DB state other same-import actions will produce once they run.

**1. Two rows with the same `previous_id` pass validation, then crash at execute time instead of failing cleanly.**

In `src/openedx_tagging/import_export/actions.py`, `RenameTagExternalId._validate_new_id` checks the new `id` against the database and against other queued actions' `id`, but never checks whether another row in the same import already claims this row's `previous_id`. Given a file with row 1 `{previous_id: "A", id: "B"}` and row 2 `{previous_id: "A", id: "C"}`: at validate time, both rows call `taxonomy.tag_set.get(external_id="A")` against the unmodified database and both find the same tag, since nothing has executed yet, so both rows validate cleanly (assuming B and C don't otherwise collide). At execute time, row 1's `execute()` renames tag A's `external_id` to B. Row 2's `execute()` then calls `self.taxonomy.tag_set.get(external_id=self.tag.previous_id)`, i.e. looks up `external_id="A"` again, which no longer exists, and raises `Tag.DoesNotExist`. That's unhandled inside the `@transaction.atomic()` block in `TagImportPlan.execute()`, so it propagates to the broad `except Exception` in `api.py`'s `import_tags()`, which logs the raw exception to the task log and reports the import as failed, instead of surfacing a clean "duplicate previous_id" validation error at the plan step the way the wizard's other rejections work. A duplicated `previous_id` value from a copy-paste mistake in the source spreadsheet hits this today. Can you add a same-import `previous_id` collision check to `_validate_new_id`, the same way it already checks for `id` collisions against prior `RenameTagExternalId` actions, and add a test with two rows sharing one `previous_id` to lock the fix in?

**2. Reusing an external_id that a replace-mode delete is freeing up in the same import is rejected, and needs to be allowed.**

`RenameTagExternalId._validate_new_id`'s existence check (`self.taxonomy.tag_set.filter(external_id=self.tag.id).exists()`) in `src/openedx_tagging/import_export/actions.py` runs against the database as it is right now, not as it will be once other actions in the same import have run. If a file both omits tag Z (`external_id="Z"`, which a replace-mode import will therefore delete) and renames some other tag onto `id="Z"`, the rename row is rejected with "A tag with external_id (Z) already exists," because Z still exists in the database at validate time.

The fix looks like widening `_validate_new_id`: don't treat a collision as real if the colliding tag is itself in the `tags_for_delete` set that `import_plan.py`'s `generate_actions` builds for the replace-mode delete sweep. Execution order should already be safe once that validation is relaxed: `_build_delete_actions` runs before the per-row action loop in `generate_actions`, so delete actions get lower indices and `TagImportPlan.execute()` runs them before the row-based rename that reuses the freed id. That said, this is inference from reading the ordering, not something we've run, so it needs a test that actually executes a replace-mode import doing this, not just a validation-level check.

**3. Renaming two tags to each other's prior ids (a swap) needs to work, and it needs more than a validation fix.**

Given a taxonomy with tag X (`external_id="A"`) and tag Y (`external_id="B"`), a file with row 1 `{previous_id: "A", id: "B"}` and row 2 `{previous_id: "B", id: "A"}` fails at row 1 for the same reason as item 2: `tag_set.filter(external_id="B").exists()` is `True` because Y still has `external_id="B"` at validate time, so the row is rejected with "A tag with external_id (B) already exists." The same rejection happens in the opposite row order.

Here the underlying `(taxonomy, external_id)` unique_together constraint is enforced immediately on `save()`, not deferred, and that's true for every backend this project runs on (SQLite and MySQL don't support deferrable unique constraints the way Postgres does). So even if validation is relaxed the same way as item 2, executing the two renames in either order still hits that constraint: renaming X to B first collides with Y, which still holds B at that point, and renaming Y to A first collides with X, which still holds A. Whichever order `execute()` uses, the second `save()` raises `IntegrityError`, and since that's worse than today's behavior (a validation-time rejection would become a raw execute-time database error), this can't be closed by only touching `_validate_new_id`. Supporting the swap needs `execute()` to stage the affected tags through an intermediate external_id, or the action list to detect the cycle and reorder around it, and the plan should say which approach it's taking rather than leaving it to fall out of whatever `execute()` happens to do today. Whichever approach it takes, please add a test that runs an actual two-tag swap through `import_tags()` end to end, not just a `generate_actions()`-level check.

**4. `parent_id` must reference a tag's desired end-state external_id, and `_validate_parent` needs to reject a stale one instead of accepting it.**

The import already supports moving a tag to a different parent via `UpdateParentTag`, so `parent_id` already means "this tag's desired parent," not "the parent it had before this import." `_validate_parent` in `actions.py` doesn't enforce that consistently: a child row referencing the parent's new id only validates when the rename row comes first in the file, via the existing `_search_action` check against `RenameTagExternalId`, the same rule `CreateTag` already follows, but a child row referencing the parent's old id validates unconditionally, because `taxonomy.tag_set.get(external_id=self.tag.parent_id)` still finds the not-yet-renamed tag in the database. That's the same execute-time crash as item 1 whenever the rename runs first, and it gets worse once item 3's swap support lands, since a stale old id could then resolve to a different tag entirely instead of just failing. Can you tighten `_validate_parent` to reject a `parent_id` that matches a tag whose external_id is being vacated by another row's `previous_id` in this same import, instead of resolving it against live database state, and add tests for both the accepted new-id reference and the rejected old-id one?

**5. The AC in #673 names a CSV-specific verification scenario, but the round-trip rename tests only run through JSON.**

`test_parsers.py` has parallel CSV and JSON tests for parsing `previous_id` and for confirming it's excluded from export, so the parser layer is genuinely format-agnostic, `_parse_tags` in `parsers.py` handles `import_only_fields` the same way for both. But every test that runs a rename through the full `import_tags()` / `export_tags()` pipeline in `test_api.py` (`test_import_rename_external_id_preserves_pk`, `test_import_rename_external_id_then_export`, the two rejection tests) builds its import file with `json.dumps(...)`. Issue #673 lists "rename an external_id, verified via CSV export" as its own acceptance scenario, separate from the JSON one. Can you add a CSV version of the preserves-pk/then-export pair, so the CSV path gets the same end-to-end coverage as JSON, not just parser-level coverage?

**6. The replace-mode interaction, the primary path per the ticket, is only tested at plan-generation level, not through a real execute.**

`test_import_plan.py`'s `test_generate_actions_rename_external_id_replace_skips_delete` confirms that a renamed tag's old id is excluded from the generated delete-action list, which is the right check at that level. But nothing calls `.execute()` to confirm the tag actually survives in the database after `import_tags(..., replace=True)` runs end to end. Issue #673 says "the Studio taxonomy import is always a full replace... this is the primary path," so the case this PR exists to fix is exercised by the wizard exclusively with `replace=True`. Can you add one `test_api.py` test that runs a rename through `import_tags(replace=True)` and asserts the renamed tag is still present (not just that its delete action wasn't queued at the plan stage)?

**7. Issue #673's idempotent re-import scenario, `previous_id` equal to `id`, has no end-to-end test.**

`RenameTagExternalId.applies_for`'s test data covers the unit-level guard (`('tag_1', 'tag_1', False)`), confirming the action correctly declines to fire when `previous_id` equals `id`. But no `test_api.py` test runs that case through `import_tags()`, so nothing confirms the AC's actual scenario: re-importing a tag with `previous_id` set to its own current `external_id` succeeds with no error, and a follow-up export still shows the same id. Can you add that as a `test_api.py` test alongside the other rename scenarios?

I'll follow up with @thelmick-unicon to make sure that the AC in the ticket cover all of the relevant cases.

mgwozdz-unicon (changes requested):
Thanks for the update, I re-reviewed and all seven items from the previous review are addressed. Great work!

Three things I'd like addressed before merge.

**1. Naming: rename `StageTagExternalId` to `StageTagExternalIdForSwap`, and `is_staged_away` to `is_awaiting_placeholder_swap`.** Both current names read clearly once you've read their docstrings, but "staged" alone doesn't say what it's staged for, and a reader hitting either name cold has to go find the docstring to learn that a tag is moved to a placeholder id specifically so another row's rename can claim its old id, ahead of the tag's own rename landing on its final id right after. `StageTagExternalIdForSwap` and `is_awaiting_placeholder_swap` say that directly and share the same word, so the class and the check it feeds read as one connected idea instead of two.

**2. Duplication in `actions.py` between `RenameTag.applies_for` and `UpdateParentTag.applies_for`.** Both have the identical docstring paragraph about not applying when the matched tag is queued for deletion or is itself a `previous_id` rename row, and near-identical bodies: the same delete-queued check, the same early return for `previous_id` rows, differing only in the final comparison (`value` changed vs. `parent` changed). Please extract the shared check into one helper both call.

**3. Confirmed bug: two rows in the same import can specify the same final `id` with nothing catching it, and depending on row order that either crashes the import or silently overwrites the wrong tag.** Claude ran this end to end against your branch to confirm it. Given tag_1 and tag_3 swapping external_ids in one import, plus a third, unrelated row that also targets `id: "tag_1"` with a different value:

```json
{"tags": [
  {"id": "tag_3", "value": "Tag 1", "previous_id": "tag_1"},
  {"id": "tag_1", "value": "Tag 3", "previous_id": "tag_3"},
  {"id": "tag_1", "value": "Something Else Entirely"}
]}
```

If the third row is processed after the two rename rows, as above, the import reports success but silently overwrites the tag that the second row just renamed onto `tag_1`, discarding the value that row asked for. Move the third row to the front of the file instead, and the import fails, with an uncaught `Tag.DoesNotExist` in the task log. Neither outcome is right: this file should be rejected outright, since the second row and the third row both claim `tag_1` as their final id, and the same tag can't simultaneously be the one renamed in from tag_3 and the one holding an unrelated row's new value.

That's the actual bug: a tag only ever gets staged by `StageTagExternalIdForSwap` (the rename from item 1) when its current external_id is already another row's explicit target `id`, so any row that collides with a staged tag is, by construction, also duplicating that other row's `id`. Please add a check that rejects two rows in the same import specifying the same final `id`, run before the swap-staging and per-row action logic, so a file like the one above fails cleanly at the Plan step with an error identifying both offending rows, the same way a duplicate `previous_id` already does.

mgwozdz-unicon (approved):
Thank you for addressing those! This looks good. My only remaining nit would be to remove where some of the docstrings talk about how things were "previously". In probably all cases except maybe the unit tests, this reads as AI bloat to me.

src/openedx_tagging/import_export/actions.py
```diff
@@ -98,15 +101,30 @@ def _search_action(
...
+
+        parent_id must reference a tag's desired end-state external_id, not
+        whatever external_id currently resolves to some tag in the database:
+        UpdateParentTag/RenameTagExternalId already let a tag's own identity
+        change mid-import, so a parent_id matching a tag that's being renamed
+        away from that exact external_id in this same import is stale and must
+        not be accepted at face value -- fall through to the same
+        "landed/created earlier in this import" check already used for a
+        brand-new or renamed-in parent, so a reference to the correct, new id
+        still works when that rename comes first in the file.
```
kdmccormick: this is very long and rambly. please reword it so that it's as straightforward as possible while still including the things a future developer would need to know (and nothing more).

you cannot count on LLMs to write good docstings; you need to review them yourself and edit them until they are easy for a human to read.

src/openedx_tagging/import_export/actions.py
```diff
@@ -35,10 +37,11 @@ class ImportAction:
 
     name = "import_action"
 
-    def __init__(self, taxonomy: Taxonomy, tag, index: int):
+    def __init__(self, taxonomy: Taxonomy, tag, index: int, target_pk: int | None = None):
```
kdmccormick:
```suggestion
    def __init__(self, taxonomy: Taxonomy, tag: TagItem, index: int, target_pk: int | None = None):
```

could you type-annotate this and all other instances of `tag` that you've touched? I believe the annotation here is `TagItem`, from `..import_plan`.

not your fault that it was originally un-annotated, but going forward this will help with review and help make sure that the typechecker is finding as many bugs as possible in the new code.

kdmccormick (commented):
Hey @ufedaseyeuconsultant , thanks for the PR. Here are couple preliminary comments. I'm still reviewing the business logic, should have that done today or tomorrow.

Please take a look through the Open edX AI policy and make sure you're following it: https://github.com/openedx/.github/blob/master/AI_POLICY.md

Also please share in the PR description how you manually tested this PR.

src/openedx_tagging/import_export/actions.py
ufedaseyeuconsultant: done, check if docstring are ok now please

src/openedx_tagging/import_export/actions.py
ufedaseyeuconsultant: done

kdmccormick:
@ufedaseyeuconsultant Please read and follow the AI Policy. https://github.com/openedx/.github/blob/master/AI_POLICY.md

ufedaseyeuconsultant:
@kdmccormick Please let me know which part of the policy I need to correct. Are you referring to co-authorship?

kdmccormick:
@ufedaseyeuconsultant yes I was, thank you for adding that. Working on another review round now

PR #812, opened 2026-09-10
ADR proposals for how learning packages relate to courses/runs [FC-0138]

bradenmacdonald:
Before I propose `AssetSet`, I want to settle some open questions around Learning Packages, and copy-on-write with these two ADRs.

View the rendered ADRs:
* [11. Do not implement copy-on-write](https://openedx-core--812.org.readthedocs.build/en/812/openedx_content/decisions/0011-do-not-implement-copy-on-write.html)
* [12. Course Learning Packages](https://openedx-core--812.org.readthedocs.build/en/812/openedx_content/decisions/0012-course-learning-packages.html)

AI disclosure: Claude helped me with research, writing, and formatting the RST. I wrote and edited much of the content myself.


.importlinter
```diff
@@ -23,13 +24,16 @@ layers =
     # particular, openedx_tagging must never know that CBE exists.
     openedx_learning
 
+    # Catalog: defines how content is organized into enrollable courses
+    openedx_catalog
```
bradenmacdonald: While I was at it, I've clarified that the catalog models can be aware of the content models, but not vice versa.

Edit: this will be moved to https://github.com/openedx/openedx-core/pull/818 instead.

docs/openedx_content/decisions/0011-do-not-implement-copy-on-write.rst
```diff
@@ -0,0 +1,50 @@
...
+- **Storage savings from copy-on-write specifically are insignificant.** :class:`Media` is already deduplicated by hash within a learning package. Once copies live in the same package, a plain copy of a component version links to the existing :class:`Media` row, so OLX text is stored only once, with or without copy-on-write; the same applies to asset files like images. The major storage space savings come from the hash-based deduplication within each learning package, not from copy-on-write specifically, and future ADRs will explore other ways of achieving similar benefits across learning packages.
+- **Containers cannot be shared anyway.** A :class:`ContainerVersion` points at an :class:`EntityList` whose rows reference specific child entities. A copy needs its own container versions regardless, so only leaf component versions would ever be shared.
```
kdmccormick: this, on its own, is very convincing

docs/openedx_content/decisions/0011-do-not-implement-copy-on-write.rst
```diff
@@ -0,0 +1,50 @@
...
+- **The blast radius is large.** Both rejected designs below change public models exposed through ``models_api``, require a DEPR of parts of the ``versioning`` helper, touch every applet, and need coordinated changes in ``openedx-platform`` (the XBlock runtime's version lookups, library history, restore, upstream sync and the clipboard).
+
+A future ADR will address reducing storage space by consolidating media across the learning packages that hold different versions of the same course, which captures most of the storage benefit without changing the publishing model.
```
ChrisChV, line 31: This is addressed in ADR 12. Would it be good to reference it directly here?

docs/openedx_content/decisions/0012-course-learning-packages.rst
```diff
@@ -0,0 +1,192 @@
...
+
+This is a compelling option with arguably more clarity, but it is a bigger change and likely not backwards-compatible in terms of API. It also slightly increases cognitive load. If we need to hang metadata off of the ``blob_namespace`` in the future, it would make sense to implement :class:`LearningPackageFamily`.
+
+One LearningPackage per CatalogCourse, with run-scoped entities
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The main alternative. All runs of a catalog course share one learning package; a new ``scope`` column on :class:`PublishableEntity` records which run each entity belongs to, with a null scope meaning "shared across runs". A rerun creates only the container spine, pointing at the previous run's components pinned to their published versions, and forks a component into its own scope on first edit.
```
ChrisChV, line 133: > A rerun creates only the container spine, pointing at the previous run's components pinned to their published versions, and forks a component into its own scope on first edit.

I feel this approach can be decoupled from the decision not to use a LearningPackage per CatalogCourse. My question is: can't this approach be used with "One LearningPackage per CourseRun"?

docs/openedx_content/decisions/0012-course-learning-packages.rst
```diff
@@ -0,0 +1,192 @@
...
+
+``LearningPackage.package_ref`` is set to the string form of the run's course key, e.g. ``course-v1:MITx+Math100+2026Fall``. This mirrors content libraries, which already use ``package_ref=str(library_key)``. Both ``package_ref`` and course keys are globally unique, so this is consistent.
+
+5. Reruns copy content but share a blob namespace
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Creating a rerun copies the source run's entities into a new learning package. To avoid also duplicating the asset *bytes*, :class:`LearningPackage` will gains an immutable ``blob_namespace`` UUID field, and :meth:`Media.path` becomes::
+
+    content/{learning_package.blob_namespace}/{hash_digest}
+
```
ChrisChV, line 82: This doesn't break the public URL contract in [ADR-0005](https://github.com/openedx/openedx-core/blob/main/docs/openedx_content/decisions/0005-serving-static-assets.rst?plain=1#L90), but we should state it explicitly in the text.

docs/openedx_content/decisions/0012-course-learning-packages.rst
```diff
@@ -0,0 +1,192 @@
...
+When a learning package is created for a rerun, it copies the ``blob_namespace`` of a sibling run's learning package rather than generating a new one. All runs of a catalog course therefore share one namespace, and identical asset files are stored once across all of them. :meth:`Media.write_file` already returns without writing when a file of matching size exists at the target path, so deduplication happens automatically on write with no change to the media API. A namespace never spans organizations, because the runs that share one are by construction runs of the same catalog course.
+
+:class:`Media` *rows* remain scoped to a learning package: the ``(learning_package, media_type, hash_digest)`` constraint is unchanged, and each package has its own rows even when they resolve to a shared blob. This preserves per-package accounting, cascading cleanup on delete, and the borrowing-by-copy model. **No code may depend on two learning packages sharing a blob namespace; it is a storage optimization only.**
```
ChrisChV, line 87: > No code may depend on two learning packages sharing a blob namespace; it is a storage optimization only

How will this be achieved in practice?

ormsbee (commented):
I'll leave a more detailed review in the next couple of days, but some high level thoughts/reactions:

### Media de-duplication

1. I think the `blob_namespace` idea is a clever workaround for our biggest storage issues, and I really like it.
2. I do wonder if we should just force the namespace to a shared org level for new entries though, since there seems to be very little downside to it.
3. I'm trying to think through whether it's really okay for it to be immutable, what the use cases might be for wanting to change it, or what cleanup looks like in the shared scenario.
4. I might quibble with the naming a bit. Something like `media_file_namespace` might make the intent clearer.

That being said, I'm broadly supportive of this, and I think whatever wrinkles there are can be worked out.

### Dependency relationship between `openedx_content` and `openedx_catalog`

The [Pathways ADR](https://github.com/openedx/openedx-core/blob/main/docs/openedx_learning/decisions/0007-pathway-catalog-content-split.rst) assumes the opposite relationship between `openedx_content` and `openedx_catalog`. (In retrospect, we should have spun off a separate ADR for this decision in one of those applets.) I was thinking that any context -> content mappings could happen from the `openedx_content` side.

### 1:1 relationship between CourseRun and LearningPackage

This is the one that I'm most concerned about. We are aiming for a future where we can have wildly flexible arrangements of content. I don't want to assume that the authoring and learning contexts are 1:1. I would like to be in a place where someday where we can export a self-contained LearningPackage that contains a Pathway, its half dozen courses, and a bunch of smaller things. Maybe this means that we'll need to make a new model to capture namespaced identifiers to resolve potential conflicts between different course runs, but I think this sort of thing is an acceptable trade-off.

Note that I'm _not_ saying that we should _force_ all runs of a course into the same LearningPackage, or try to force re-runs to share the same PublishableEntities. That presents permissions headaches for interconnected content. But I do think that we should leave the door open for scenarios where there is a team working on content in a more centralized way, and it makes sense for that team to have such broad permissions across multiple courses or smaller learning contexts (e.g. bootcamps). We could even have scenarios where people are doing a library sync for the course run as a whole--if it's mapped to the same `blob_namespace`, that will be relatively cheap.

docs/openedx_content/decisions/0012-course-learning-packages.rst
```diff
@@ -0,0 +1,192 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:class:`CatalogCourse` is a grouping for catalog, marketing and enrollment purposes. It holds no authored content and gains no relationship to ``openedx_content``.
+
+Content shared deliberately between runs of a catalog course is expressed the same way as content shared between any two learning contexts: by copying it, optionally with an upstream link back to its source.
+
+3. CourseRun holds the relationship, and openedx_catalog layers above openedx_content
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:class:`CourseRun` gains a nullable, unique foreign key to :class:`LearningPackage`. It is nullable because a course run may exist purely as a marketing or enrollment placeholder, or may still have its content in modulestore. It is unique because the relationship is one-to-one. This is also exactly analogous to how the ``ContentLibrary`` model in openedx-platform stores a relationship to :class:`LearningPackage`.
```
ChrisChV, line 67: Since the `LearningPackage` can be null, was there any consideration of deferring the course copy when creating a rerun? That is, the `LearningPackage` would remain null until a specific trigger occurs (opening the course in Studio comes to mind, though that is open for discussion). Only after the trigger is met would the course copy be generated and the `LearningPackage` set.

docs/openedx_content/decisions/0011-do-not-implement-copy-on-write.rst
```diff
@@ -0,0 +1,50 @@
+.. _openedx-content-adr-0011:
+
+11. Do not implement copy-on-write
```
kdmccormick, line 3: ADR 11 looks good to me, thank you! ✅ 

Hadn't considered "Containers cannot be shared anyway" but that seems like a slam-dunk reason on its own.

bradenmacdonald: Yeah. I know it's a bit unusual to have an ADR about _not_ doing something and keeping the status quo, but my head kept trying to loop back to some variant of that copy on write idea, so I wanted to document clearly why we should go in a different direction. And the container thing was something I hadn't realized before but is definitely a strong reason.

docs/openedx_content/decisions/0012-course-learning-packages.rst
bradenmacdonald: That's an interesting idea. We certainly could do that, and it would make sense to let people start enrolling in the new version while deferring the actual creation of it so that it can be based on the very latest version of the previous course once it actually gets created. I think it would be possible, but it's not something I planned for here.

docs/openedx_content/decisions/0012-course-learning-packages.rst
bradenmacdonald: I don't think we can strictly enforce it in practice, though we could try to write test fixtures that include two runs that _don't_ share a namespace, to try to catch any issues. That would likely live in the platform tests though.

docs/openedx_content/decisions/0012-course-learning-packages.rst
bradenmacdonald: I have a separate PR coming that makes major revisions to ADR 0005.

docs/openedx_content/decisions/0012-course-learning-packages.rst
bradenmacdonald: That's true, and it does sound like we may still want something like this in the future. Let me think about it.

ormsbee:
@kdmccormick:

> Confirming my understanding: We would have a site-wide namespace if not for the possibility of a birthday attack, right?

I agree with the points that @bradenmacdonald made. In practice, almost all sharing of content happens within an org today, so we don't lose much. And in addition to potential long term security issues, there are also cost and potentially performance issues:

1. Having an org-level separation makes it more obvious to operators when a certain org is using a disproportionate amount of resources.
2. It also guards against potential performance issues where we start getting bad behavior because the backing filesystem can't gracefully handle a really large directory. I don't think this will be an issue with modern filesystems, but keeping them separate both makes the directories smaller and gives a little more isolation so that one really bloated org media dir won't necessarily slow things down for everyone else. (Even if the fs itself is performant, there can be various libraries/utilities that can break down in those circumstances.)

bradenmacdonald:
> I would like to be in a place where someday where we can export a self-contained LearningPackage that contains a Pathway, its half dozen courses, and a bunch of smaller things

@ormsbee I'm not opposed to this idea by any means, and I'll try to incorporate that as a future option into the ADR. But I don't really grasp how it will be useful. If LearningPackages are a "behind the scenes" thing, how and why would users even do this? Do we have any requests for this from users/product teams? Or is this sort of package a developer-oriented thing that is created and imported to various instances?

> I do think that we should leave the door open for scenarios where there is a team working on content in a more centralized way, and it makes sense for that team to have such broad permissions across multiple courses or smaller learning contexts (e.g. bootcamps).

Definitely, but I don't think LearningPackage is a user-visible permissions boundary, is it? Permissions are usually set at the LearningContext level instead. The authz system should allow creating arbitrary groups of courses (e.g. all courses that match pattern `BOOTCAMP5_*` in their course code) and assigning permissions that way to groups of users.

bradenmacdonald:
> > Maybe this means that we'll need to make a new model to capture namespaced identifiers to resolve potential conflicts between different course runs, but I think this sort of thing is an acceptable trade-off.
> 
> Could this be the Usage table?

It could, but that would only namespace Components. Everything else in the course that may exist as versioned content (static assets, pages, grading policy, etc.) would also need some way of being associated with a specific run.

In https://github.com/openedx/openedx-core/issues/360 there was some discussion about making "Files & Uploads" and other content part of the hierarchy. In that case, a single table could be used both to represent the complete content hierarchy and to provide namespacing of component/container codes. But it seems weird to me to make everything part of the outline, when the learners don't see it that way.

I will try to incorporate this option into the ADR, but I'm leaning toward sketching it out in detail as a "rejected alternative" that we can consider in the future.

docs/openedx_content/decisions/0011-do-not-implement-copy-on-write.rst
bradenmacdonald: > This is addressed in ADR 12. Would it be good to reference it directly here?

Sure, fixed in 99dab6cb8c3c39aab6a558fb1452f238de983f17

bradenmacdonald:
I have made some updates to the ADR:

* Renamed `blob_namespace` -> `media_file_namespace`
* Change to de-duplicate assets at the per-org level going forward, still using the `media_file_namespace` concept though.
* [Added a big new "rejected" (but planned for the future) section for flexibly combining runs and more into shared learning packages](https://openedx-core--812.org.readthedocs.build/en/812/openedx_content/decisions/0012-course-learning-packages.html#separate-learningpackage-per-run-with-flexibility-to-combine-them)

docs/openedx_content/decisions/0012-course-learning-packages.rst
```diff
@@ -0,0 +1,219 @@
...
+
+Compared to decision 1 this requires no ``media_file_namespace`` field, because :class:`Media` rows and blobs are already deduplicated within a package, so identical assets and OLX text blobs across runs are stored once at both the row and the blob level. It does not save any entity rows: a rerun creates exactly the same entity, version, draft and published rows as under decision 1, just in a shared package.
+
+We rejected it because it pays most of the costs of the run-scoped alternative for a smaller benefit than ``media_file_namespace`` delivers on its own:
+
+- Every package-level operation that decision 6 gets for free becomes run-aware by prefix filtering: publishing one run, listing its drafts, reading its change logs, pruning it, exporting it, and deleting it. Deleting a run is a filtered bulk delete rather than a cascade, and ``backup_restore`` cannot export a single run without new filtering support.
+- Codes stop being opaque. Every lookup, URL (see :ref:`openedx-content-adr-0005`), import, export and upstream link must compose and parse the prefix, which is contrary to the existing assumption that these identifiers are opaque.
```
bradenmacdonald, line 183: > contrary to the existing assumption that these identifiers are opaque

I had some recollection that [OEP-68](https://docs.openedx.org/projects/openedx-proposals/en/latest/best-practices/oep-0068-bp-content-identifiers.html) or [ADR 0003](https://openedx-core.readthedocs.io/en/latest/openedx_content/decisions/0003-identifier-conventions.html) recommended that "codes should not be parsed" (e.g. avoid inventing custom prefixing schemes in "code" fields), but I can't find that now.

docs/openedx_content/decisions/0011-do-not-implement-copy-on-write.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Under this approach, ``version_num`` stops being unique per entity: resetting a draft and editing again produces a second version with the same number. That would break every place that treats ``(entity, version_num)`` as an identity, including the XBlock runtime's ``@version`` lookups, backup and restore, and upstream sync. It also removes the database constraint that today turns concurrent edits into an ``IntegrityError``, which would have to be replaced with optimistic locking in ``set_draft_version``.
+
+Link table: ``PublishableEntityVersionLink``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Move the entity relationship into a many-to-many table ``(entity, version, version_num)`` with unique constraints on ``(entity, version)`` and ``(entity, version_num)``. A copy is made by copying the source's link rows, so each entity keeps its own unique, monotonic numbering, the concurrency constraint survives, and the ``versioning`` helpers keep their contract. An optional ``previous_version`` field can still record lineage.
```
ormsbee, line 48: One thing we could consider for re-runs is to have optional fields on `PublishableEntity` that capture another entity and version that it was derived from. So when a re-run happens, we could create complete copies of everything, but basically say, "Component C2 in CourseRun2 was derived from Component C1 in CourseRun1, and the draft version of C1 was v10 at the time C2 was created." If we decided to support multiple runs in the same LearningPackage, it could be a lightweight way of tracking/differentiating what would otherwise be confusing duplicate data.

Or I guess we could also re-purpose `copied_from_block`, though I feel like this is a bit different.

In any case, none of this contradicts this ADR. It just came to mind when considering the idea of links.

ormsbee:
> @ormsbee I'm not opposed to this idea by any means, and I'll try to incorporate that as a future option into the ADR. But I don't really grasp how it will be useful. If LearningPackages are a "behind the scenes" thing, how and why would users even do this? Do we have any requests for this from users/product teams? Or is this sort of package a developer-oriented thing that is created and imported to various instances?

@bradenmacdonald: I admit that it's still a bit fuzzy in my head, but I think of LearningPackages as the authoring equivalent of a learning context, i.e. the standard owned thing that can be authored, imported, and exported. Maybe that one thing is a library that has a bunch of assorted content in it. Or maybe that shared, authorable thingy has a bunch of course runs. Or maybe all LearningPackages are Libraries, and we just happen to put various combinations of larger things in those Libraries. But I like the idea of the bounds of the LearningPackage being whatever is convenient and useful for the purposes of authoring. So if it makes sense for one team to own a bootcamp pathway and a bunch of mini-courselets that are associated with that bootcamp, then they should be able to operate on it in a single place.

ormsbee:
One of the reasons I assume that the LearningPackage is going to be the boundary for this is that our content is so interconnected that doing permissions on only subsets of content within a LearningPackage is going to be a real pain. In my mind, different authoring ownership == different LearningPackage + sync mechanism if they need to collaborate.

ormsbee:
> > > Maybe this means that we'll need to make a new model to capture namespaced identifiers to resolve potential conflicts between different course runs, but I think this sort of thing is an acceptable trade-off.
> > 
> > Could this be the Usage table?
> 
> It could, but that would only namespace Components. Everything else in the course that may exist as versioned content (static assets, pages, grading policy, etc.) would also need some way of being associated with a specific run.

Is this something that we could actually leverage Collections for? Introduce a new, nullable field to `CollectionPublishableEntity` that gives a Collection-local identifier? Entities still need a LearningPackage-scoped ref, but any custom mapping/conflict resolution could happen via this new field? We already import courses into their own Collection anyhow.

ormsbee:
I suppose at that point, the confusing part would be when to use which identifier, particularly if we think there would be fallback rules. Still, I think there's something there we can use without it turning into a giant mess.

ormsbee:
Riffing on the Collection thing a bit more—in the longer run, we could introduce Collection types, one of which is a CourseRunCollection. Each Collection type could be associated with its own libraries detail view.

One place where there's a mismatch in semantics is that creating child elements for something in a Collection doesn't automatically add those things to that Collection. I'm not sure if that would be a feature or a bug in this case...

docs/openedx_content/decisions/0011-do-not-implement-copy-on-write.rst
bradenmacdonald: Yes, we already have `copied_from_block` and [`ComponentLink`](https://github.com/openedx/openedx-platform/blob/d1d2e7889cddb44620af7693ad0f7e0cdc34471b/cms/djangoapps/contentstore/models.py#L143-L162), so I do think there's plenty of precedent for models that track where the copies came from. That's also what I was trying to get at with this part of the ADR:

> There are other ways to track a component’s history. Between DraftChangeLog, PublishLog, openedx-platform’s ComponentLink, and potential future RerunLog models, we can find ways to show a component’s complete provenance history without necessarily implementing copy-on-write nor completely duplicating it.

bradenmacdonald:
@ormsbee 

> One of the reasons I assume that the LearningPackage is going to be the boundary for this is that our content is so interconnected that doing permissions on only subsets of content within a LearningPackage is going to be a real pain. In my mind, different authoring ownership == different LearningPackage + sync mechanism if they need to collaborate.

If we're going to be designing around that idea, then I'd really like to see a product definition, and at least some sort of minimal UX plan and input from the authz team. Otherwise, I think the idea is too vague to be useful for architecture planning purposes. I'm also wary of the many times I've personally built out a backend feature that never became useable by actual users (e.g. pinned versions of containers).

bradenmacdonald:
@ormsbee I like the collections idea and have included a mention of it in the ADR now, 61baea906579db24c025b963b311130077db21db

ormsbee:
> > One of the reasons I assume that the LearningPackage is going to be the boundary for this is that our content is so interconnected that doing permissions on only subsets of content within a LearningPackage is going to be a real pain. In my mind, different authoring ownership == different LearningPackage + sync mechanism if they need to collaborate.
> 
> If we're going to be designing around that idea, then I'd really like to see a product definition, and at least some sort of minimal UX plan and input from the authz team. Otherwise, I think the idea is too vague to be useful for architecture planning purposes. I'm also wary of the many times I've personally built out a backend feature that never became useable by actual users (e.g. pinned versions of containers).

I take the point on containers pinning their target versions. But when it comes to LearningPackage permissions, I think we're already kind of locked in by existing UX and Product decisions. We've explicitly made content very interconnected in terms of edit and publish behavior. The last time we tried to [examine this issue](https://docs.google.com/document/d/1fAHmageqXBQGDvpXT0rCK584Yy_wO0DMkG8k2a_iJOQ/edit?usp=sharing) with the perspective of split ownership, I don't think we came up with a good way to do that split and keep that connected vision. One of the possible paths was to do a massive amount of implicit forking, where different versions of things were draft/publish at different versions depending on where they're referenced, but that has some really concerning knock-on effects that I think make it impractical.

docs/openedx_content/decisions/0012-course-learning-packages.rst
```diff
@@ -61,12 +61,12 @@ The cost is that reruns duplicate rows. This is accepted; see "Consequences".
...
+3. CourseContent holds the relationship
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-:class:`CourseRun` gains a nullable, unique foreign key to :class:`LearningPackage`. It is nullable because a course run may exist purely as a marketing or enrollment placeholder, or may still have its content in modulestore. It is unique because the relationship is one-to-one. This is also exactly analogous to how the ``ContentLibrary`` model in openedx-platform stores a relationship to :class:`LearningPackage`.
+A django model in ``openedx_content``, called ``CourseContent``, will hold the mapping between ``CourseRun`` and ``LearningPackage``. The ``CourseRun`` field will be unique, such that each run is associated with zero or one ``LearningPackage``, never multiple.
 
-This aligns with the `Proposed Catalog Models ADR`_, which states that the dependency runs from catalog to content, and the content applets are deliberately ignorant of what a package represents.
+To emphasize that ``CourseContent`` should not be used as the canonical "Course" model nor a target for foreign keys (``CourseRun`` should be used), the foreign key to ``CourseRun`` can double as the primary key for ``CourseContent``.
```
ormsbee: Nit: Can this be summarized as: ``CourseContent`` will have a `OneToOneField(primary_key=True)` to ``CourseRun``?

PR #818, opened 2026-09-14
ADR for openedx_catalog models (layering openedx_content above catalog) [FC-0138]

bradenmacdonald:
Based on the discussion in https://github.com/openedx/openedx-core/pull/812, we need to clarify whether the new **catalog** app depends on the **content** app, or vice versa.

See the new ADR itself for all the context: [**1. Role of Catalog’s CourseRun and CatalogCourse Models**](https://openedx-core--818.org.readthedocs.build/en/818/openedx_catalog/decisions/0001-catalog-models-usage.html)

~~This PR also revises a Pathways ADR, [7. Pathways: Split Between Catalog and Content](https://openedx-core--818.org.readthedocs.build/en/818/openedx_learning/decisions/0007-pathway-catalog-content-split.html)~~

Private ref: FAL-4383


kdmccormick:
thanks for writing this up. "Catalog layers over content" has always felt intuitive to me as well.

that said, applying this same direction to Pathways means that:

* PathwayContents will point to CourseRuns, thus PathwayContent must be in openedx_learning
* CatalogPathways will point to PathwayContents, thus CatalogPathways must also be in openedx_learning

it feels weird that CatalogCourses and CourseRuns would be in openedx_catalog, but CatalogPathways will be in openedx_learning. whereas if we flip the relationship so that openedx_catalog is at the bottom, then everything can point "down" to openedx_catalog. but there are other questions to address if we flip the relationship, so I am not sure yet.

a question that's been on my mind that might guide us a bit: where do you think the LearningContext and Usage tables would go?

src/openedx_catalog/ARCHITECTURE.md
```diff
@@ -24,8 +24,9 @@ flowchart TB
     style Modes fill:#ccc
     style Organizations fill:#ccc
 
-    Pathways["<a href='https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5148147732/Brief+Modular+Content+Delivery+-+Platform+Strategy'>**openedx_pathways**</a> (Pathway, PathwaySchedule, PathwayEnrollment, PathwayCertificate, etc.)"]
+    Pathways["<a href='https://openedx.atlassian.net/wiki/spaces/OEPM/pages/5148147732/Brief+Modular+Content+Delivery+-+Platform+Strategy'>**openedx_learning: pathways**</a> (Pathway definition, PathwayItem, PathwayEnrollment, PathwayCertificate, etc. The unversioned CatalogPathway lives in openedx_catalog.)"]
```
Agrendalath, line 27: Nit: I believe PathwayEnrollment should be tied to the CatalogPathway.

Agrendalath:
@bradenmacdonald, I like this idea! I already implemented these changes in https://github.com/openedx/openedx-core/pull/819.

@kdmccormick,
> CatalogPathways will point to PathwayContents, thus CatalogPathways must also be in openedx_learning

We could point the content to the `PublishableEntity` and ensure we bind the `Pathway` (content) within the `openedx_learning` API. This way, we won't handle it at the DB level, but it's a similar mechanism to `ContentLibrary` pointing to the `LearningPackage`. Please see the initial implementation: https://github.com/openedx/openedx-core/pull/819/changes#diff-80730dea8b20ad4b376488da48af09bb805b17470a49ec0165c32434bd304449R111-R125

bradenmacdonald:
> it feels weird that CatalogCourses and CourseRuns would be in openedx_catalog, but CatalogPathways will be in openedx_learning.

In some ways, it could also make sense, if we think of `catalog` and `content` applets as providing basic primitives and then "Pathways" being an application built with those primitives that lives entirely in `openedx_learning`. I do like the idea of Pathways being a fully self-contained app that can easily be turned off or on, and only using extension points rather than having its code spread out among many different apps. However, we haven't (yet) designed any suitable extension points in `catalog`.

I think we've always planned for Pathways to be a first-class thing within the catalog, and I'm also completely fine with that approach. To make it work, we can use the "FK to PublishableEntity" approach @Agrendalath is suggesting in https://github.com/openedx/openedx-core/pull/819 or create a join table in `openedx_learning` that links `CatalogPathway` to `PathwayContents`.

> a question that's been on my mind that might guide us a bit: where do you think the LearningContext and Usage tables would go?

We haven't yet had any need for a concrete `LearningContext` table, and I'm not sure if we ever will. If we do, there's also a question of if we'd need a `LearningContextType`. So far however, we've gotten by by having LearningContext as a vague concept with disparate concrete implementations (courses and libraries). If we did need it though, I would lean toward putting it in `catalog`.

Usage tables are a mapping of IDs to content, and so they feel like part of the content to me. I'd put them in `content`, and assume that they're generally modified alongside the content itself, in a similar manner.

Related to that, @Agrendalath's PR https://github.com/openedx/openedx-core/pull/819 reminds me that we have to decide where the enrollment tables should go (for now just pathways, but long-term we may move course enrollment as well). Since `catalog` manages enrollables it could arguably be catalog, but `learning` or even somewhere else could also make sense.

src/openedx_catalog/ARCHITECTURE.md
bradenmacdonald: I've mention this below, but I've been unclear on where enrollments will live.

docs/openedx_catalog/decisions/0001-catalog-models-usage.rst
```diff
@@ -0,0 +1,86 @@
...
+
+``openedx_catalog`` may import and hold foreign keys to ``openedx_content``. But ``openedx_content`` must never import ``openedx_catalog``.
+
+``openedx_learning`` (Pathways, Competency-Based Education, and more) sits above both.
+
+The resulting order, enforced by the ``src_layering`` contract in ``.importlinter``, is::
+
+    openedx_learning > openedx_catalog > openedx_content > openedx_tagging
+
+The general rule behind this ordering is: **a context model points at its content; content never points at contexts.** ``openedx_content`` is generic infrastructure used by courses, libraries, pathways and future context types, and its applets are deliberately ignorant of what a learning package represents. A course run, a library, or a pathway is the thing that knows which package (or which container within a package) holds its content, in exactly the way ``ContentLibrary`` already does.
```
ormsbee: Thank you for writing this out—I think I was struggling to understand the disconnect I was having, and this really helps to bring it into focus:

> ``openedx_content`` is generic infrastructure used by courses, libraries, pathways and future context types, and its applets are deliberately ignorant of what a learning package represents.

1. **To me, the ContentLibrary mapping exists in `openedx-platform` because it wasn't practical to do otherwise**, especially with the state of permissions at that time—one of the main motivations for the bottom-first build-out of `openedx-core` was the difficulty in extracting all the dependencies of higher level concepts. I think that it's actually perfectly reasonable in the long term that an app named `openedx_content` would contain the API for content libraries. This would be in a new applet that sits above the existing ones in terms of layering.
2. **I believe that the Catalog exists to serve the narrow "find a course/pathway and enroll in it" experience**, along with various queries for basic metadata like the title. I don't think the Catalog is a place that should be responsible for piecing together the broader notion of a course as students experience it.

My specific worry with the framing of this ADR is that I don't think it's the catalog's responsibility to map a CourseRun to the content that backs it. I believe that the question of, "What content do I display for this CourseRun, and how does that map to LearningPackages?" is a _content_ (and eventually _learning_) concern. Similarly, I would expect credentials/certificate information to sit in its own app and reference catalog, rather than the other way around. The modeling here doesn't need to be exotic, just a 1:1 table from the content side, referencing CourseRun.

Imagine that some CourseRuns will have their own LearningPackages, but we later decide that some other CourseRuns will point to different root nodes within the same LearningPackage. Maybe some wiki-like content is associated at the CatalogCourse level. Even later, we decide that CCX courses will be modeled as LearningPackage + root node + fkey to some other config. We then go off and have a long discussion about whether multiple CourseRuns should be allowed to point to the exact same content tree. Another conversation surrounds whether we run a data migration to consolidate existing runs into shared LearningPackages or not. All of those would be content modeling decisions, and it shouldn't require changes to `openedx_catalog`.

We've long been in a world where the enrollment context, the authoring context, and the learning context are all the same, and share the same identifier key. As a platform, I think that's limiting. I should be able to enroll in a Pathway and become associated with a variety of learning contexts inside of it. I should be able to enroll in a CourseRun that points into a content tree that lives in a centrally managed library. I believe that the Catalog is the start of that chain of things, but we should let things outside the Catalog figure out how they're going to map CourseRuns to their own data and areas of responsibility.

bradenmacdonald: > Imagine that some CourseRuns will have their own LearningPackages, but we later decide that some other CourseRuns will point to different root nodes within the same LearningPackage.... All of those would be content modeling decisions, and it shouldn't live or require changes to openedx_catalog.

I've tried to mention exactly those possibilities and keep them open in the latest version of each ADR, but you make a good point that they don't feel like things that should require changes to openedx_catalog. We can discuss this in our call tomorrow.

> We've long been in a world where the enrollment context, the authoring context, and the learning context are all the same, and share the same identifier key. As a platform, I think that's limiting. I should be able to enroll in a Pathway and become associated with a variety of learning contexts inside of it.

This may be only tangentially related, but for me it was often an enormous pain that the edx.org marketing site used completely different course IDs than the LMS, and worked pretty hard to hide the `course-v1:...` key. Several times I needed to correlate them and it required either actually enrolling into the course or digging through the HTML to find the hidden form with the actual course ID buried somewhere in there.

docs/openedx_catalog/decisions/0001-catalog-models-usage.rst
```diff
@@ -0,0 +1,86 @@
...
+
+This does not contradict the intent of :ref:`openedx-learning-adr-0007`, whose real requirements are that the versioned Pathway definition holds the references to the unversioned catalog objects. Those definition models live in ``openedx_learning``, above the catalog, so they can reference :class:`CourseRun` and content freely. Decision 4 of that ADR has been amended to name ``openedx_learning`` rather than ``openedx_content`` as the side that knows about the catalog.
```
ChrisChV: Was this same pattern of  `openedx_learning`, a neutral layer sitting above both catalog and content, with neither depending on the other, considered for the base `CourseRun - LearningPackage` mapping?

bradenmacdonald: That's a really good point, and worth discussing, so I've added it to the ADR as 831128d12bb609ea847382dfd1a28633aacb111e. I think it's a reasonable option.

bradenmacdonald:
OK everyone, based on the discussion at the Core Arch Working Group this week, we've found @ormsbee's argument pretty compelling, and I've inverted the proposal here. It's now consistent with the existing pathways ADR, and says content layers above catalog.

[Revised ADR](https://openedx-core--818.org.readthedocs.build/en/818/openedx_catalog/decisions/0001-catalog-models-usage.html)

docs/openedx_catalog/decisions/0001-catalog-models-usage.rst
kdmccormick: I really like this idea too. If the Pathway->Content and CompetencyCriteria->Content mappings will be in `openedx_learning`, then why not have the CourseRun->Content mappings there too?
 
fyi @ormsbee

bradenmacdonald: I think the pathways ADR assumed originally that the Pathway->Content mapping would also be in `openedx_content`, but yeah it could just as easily go into `learning`.

ormsbee (approved):

ChrisChV:
Looks good to me :+1:

docs/openedx_catalog/decisions/0001-catalog-models-usage.rst
```diff
@@ -0,0 +1,81 @@
...
+
+The general principle behind this is that changes in how content is represented should not require changes to the catalog app. For example, if we were to change from associating each course run with a :class:`LearningPackage` to associating each course run with an ``OutlineRoot`` in a :class:`LearningPackage` that contains multiple runs, that should not require changes to the catalog app, which would be the case if we used foreign keys from :class:`CourseRun` to :class:`LearningPackage` within the catalog app.
+
+3. Catalog models are the canonical target for course foreign keys
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For performance and correctness, any Django model in this repository or in ``openedx-platform`` that needs to reference a course should do so with a foreign key to :class:`CourseRun` (or, rarely, :class:`CatalogCourse`), rather than by storing a course key string or pointing at ``CourseOverview`` (although much existing code does not yet follow this new convention).
+
+On the other hand, public APIs and events should continue to identify courses by their full string course key and never expose the integer primary keys.
```
kdmccormick:
```suggestion
On the other hand, public REST APIs and events should continue to identify courses by their full string course key and never expose the integer primary keys.
```

public Python APIs can accept the PKs and model instances, right?

kdmccormick (approved):

docs/openedx_catalog/decisions/0001-catalog-models-usage.rst
bradenmacdonald: Yeah.

bradenmacdonald:
@ChrisChV or @Agrendalath any last comments or concerns before I merge this?

src/openedx_catalog/ARCHITECTURE.md
Agrendalath: @bradenmacdonald, I suppose we could remove `PathwayEnrollment` now?

Agrendalath (approved):
@bradenmacdonald, LGTM.

Just a minor note: the new `PathwayCategoryTranslation` model isn't mentioned in the ADR, but it shouldn't block merging this PR (https://github.com/openedx/openedx-core/pull/819 is waiting for review, so we would have a circular dependency). I can update the doc in https://github.com/openedx/openedx-core/pull/819 to include the new model (if we decide to go forward with it).

ChrisChV (approved):
Looks good!

